Issue #6, opened 2020-05-26
Write detail views for all models


blueNoteb5:
- [x] Chant
- [x] Feast
- [x] Genre
- [x] Indexer
- [x]  Notation
- [x]  Office
- [x] Provenance
- [ ]  Segment
- [ ]  Sequence
- [ ] Siglum
- [x]  Source

Issue #5, opened 2020-05-26
Write list views for all models


blueNoteb5:

blueNoteb5:
- [x] Chant
- [x] Feast
- [x] Genre
- [x] Indexer
- [x]  Notation
- [x]  Office
- [ ] Provenance
- [ ]  Segment
- [x]  Sequence
- [ ] Siglum
- [x]  Source

PR #10, opened 2020-06-12
Indexer views

blueNoteb5:
* Add a list and a detail view for the ``Indexer`` model
* The list view is searchable/filterable

refs #5, #6


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -0,0 +1,156 @@
...
+        self.assertEquals(response.status_code, 200)
+        # Check object_list (which has the whole queryset, not paginated) instead of
+        # indexers which is paginated and might not contain random_indexer if it
+        # is not on the first page
+        self.assertTrue(random_indexer in response.context["paginator"].object_list)
+
+        # Search for a three letter slice of the first name
+        first_name = random_indexer.first_name
+        first_name_length = len(first_name)
+        if first_name_length <= 3:
```
JRegimbal: Why not just test `if first_name_length > 3:` and remove the pass?

django/cantusdb_project/requirements.txt
```diff
@@ -1,6 +1,22 @@
+appdirs==1.4.4
 asgiref==3.2.7
-Django==3.0.7
+attrs==19.3.0
+black==19.10b0
+click==7.1.2
+Django==3.0.6
```
JRegimbal: Was the downgrade intentional?

django/cantusdb_project/main_app/templates/indexer_list.html
```diff
@@ -0,0 +1,45 @@
+{% extends "base.html" %}
+{% block content %}
+<div class="mr-3 p-3 col-md-10 mx-auto bg-white rounded">
+    <h3>List of Indexers</h3>
+    {% with count=indexers.count %}
+    <small>Displaying 1-{{count}} of <b>{{count}}</b></small>
```
JRegimbal: Is there no pagination for this view?

JRegimbal (commented):
Overall looks good with very useful comments! A few minor questions but that's it.

django/cantusdb_project/main_app/templates/indexer_list.html
blueNoteb5: There is, that's a mistake, I'll fix it

django/cantusdb_project/requirements.txt
blueNoteb5: Nope, GitHub had updated it but when I did ``pip freeze`` it went back. I'll fix it, good catch

django/cantusdb_project/main_app/tests/test_views.py
blueNoteb5: Good point, I'll rewrite it

PR #11, opened 2020-06-16
Feast views

blueNoteb5:
Adds feast detail view and feast list view, with unit tests


django/cantusdb_project/main_app/management/commands/add_feasts.py
```diff
@@ -25,30 +26,34 @@ def handle(self, *args, **options):
...
+                        if len(split_str) > 1:
+                            day = int(split_str[1])
+                            month = int(datetime.strptime(split_str[0], "%b").month)
+                    if element_name == "feast_code":
+                        if element_value != "xxx":
+                            feast_code = int(element_value)
+                    if element_name == "id":
+                        feast_id = int(element_value)
+                    if element_name == "notes":
+                        notes = element_value
```
JRegimbal: Not sure how the interpreter handles this but if it doesn't do any optimization then `elif` might avoid unnecessary checks (unless you plan on modifying `element_name` in a branch here).

django/cantusdb_project/main_app/templates/feast_list.html
```diff
@@ -0,0 +1,113 @@
...
+                    <option value="1">Jan</option>
+                    <option value="2">Feb</option>
+                    <option value="3">Mar</option>
+                    <option value="4">Apr</option>
+                    <option value="5">May</option>
+                    <option value="6">June</option>
+                    <option value="7">July</option>
+                    <option value="8">August</option>
+                    <option value="9">September</option>
+                    <option value="10">October</option>
```
JRegimbal: Is there a reason for the order and why some months are abbreviated while others aren't?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -152,3 +154,134 @@ def test_view_context_data(self):
...
+        for page_num in range(1, pages + 1):
+            response = self.client.get(reverse("feast-list"), {"page": page_num})
+            self.assertEqual(response.status_code, 200)
+            self.assertTrue("is_paginated" in response.context)
+            self.assertTrue(response.context["is_paginated"] == True)
+            if page_num == pages and (self.number_of_feasts % self.PAGE_SIZE != 0):
+                self.assertTrue(
+                    len(response.context["feasts"])
+                    == (self.number_of_feasts % self.PAGE_SIZE)
+                )
```
JRegimbal: `self.assertEqual` might be more descriptive code.
```suggestion
                self.assertEqual(
                    len(response.context["feasts"]),
                    self.number_of_feasts % self.PAGE_SIZE
                )
```

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -152,3 +154,134 @@ def test_view_context_data(self):
...
+        # To get total number of pages do a ceiling integer division
+        q, r = divmod(self.number_of_feasts, self.PAGE_SIZE)
+        pages = q + bool(r)
+
+        # Test all pages
+        for page_num in range(1, pages + 1):
+            response = self.client.get(reverse("feast-list"), {"page": page_num})
+            self.assertEqual(response.status_code, 200)
+            self.assertTrue("is_paginated" in response.context)
+            self.assertTrue(response.context["is_paginated"] == True)
```
JRegimbal: Seems redundant.
```suggestion
            self.assertTrue(response.context["is_paginated"])
```

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -152,3 +154,134 @@ def test_view_context_data(self):
...
+            if page_num == pages and (self.number_of_feasts % self.PAGE_SIZE != 0):
+                self.assertTrue(
+                    len(response.context["feasts"])
+                    == (self.number_of_feasts % self.PAGE_SIZE)
+                )
+            else:
+                self.assertTrue(len(response.context["feasts"]) == self.PAGE_SIZE)
```
JRegimbal:
```suggestion
                self.assertEqual(len(response.context["feasts"]), self.PAGE_SIZE)
```

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -152,3 +154,134 @@ def test_view_context_data(self):
...
+
+        # Test the "last" syntax
+        response = self.client.get(reverse("feast-list"), {"page": "last"})
+        self.assertEqual(response.status_code, 200)
+
+        # Test some invalid values for pages
+        response = self.client.get(reverse("feast-list"), {"page": -1})
+        self.assertEquals(response.status_code, 404)
```
JRegimbal: `self.assertEquals` seems to be [deprecated in favor of `self.assertEqual`](https://bugs.python.org/issue9424). This isn't the only instance of this.

JRegimbal (commented):
Good comments again! A couple of recommendations regarding using `assertEqual` over `assertTrue` when the latter is only being used to check equality. Also there's a mix of `assertEqual` and `assertEquals` used in here and it seems like [`assertEquals` is deprecated](https://bugs.python.org/issue9424).

django/cantusdb_project/main_app/management/commands/add_feasts.py
blueNoteb5: Good point, I'll change it

django/cantusdb_project/main_app/templates/feast_list.html
blueNoteb5: Good catch

django/cantusdb_project/main_app/management/commands/add_feasts.py
blueNoteb5: Changed it!

PR #12, opened 2020-06-18
Genre views

blueNoteb5:
* Add Genre detail and list views with unit tests
* Refactor the ``add_genres`` management command
* Some light formatting in already existing code

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/add_genres.py
+++ b/django/cantusdb_project/main_app/management/commands/add_genres.py
@@ -13,28 +13,25 @@ class Command(BaseCommand):
         file_path = options["xml_file"]
         root = ET.parse(file_path).getroot()
 
-        for feast in root.findall("genre"):
-            name, description, mass_office = (None, None, None)
-            for element in feast:
+        for genre in root.findall("genre"):
+            name, description, mass_office, genre_id = (None, None, None, None)
+            for element in genre:
                 element_name = element.tag
-                element_value = feast.find(element.tag).text
-                cleanr = re.compile("<.*?>")
-                element_value = re.sub(cleanr, "", element_value)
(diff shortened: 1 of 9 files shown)


django/cantusdb_project/main_app/management/commands/add_genres.py
```diff
@@ -13,28 +13,25 @@ def handle(self, *args, **options):
...
-                    description = element_value
-                if element_name == "mass_or_office":
-                    mass_office = element_value.split(", ")
-                    mass_office = [
-                        "Hispanic" if value == "Old Hispanic" else value
-                        for value in mass_office
-                    ]
-                if element_name == "id":
-                    genre_id = int(element_value)
+                element_value = genre.find(element.tag).text
```
JRegimbal, line 20: Is there a reason you can't do `element.text`, or something to avoid searching for the element you already have? If so there should probably be a comment explaining why.

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -264,24 +268,110 @@ def setUp(self):
...
+            # compare model objects to model objects
+            transform=lambda x: x,
+        )
+
+    # TODO: maybe make a more general method to test pagination that I can apply
+    # to all list views?
+    def test_pagination(self):
+        # To get total number of pages do a ceiling integer division
+        q, r = divmod(self.number_of_genres, self.PAGE_SIZE)
+        pages = q + bool(r)
```
JRegimbal, line 346: I'd recommend importing `math.ceil` and replacing these lines with `pages = ceil(self.number_of_genres / self.PAGE_SIZE)`.

JRegimbal (approved):

PR #93, opened 2022-05-19
implemented the user login/logout system

### LINKED ISSUE #89: User login interface
A user login interface needs to be implemented. It should take the username(email) and the password as input. This can be either in the footer of every page or in a designated login page.

jiwoojeongjj:
fixes #89


jiwoojeongjj:
**note:** 

The following path in `urls.py` is still in development and should not be used

```
path(
        "edit-volpiano/", 
        ChantEditVolpianoView.as_view(), 
        name="source-edit-volpiano"
    )
```

PR #116, opened 2022-05-30
implemented the admin navigation menu

### LINKED ISSUE #105: Admin navigation menu
Admin navigation menu needs to be implemented. Once the user logs in, the login interface should be replaced by the Admin navigation menu, which links to various editing-related features.

### LINKED ISSUE #108: The user-profile page and my-sources page need to have an entrance
Currently these two pages can be accessed by typing the url. We should put hyperlinks on the page, so that the user can easily navigate to those pages upon logging in. This is related to the admin navigation menu mentioned in #105 .

jiwoojeongjj:
fixes #105 and fixes #108

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -37,9 +37,10 @@
             function handleLogButtonClick() {
                 if (logButton.value === "Log In") {
                     location.replace(`${base_url}/login/`);
-                } else if (logButton.value === "Log Out") {
-                    location.replace(`${base_url}/logout/`);
-                }
+                } 
+                //else if (logButton.value === "Log Out") {
+                //    location.replace(`${base_url}/logout/`);
+                //}
             }
             logButton.addEventListener("click", handleLogButtonClick);
         }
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/templates/base.html
```diff
@@ -338,11 +339,46 @@ <h5>Database Links</h5>
                         </ul>
                     </div>
                     <div class="col-md" id="login">
-                        <h5>Login for Contributors</h5>
                         {% if not request.user.is_anonymous %}
```
jwang44, line 326: This `if` statement seems to be repeated in line 379. How about converting them into an `if`, `else` pair?

django/cantusdb_project/templates/base.html
```diff
@@ -37,9 +37,10 @@
             function handleLogButtonClick() {
```
jwang44: Is it possible to remove this piece of js all together? If I understand correctly, with the current design, we only need this button as a hyperlink to the login page, while logout will be handled by another hyperlink.

jiwoojeongjj: I think we still need to keep this around because the "Log in" button uses this. Unless you want the "Log in" button to be replaced with a hyperlink.

jiwoojeongjj: Fixed!

jwang44 (approved):

PR #131, opened 2022-06-01
reimplemented the my-sources page

### LINKED ISSUE #110: Created sources not showing up in "sources created by user"
I added a new source after logging in. However, when checking back on the my-sources page, I don't see that source in the "sources created by user" list to the right of the page. It is also not listed under "my sources". 
As I'm the one who created the source, I think it's reasonable to associate that source to me in some way. This needs to be implemented.

jiwoojeongjj:
fixes #110.

note: the pagination UI is ugly (but I can see that the pagination UI hasn't been customized yet). perhaps this is something we can work on in the future.


jwang44 (approved):

fujinaga:
> fixes #110.
> 
> note: the pagination UI is ugly (but I can see that the pagination UI hasn't been customized yet). perhaps this is something we can work on in the future.

Please post this as an issue with a lower priority.

PR #158, opened 2022-06-08
extracted out pagination UI into own template and customized

### LINKED ISSUE #157: make a template for the pagination UI
it would be ideal to have the pagination ui be extracted as its own template to reduce redundancy and make it easier for future customization

jiwoojeongjj:
fixes #157 

note: for pages where the pagination occurs in a card (on the right side of the page), the layout is not ideal. this will be improved upon.

for example, there is a pagination that occurs in a card in the my-sources page.


jwang44:
This is great! No more pagination copy and paste 👍

Issue #168, opened 2022-06-13
Don't use "first name" and "family name" for users


jacobdgm:
Further to Ich's comment on [pull request 111](https://github.com/DDMAL/CantusDB/pull/111), it's better to simply use a "name" field for users rather than having "first name" / "given name", "last name" / "family name" fields and then having to guess which order they go in.

fujinaga:
Yes, I agree; having just a "name" field is best. There would be an extra work of providing a good searching capability when listing all users in certain orders.

jacobdgm:
this is partially implemented: currently, we have a `full_name` field defined for the User model; users have a `first_name` and `last_name` that are inherited from the django's built-in AbstractUser, so there are First Name, Last Name and Full Name fields all present in the admin interface.

A direction we could go from here: I could write a script to populate all users' `full_name` fields, which Junhao can run once all the data has been pulled from OldCantus. once it's run, we can look into overwriting/deleting the first_ and last_name fields so they don't show up in the admin interface, and so no data is lost. Is there something else/different we should do with this?

lucasmarchd01:
On our site, users are now represented by their full name. The first name and last name fields are maintained in the admin area to avoid loss of data from OldCantus.

PR #169, opened 2022-06-13
Add name attribute to User class

jacobdgm:
A step towards addressing #168.

At some point, we'll need to convert all the references to "first_name" and "last_name" in the code base (`first_name` and `last_name` seem not to have been defined in our code, but rather they're inherited from Django's `AbstractUser` class), and update the scripts which import info on indexers etc. from OldCantus into NewCantus (as they appear to use first name and last name, so they'll need to be converted to `full_name`)

### DIFF
--- a/django/cantusdb_project/main_app/models/chant.py
+++ b/django/cantusdb_project/main_app/models/chant.py
@@ -1,3 +1,4 @@
+from this import d
 from django.contrib.postgres.search import SearchVectorField
 from django.db import models
 from django.db.models.query import QuerySet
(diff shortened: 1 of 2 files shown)


jacobdgm:
oops! a Chant string representation improvement is tagging along on this pull request, it seems!

django/cantusdb_project/main_app/models/chant.py
```diff
@@ -1,3 +1,4 @@
+from this import d
```
jwang44: artifacts from auto-import

jwang44 (approved):

PR #176, opened 2022-06-14
implemented log in and permissions requirements on all protected routes

### LINKED ISSUE #174: check for login and permissions on all protected routes
For protected routes, we first have to check if they're logged in. If they're not logged in, redirect to log in page. Otherwise, check if the logged in user is authorized/has permissions to the requested route.

jiwoojeongjj:
fixes #174 

LoginRequiredMixin and UserPassesTestMixin have been implemented for the following views:

- ChantCreateView
- ChantDeleteView
- ChantEditVolpianoView
- SequenceEditView
- SourceCreateView
- SourceEditView

The privileges are as follows:

- "Create" views can only be accessed by contributors and project managers
- "Edit" and "Delete" views can only be accessed by editors and proofreaders who were granted access to the source, contributors who created the source, and project managers

### **IMPORTANT (PLEASE READ)**
This should be the workflow for CantusDB:

1. In CantusDB's admin interface, the following groups should be added:

- "contributor"
- "editor"
- "proofreader"
- "project manager"

**It's important that the groups are named exactly like above (else, we would have to modify the code accordingly)**

2. When a new user is created, the administrator should assign the user to one or more of the groups listed above, as well as designate sources that they are able to make changes to.

### DIFF
--- a/django/cantusdb_project/main_app/views/sequence.py
+++ b/django/cantusdb_project/main_app/views/sequence.py
@@ -6,8 +6,6 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib import messages
 from django.contrib.auth.mixins import UserPassesTestMixin
 
-from main_app.models.source import Source
-
 
 class SequenceDetailView(DetailView):
     """
@@ -71,7 +69,7 @@ class SequenceEditView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
         sequence_id = self.kwargs.get(self.pk_url_kwarg)
         sequence = Sequence.objects.get(id=sequence_id)
         # find the source of this sequence
-        source = Source.objects.get(sequence=sequence)
+        source = sequence.source
         # checks if the user is an editor or a proofreader,
         # and if the user is given privilege to edit this source and thus, it's sequences
         is_editor_proofreader = user.groups.filter(Q(name="editor")|Q(name="proofreader")).exists()
--- a/django/cantusdb_project/main_app/views/sequence.py
+++ b/django/cantusdb_project/main_app/views/sequence.py
@@ -4,6 +4,9 @@ from django.db.models import Q
 from main_app.forms import SequenceEditForm
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib import messages
+from django.contrib.auth.mixins import UserPassesTestMixin
+
+from main_app.models.source import Source
 
 
 class SequenceDetailView(DetailView):
@@ -50,7 +53,7 @@ class SequenceListView(ListView):
 
         return queryset.filter(q_obj_filter).order_by("siglum", "sequence")
 
-class SequenceEditView(LoginRequiredMixin, UpdateView):
+class SequenceEditView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
     template_name = "sequence_edit.html"
     model = Sequence
     form_class = SequenceEditForm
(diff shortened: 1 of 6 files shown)


django/cantusdb_project/main_app/views/sequence.py
```diff
@@ -62,3 +65,27 @@ def form_valid(self, form):
             "Sequence updated successfully!",
         )
         return super().form_valid(form)
+
+    def test_func(self):
+        user = self.request.user
+        sequence_id = self.kwargs.get(self.pk_url_kwarg)
+        sequence = Sequence.objects.get(id=sequence_id)
+        # find the source of this sequence
+        source = Source.objects.get(sequence=sequence)
```
jwang44: Does this work? Looks like you're trying to do `source=sequence.source`.

django/cantusdb_project/users/models.py
```diff
@@ -7,6 +7,7 @@ class User(AbstractUser):
     city = models.CharField(max_length=255, blank=True, null=True)
     country = models.CharField(max_length=255, blank=True, null=True)
     website = models.URLField(blank=True, null=True)
+    sources_user_can_edit = models.ManyToManyField("main_app.Source", related_name="sources_user_can_edit", blank=True)
```
jwang44: I think the `related_name` attribute is used when you try to access the users from the source. 
To get the sources that a user can edit: `user.sources_user_can_edit.all()`.
To get the users who can edit a certain source: `source.users_who_can_edit_this_source.all()`.
In this example, `users_who_can_edit_this_source` is the `related_name`.
See doc here: https://docs.djangoproject.com/en/4.0/topics/db/queries/#following-relationships-backward

django/cantusdb_project/main_app/views/sequence.py
jiwoojeongjj: Yes it works but, I guess that `source=sequence.source` would be cleaner (I forgot that sequence had a source field...). I don't see any problems with `source=sequence.source` because you're just assigning `sequence.source`, a `Source` object, to a variable called `source`

django/cantusdb_project/users/models.py
jiwoojeongjj: Thanks, I'll make sure to fix it accordingly

django/cantusdb_project/main_app/views/sequence.py
jwang44: I'm confused by this because the source model doesn't have a `sequence` field.

jiwoojeongjj: Ah, sorry. I understand what you were trying to say. Yes, this is a clear oversight from my part, but I believe it was working when I was testing it out with various accounts with different privileges. In hindsight, I don't know why it worked or maybe I tested it out on too small of a sample and it happened to work for those. I'll make sure to fix it with your suggestion.

jwang44: If it didn't throw an error, perhaps the `test_func` is not called by anything. Please test more to make sure it does work as expected.

jiwoojeongjj: **Update**: As talked about with Junhao, Django does not seem to throw an error when we erroneously call `source = Source.objects.get(sequence=sequence)`, even though `sequence` is not defined as a field in the Source model. By printing out the object assigned to `source`, I was able to see that it was correctly assigned the Source object I was expecting. Perhaps this is a feature that I am unaware of for the `.get()` method. However, I have changed `Source.objects.get(sequence=sequence)` and `Source.objects.get(chant=chant)` to `sequence.source` and `chant.source` respectively, for clarity.

PR #186, opened 2022-06-20
create abstract model BaseChant from which Chant and Sequence can inherit their fields

jacobdgm:


jacobdgm:
it already passed `test_models` before the pull request was opened; this comment is to confirm it also passes `test_views_general`.

PR #187, opened 2022-06-22
Add JsonNextChantsTest, and refactor a bit

jacobdgm:
this PR is mostly to add a JsonNextChantsTest, but there are a few things that are refactored, both related to and unrelated the implementation of the test. Commit messages explain them.


django/cantusdb_project/main_app/tests/test_views_general.py
```diff
@@ -1,6 +1,9 @@
+from struct import unpack
```
jwang44, line 1: I'll merge this for now, but make sure to remove this on next commit

PR #198, opened 2022-06-30
Migrate users

### LINKED ISSUE #196: migrate users from old to new Cantus

### LINKED ISSUE #197: deleting a user should delete the sources + chants that they have created
likewise, deleting a source should delete the chants inside that source

jiwoojeongjj:
fixes #196 and fixes #197 

1. run `python old_users_list.py`. This will scrape data from the old Cantus and make a new detailed csv.
2. run `python manage.py add_all_users`. This will add all users specified in the newly created csv.

**NOTE: 

Before running the `add_all_users` command, you would have to delete all users that have been created thus far. After running the command, you can run `python manage.py createsuperuser` to create an admin account.**


django/cantusdb_project/old_users_list.py
```diff
@@ -0,0 +1,67 @@
+import csv
+from itertools import count
```
jwang44, line 2: Unused import

jwang44 (approved):

PR #201, opened 2022-07-05
update user permissions + update corresponding test cases

jiwoojeongjj:
updated user permissions according to the specifications in the following spreadsheet: 
https://docs.google.com/spreadsheets/d/1v-9IwsR6JBWFDaHaJlEzfDvZzK7W3fl13kLldOdmu0I/edit#gid=1535637058

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views_general.py
+++ b/django/cantusdb_project/main_app/tests/test_views_general.py
@@ -1487,6 +1487,13 @@ class PermissionsTest(TestCase):
         contributor.user_set.add(self.user)
         self.client.login(username='user', password='pass')
 
+        # a source assigned to the current user
+        assigned_source = make_fake_source()
+        self.user.sources_user_can_edit.add(assigned_source)
+        for i in range(5):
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/tests/test_views_general.py
```diff
@@ -1487,6 +1487,13 @@ def test_permissions_contributor(self):
         contributor.user_set.add(self.user)
         self.client.login(username='user', password='pass')
 
+        # a source assigned to the current user
+        assigned_source = make_fake_source()
+        self.user.sources_user_can_edit.add(assigned_source)
+        for i in range(5):
+            Chant.objects.create(source = assigned_source)
```
jwang44: Could you remove the spaces around the equal sign? I think it's a general rule to use spaces when assigning values like `a = 0` and not to use spaces when passing values to arguments like `Chant.objects.create(source=assigned_source)`
https://peps.python.org/pep-0008/#other-recommendations

jwang44 (approved):

PR #206, opened 2022-07-06
Source forms

jiwoojeongjj:
This PR allows the workflow where when a user creates/edits a source and modifies the current editors field, the selected users will be assigned to that source automatically.

**note**: sorry, I seemed to have created a branch in the wrong place, the first two commits are the same commits from PR #204


django/cantusdb_project/main_app/forms.py
```diff
@@ -209,12 +209,18 @@ class Meta:
     century.widget.attrs.update({"class": "form-control custom-select custom-select-sm"})
 
     current_editors = forms.ModelMultipleChoiceField(
-        queryset=get_user_model().objects.all().order_by("last_name"), required=False
+        queryset=get_user_model().objects.filter(
```
jwang44, line 212: If we do `project_manager OR contributor OR Editor`, perhaps we don't need the filtering? That's just all users.

django/cantusdb_project/main_app/views/source.py
```diff
@@ -292,3 +297,16 @@ def test_func(self):
...
         else:
             return False
+
+    def form_valid(self, form):
+        source = form.save()
+        current_editors = source.current_editors.all()
+        assigned_users = source.users_who_can_edit_this_source.all()
+
+        for user in assigned_users:
+            user.sources_user_can_edit.remove(source)
```
jwang44, line 307: Why are users removed from the source? Is this a one-off thing where `users_who_can_edit_this_source` is replaced by `current_editors`?

django/cantusdb_project/main_app/forms.py
jiwoojeongjj: Will there never come a time when a user is given an account and not assigned to any group?

django/cantusdb_project/main_app/views/source.py
jiwoojeongjj: When a new source is created with users A B C assigned as the current editors, the source object will have A B C in the `current_editors` field and the users will have the newly created source in their `sources_user_can_edit field`.

When the current editors field is updated to B C, the source object's `current_editors` field will update. However, user A will still have the source in their `sources_user_can_edit` field.

A quick and easy solution would be to first remove this source from all users' `sources_user_can_edit` and re-assign them to users accordingly.

django/cantusdb_project/main_app/forms.py
jiwoojeongjj: Also, I did the filtering in case additional groups may be added in the future. If these two cases aren't likely to happen, I'll update accordingly.

jwang44: For now these two cases are unlikely to happen, but this makes sense. We can keep it like this.

django/cantusdb_project/main_app/views/source.py
jwang44: Makes sense. Thanks!

PR #231, opened 2022-07-28
Content overview

jiwoojeongjj:
This PR implements the content overview and makes necessary changes to the `BaseModel` (added `created_by` and `last_updated_by` fields).


django/cantusdb_project/main_app/admin.py
```diff
@@ -1,22 +1,60 @@
+from lib2to3.pytree import Base
```
jwang44: Looks like an unused auto-import

django/cantusdb_project/main_app/admin.py
```diff
@@ -1,22 +1,60 @@
...
+    pass
+
+class SequenceAdmin(BaseModelAdmin):
+    pass
+
 class SourcesUserCanEditInline(admin.TabularInline):
     model = get_user_model().sources_user_can_edit.through
 
-class SourceAdmin(admin.ModelAdmin):
+class SourceAdmin(BaseModelAdmin):
```
jwang44, line 46: Why is `SourceAdmin` using `BaseModelAdmin` while other models are using their custom ones? More inline comments will be very helpful.

django/cantusdb_project/main_app/templates/content_overview.html
```diff
@@ -0,0 +1,80 @@
+{% extends "base.html" %}
+{% load helper_tags %}
+{% block content %}
+<title>Content Overview | Cantus Manuscript Database</title>
+<script src="/static/js/source_list.js"></script>
```
jwang44: Is this js necessary on this page?

django/cantusdb_project/main_app/templates/content_overview.html
```diff
@@ -0,0 +1,80 @@
...
+                            <a href="{{ object.get_absolute_url }}"><b>{{ object.first_name }} {{ object.family_name }}</b></a>
+                        </td>
+                    {% endif %}
+                    <td class="text-wrap" style="text-align:center">{{ object|classname }}</td>
+                    <td class="text-wrap" style="text-align:center">{{ object.date_created|date:'Y-m-d H:i' }}</td>
+                    <td class="text-wrap" style="text-align:center">
+                        {% if object.created_by is None %}
+                            {{ object.created_by }}
+                        {% else %}
+                            <a href="/admin/users/user/{{ object.created_by.id }}/change/">
```
jwang44, line 52: Did you look into ways to avoid hardcoding the urls? Can the `{% url ... %}` tag be used to generate admin urls as well?

django/cantusdb_project/main_app/views/views.py
```diff
@@ -576,4 +585,42 @@ def change_password(request):
         form = PasswordChangeForm(request.user)
     return render(request, 'registration/change_password.html', {
         'form': form
-    })
\ No newline at end of file
+    })
+
+def pm_check(user):
```
jwang44, line 590: Can you add some comments/doc strings indicating the necessity of this function where this function is being used?

django/cantusdb_project/main_app/templatetags/helper_tags.py
```diff
@@ -44,3 +44,7 @@ def source_links():
         options += option_str
 
     return mark_safe(options)
+
+@register.filter
+def classname(obj):
```
jwang44, line 49: Can you add some comments/doc string to clarify why and where this tag is used? I understand what it does since I'm looking at the changed files side by side, but in the future, this might confuse people. Adding some docs is always helpful.

django/cantusdb_project/main_app/admin.py
jiwoojeongjj: I'm not quite sure I understand. Currently, all of the other models' admins, including `SourceAdmin`, are inheriting from `BaseModelAdmin`.

jwang44 (approved):

PR #234, opened 2022-08-02
Update how "current editors" are assigned to sources

jiwoojeongjj:
Currently, users can be assigned to a source in two different ways:
1. assigned as one of the current editors (in source-create and source-edit views)
2. assigned to a source, manually, by the project manager (through the admin interface's Source page)

Desired functionality:
Even if the "current editors" have changed, users who have been assigned to a source by method 2 should not be affected.

With this PR:
If the set of "current editors" for this source has changed, remove this source from the "sources user can edit" set of all users who were previously "current editors", then reassign this source to new "current editors".

This PR also ensures that when a source is created/changed in the admin interface, the selected `current_editors` will be assigned this source, and un-selected `current_editors` will have their assignment withdrawn.


jwang44:
Just to make sure I understand, in situation 2, if a user is assigned to a source (manually, in the admin interface) by the project manager, is that user added to `source.current_editors`?

PR #235, opened 2022-08-03
fix search bar

### LINKED ISSUE #230: search (incipit) box
On OldCantus, the search box in the corner seems to search incipits for a string, so one gets e.g. 
image

On NewCantus we seems to be searching "contains" on the whole chant, which means it takes longer and probably the chant you wanted won't show up:
image

jacobdgm:
fixes #230 

Up until now, ajax_search_bar used `.isalpha()` as a negative test to determine whether or not the search term might be a Cantus ID. But `.isalpha` returns `False` as soon as there is a space in the search term, and that seems to break things. Switching the test to use `.isdigit()`  as a positive test causes the search box to behave as expected.

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -429,13 +429,13 @@ def ajax_search_bar(request, search_term):
     # load only the first seven chants
     CHANT_CNT = 7
 
-    if search_term.isdigit():
-        # if the search term contains only digits, search Cantus ID
+    if any(char.isdigit() for char in search_term):
+        # if the search term contains at least one digit, assume user is searching by Cantus ID
         chants = Chant.objects.filter(cantus_id__istartswith=search_term).order_by(
             "id"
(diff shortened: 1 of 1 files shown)


jwang44:
I just did some testing and one thing came to mind. Cantus ID may not necessarily be all digits. Some Cantus IDs have letters in them. Incipit must be pure letters, but Cantus ID may not be pure digits.

jacobdgm:
most recent commit changes logic so that, if there is at least one digit in the search term, it will search for a Cantus ID rather than for an incipit.

It might make sense to search both by cantus ID and by incipit right off the bat, since there are a number of chants with a Cantus ID that begins with "Lamen2:". But OldCantus seems to use the same logic that this commit uses: it shows a bunch of chants whose incipits contain "lamen" before the "2" is added, and only shows chants with "Lamen2:" cantus IDs once the "2" has been added.

jwang44 (approved):

PR #259, opened 2022-08-08
add url to 'Add News' link in footer

### LINKED ISSUE #258: "Add News" link broken
In the footer, the Add News link doesn't point to a URL. it should point to the page in the admin panel where users can create news articles.

jacobdgm:
fixes #258

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -325,7 +325,7 @@
                                 {% endif %}
                                 {% if request.user|has_group:"project manager" %}
                                     <li class="list-group-item p-0">
-                                        <a href="">Add News</a>
+                                        <a href="admin/articles/article/add/">Add News</a>
                                     </li>
                                     <li class="list-group-item p-0">
                                         <a href="/admin/main_app/feast/add/">Add Feast</a>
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/templates/base.html
```diff
@@ -325,7 +325,7 @@ <h5>Contributor Navigation</h5>
                                 {% endif %}
                                 {% if request.user|has_group:"project manager" %}
                                     <li class="list-group-item p-0">
-                                        <a href="">Add News</a>
+                                        <a href="/admin/articles/article/add/">Add News</a>
```
jwang44, line 328: Is it possible to avoid hard-coding this url? Does `{% url ... %}` work?

jacobdgm: I'm not certain, but I think that to link to the Admin panel, we have to hard-code it. I think if we had an article-create view, we could do it with `{% url ... %}`, but we don't so we can't.

jacobdgm: this is the same format we use a couple of lines later, for "Add Feast", "Add Indexer", etc.

jwang44: Can you look into this? https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#reversing-admin-urls
If possible, we should replace all hard-coded urls to Django admin interface with url tags.

jwang44: Merging for now. Replacing the urls later

PR #271, opened 2022-08-09
Add "My sources" panel to homepage

### LINKED ISSUE #256: "My sources" panel should appear on homepage
In OldCantus on the homepage when a user is logged in, a list of sources appears in a "My sources" box in the left sidebar. This box replaces the "jump to source" and "what's new" boxes that are visible when a user is not logged in.

In NewCantus, the "jump to source" and "what's new" boxes appear whether or not a user is logged in. Behavior on NewCantus should be changed to match that of OldCantus

jacobdgm:
fixes #256

I'm not super happy with the formatting (the line spacing is too large imo), so maybe this can be revised eventually. In any case, all the necessary links are present.

If anyone else wants to take a look with me at alternatives to `simple_tag`s for adding content to the homepage, please do let me know - the approach of building a bunch of html tags in python as one long string feels super hacky, and there is almost certainly a way that would allow us to do it in DTL.

### DIFF
--- a/django/cantusdb_project/templates/flatpages/default.html
+++ b/django/cantusdb_project/templates/flatpages/default.html
@@ -32,16 +32,17 @@
                     }
                 </script>
             </div>
-            <div class="card mt-3 w-100">
-                <div class="card-header">
-                    What's New
-                </div>
-                <div class="card-body">
-                    <ul>
+
+            {% if request.user.is_anonymous %}
+                <div class="card mt-3 w-100">
+                    <div class="card-header">
+                        What's New
+                    </div>
+                    <div class="card-body">
                         {% recent_articles %}
-                    </ul>
+                    </div>
                 </div>
-            </div>
+            {% endif %}
         </div>
     </div>
 </div>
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/templates/flatpages/default.html
```diff
@@ -32,16 +32,29 @@ <h2>{{ flatpage.title }}</h2>
...
+                    </div>
+                    <div class="card-body">
                         {% recent_articles %}
-                    </ul>
+                    </div>
                 </div>
-            </div>
+            {% endif %}
+
+            {% if not request.user.is_anonymous %}
```
jwang44: Django template language supports `{% if ... %}`, `{% else %}`, `{% endif %}`. That could make the code cleaner.

jacobdgm: very good - done!

jacobdgm:
update - Dylan and I went over the code together, and decided that my implementation is not as unreasonable as I initially thought - creating custom tags would likely create just as much complexity as the current implementation. So I'm comfortable letting the code stand as-is.

jwang44 (approved):

PR #276, opened 2022-08-10
add 'position' as search field for chant-search

### LINKED ISSUE #265: Add "position" as a search field for chant-search
We want to be able to search for entries in the Position field. It's a specific type of chant that our user is needing to isolate, but I think with the combination of search terms already on that page plus "Position" they should be able to get what they need.

TLDR: Add `position` as a search field on the chant-search page.

jacobdgm:
fixes #265

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1,3 +1,4 @@
+from turtle import position
 import lxml.html as lh
 import requests
 import urllib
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1,3 +1,4 @@
+from turtle import position
```
jwang44: wayward import

jacobdgm: personally, I think we should do away with the position field and just use the one from `turtle`

jacobdgm: fine. done.

jwang44 (approved):

PR #286, opened 2022-08-15
Feast and Genre Detail Views: filter sources by published status if user is anonymous

jacobdgm:


django/cantusdb_project/main_app/views/feast.py
```diff
@@ -30,8 +30,15 @@ class FeastDetailView(DetailView):
     template_name = "feast_detail.html"
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        chants_in_feast = self.get_object().chant_set.filter(source__published=True)
+        context = super().get_context_data(**kwargs) 
+        print(context)
```
jwang44: Please remove this

jacobdgm: `context` is still used later on, but I removed the print statement. Thanks for catching this.

PR #293, opened 2022-08-16
In functions in views/views.py, filter results based on whether source is published

jacobdgm:
Functions in views/views.py that returns items from multiple sources now filter those results based on whether the source they're associated with is published. For ajax functions, functions only display results from unpublished sources if the user is logged in. for json APIs, functions never display results from unpublished sources.

One function, json_node_export, does not filter results but rather delivers a Not Found http response for items associated with unpublished sources.

the json_nextchants API still requires work - it will be included in a future pull request (since it will involve reworking next_chants.py)

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -127,6 +127,10 @@ def ajax_melody_list(request, cantus_id):
         Chant.objects.filter(cantus_id=cantus_id).exclude(volpiano=None).order_by("id")
     )
 
+    display_unpublished = request.user.is_authenticated
+    if not display_unpublished:
+        chants = chants.filter(source__published=True)
+    
     # queryset(list of dictionaries)
     concordance_values = chants.values(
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/views.py
```diff
@@ -433,12 +443,17 @@ def ajax_search_bar(request, search_term):
...
-        chants = Chant.objects.filter(incipit__icontains=search_term).order_by("id")[
-            :CHANT_CNT
-        ]
+        chants = Chant.objects.filter(incipit__icontains=search_term).order_by("id")
+
+    display_unpublished = request.user.is_authenticated
+    if not display_unpublished:
+        chants = chants.filter(source__published=True)
+    
+    chants = chants[:CHANT_CNT]
```
jacobdgm, line 455: slicing has moved here since one cannot filter a query once it has been sliced

jwang44 (approved):

PR #296, opened 2022-08-17
Set up next_chants function to filter chants based on whether or not they come from published sources

jacobdgm:
This pull request deals with the next_chants function. The function now has two modes, controlled with the `display_unpublished` boolean flag. If display_unpublished is False, next_chants filters results based on whether the chant comes from a published source.

The json_nextchants API calls next_chants with display_unpublished set to False, while the ChantCreate view calls it with the flag set to True.

### DIFF
--- a/django/cantusdb_project/next_chants.py
+++ b/django/cantusdb_project/next_chants.py
@@ -2,7 +2,7 @@ from main_app.models import Chant
 from collections import Counter
 
 
-def next_chants(cantus_id):
+def next_chants(cantus_id, display_unpublished=False):
     """find instances of the chant with the given cantus_id
     in all manuscripts, gather a list of other chants that follow the
     specified chant in those manuscripts, and count how often each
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/next_chants.py
```diff
@@ -14,11 +14,25 @@ def next_chants(cantus_id):
...
-    next_chants = [chant.next_chant for chant in concordances]
+    # if not display_unpublished:
+    #     concordances = concordances.filter(source__published=True)
+    
+    next_chants = [chant.next_chant
+            for chant
+            in concordances
+            if chant.next_chant is not None
+        ]
+    if not display_unpublished:
```
dchiller: Just curious, but is there a reason you did it this way rather than with the commented out lines (17, 18) ? 

Not sure those commented lines are useful now

dchiller (approved):

django/cantusdb_project/next_chants.py
jacobdgm: thanks for catching this! I've changed it to use the version that was commented out. Either approach works (since a chant's next_chant will necessarily have the same source), but this one is simpler.

PR #307, opened 2022-08-27
Update logos and footer layout

jwang44:
This PR updates the images and naming schema for some of the sponsor logos. It also updates the footer layout so that it looks nicer. A preview of what it looks like:
Screen Shot 2022-08-27 at 18 07 22


django/cantusdb_project/templates/base.html
```diff
@@ -275,37 +275,33 @@ <h5>Creative Commons License</h5>
                                 Attribution-NonCommercial-ShareAlike 4.0 International License.</a>
                         </p>
                     </div>
+                    <div class="col-md" id="contact">
+                        <h5>Contact Us</h5>
+                        <a href="{% url 'contact' %}">      
+                            <svg xmlns="http://www.w3.org/2000/svg" width="30" height="30" fill="currentColor" class="bi bi-envelope" viewBox="0 0 16 16">
+                                <path d="M0 4a2 2 0 0 1 2-2h12a2 2 0 0 1 2 2v8a2 2 0 0 1-2 2H2a2 2 0 0 1-2-2V4Zm2-1a1 1 0 0 0-1 1v.217l7 4.2 7-4.2V4a1 1 0 0 0-1-1H2Zm13 2.383-4.708 2.825L15 11.105V5.383Zm-.034 6.876-5.64-3.471L8 9.583l-1.326-.795-5.64 3.47A1 1 0 0 0 2 13h12a1 1 0 0 0 .966-.741ZM1 11.105l4.708-2.897L1 5.383v5.722Z"/>
```
jacobdgm, line 283: It took me a while to figure out what the path element is doing here - that it's drawing the envelope next to the Contact Us link. Could you add a comment to explain this?

annamorphism:
stupid aesthetic question, but is there any way to make the two university logos side-by-side? or do they end up aligning weirdly or something? it bugs me that there is a weird hole in the footer. (This is true of OldCantus too, so hardly urgent...)

jwang44:
Yes. It is possible to do that. We'll fix this.

jwang44:
With https://github.com/DDMAL/CantusDB/commit/b62073c772638669c60873ab22b26cacb3123532, the two logos are now on the same row like this:
Screen Shot 2022-08-29 at 15 16 17

PR #308, opened 2022-08-29
Fixing test_models as well as other issues

jacobdgm:
With this pull request, the test_models test suite now runs without errors or failures.

Along the way, I fixed an error that had the possibility of overwriting the name of a source's segment. I also fixed an error that caused the source-create page to display an error message, since the page doesn't include a field to specify whether the source is published or not.

### DIFF
--- a/django/cantusdb_project/main_app/models/source.py
+++ b/django/cantusdb_project/main_app/models/source.py
@@ -23,7 +23,7 @@ class Source(BaseModel):
     # Here in the new Cantus, we only use one field, and there are two levels: published and unpublished.
     # Published sources are available to the public. 
     # Unpublished sources are hidden from the list and cannot be accessed by URL until the user logs in.
-    published = models.BooleanField(blank=False, null=False)
+    published = models.BooleanField(blank=False, null=False, default=False)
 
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/models/source.py
```diff
@@ -23,7 +23,7 @@ class Source(BaseModel):
     # Here in the new Cantus, we only use one field, and there are two levels: published and unpublished.
     # Published sources are available to the public. 
     # Unpublished sources are hidden from the list and cannot be accessed by URL until the user logs in.
-    published = models.BooleanField(blank=False, null=False)
+    published = models.BooleanField(blank=False, null=False, default=False)
```
jacobdgm, line 26: without this, the source-create page is broken. Is there ever a situation where we want a user to be able to create a source that is published=True right off the bat? If so, we should add a boolean field to the source-create page.

jwang44: The sources should be unpublished by default. I don't think we need to allow users to create an immediately-published source.

jwang44 (approved):

PR #313, opened 2022-08-30
Clean up tests

jacobdgm:
This PR cleans up some of the tests:
- adds a unittest.skip decorator to a test that had been failing, a test for which I don't understand what the expected behaviour is meant to be.
- rename test_views -> test_views_old (this one is out-of-date) and test_views_general -> test_views (this one is up-to-date)


jwang44:
The skipped test is meant to test the behaviour after a user makes an edit to the chant through the edit-volpiano page. It doesn't pass because the redirect behaviour is changed. The test assumes a redirect to the chant-detail page (we had it like that sometime ago) but the current view keeps the user at the same page (no redirect).

jwang44:
I'll refactor it and make another PR.

PR #314, opened 2022-08-30
Update redirect test

jwang44:


jwang44:
After changing the asserted redirect path, the test passes as expected.

jacobdgm (approved):

PR #318, opened 2022-09-01
Move tests from test_views_old.py to test_views.py

jacobdgm:
This PR moves all tests from test_views_old.py to test_views.py, adapting them where necessary.


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -64,13 +66,82 @@ def get_random_search_term(target):
     return search_term
 
 
+class ListViewTest(ABC):
```
jwang44: Is this abstract class used anywhere? I don't see it referenced/inherited by other classes?

jacobdgm: this is a good point! I saw several "TODO: maybe write a class to test pagination in all list views" comments as I was working through the tests. It seems this was an effort to write such a class, but it never got plugged into anything. Let me do some experimenting and see whether this can be put to good use.

jacobdgm: spent a while experimenting with this, and it seems like more trouble than it's worth. I removed it for now, and I'll open a `lowpriority` issue in case someone wants to come back to it.

jwang44 (approved):

PR #323, opened 2022-09-04
Junhao optimize db

jwang44:
This PR uses `select_related()` to reduce DB queries for Source and Sequence. The model tests are updated to accommodate recent changes in signals and chant counts.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_models.py
+++ b/django/cantusdb_project/main_app/tests/test_models.py
@@ -36,7 +36,8 @@ class CenturyModelTest(TestCase):
 class ChantModelTest(TestCase):
     @classmethod
     def setUpTestData(cls):
-        make_fake_chant()
+        source = make_fake_source()
+        make_fake_chant(source=source)
 
     def test_get_ci_url(self):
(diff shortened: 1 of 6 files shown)


django/cantusdb_project/main_app/tests/test_models.py
```diff
@@ -73,7 +74,7 @@ def test_index_components(self):
 
     def test_get_concordances(self):
         chant = Chant.objects.get(id=1)
-        chant_with_same_cantus_id = Chant.objects.create(cantus_id=chant.cantus_id)
+        chant_with_same_cantus_id = Chant.objects.create(cantus_id=chant.cantus_id, source=chant.source)
```
jacobdgm, line 77: why do we need to specify the source here?

jacobdgm:
This all looks good - go ahead and merge it (with or without making changes based on my one comment).

jwang44:
When a chant gets created, the triggered signal is captured by our newly implemented receiver, and that receiver tries to update the chant/melody count of the chant’s source. An error is thrown if the chant doesn’t have a source.
________________________________

PR #335, opened 2022-09-08
Add login prompt to 403 page if user is not authenticated

### LINKED ISSUE #300: PermissionDenied vs NotAuthenticated, and login prompt on 401 page
In working through https://github.com/DDMAL/CantusDB/issues/172, it has occurred to me that when a user is not logged in and they try to access an object associated with an unpublished source, we have been using PermissionDenied errors. These errors should really be NotAuthenticated errors.

Eventually, when objects from unpublished sources are restricted even to logged in users (based on whether the user has edit permission for that source, or however we end up implementing it), we should raise PermissionDenied errors in those situations.

@dchiller made the perceptive suggestion that on our 401 page (the page that should appear when we raise a NotAuthenticated error) we could include a prompt to log in.

jacobdgm:
fixes #300 

Turns out that getting django to raise a 401 error is not trivial, so this PR simply adds the text to the 403 page if the user is not logged in. In any case, it clarifies the situation where a user tries to visit a bookmarked page while they're logged out, which is the most important one I think.

### DIFF
--- a/django/cantusdb_project/main_app/templates/403.html
+++ b/django/cantusdb_project/main_app/templates/403.html
@@ -8,6 +8,9 @@
             <br>
             <dl>
                 <dd>You are not authorized to access this page.</dd>
+                {% if not user.is_authenticated %}
+                <dd>This could be because you are not currently logged in. If you want, you can try <a href="{% url 'login' %}?next={{ request.path }}">logging in</a>.</dd>
+                {% endif %}
             </dl> 
         </div>
     </div>
--- a/django/cantusdb_project/main_app/templates/403.html
+++ b/django/cantusdb_project/main_app/templates/403.html
@@ -9,7 +9,7 @@
             <dl>
                 <dd>You are not authorized to access this page.</dd>
                 {% if not user.is_authenticated %}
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/templates/403.html
```diff
@@ -8,6 +8,9 @@ <h5>Access denied</h5>
             <br>
             <dl>
                 <dd>You are not authorized to access this page.</dd>
+                {% if not user.is_authenticated %}
+                    <dd>This could be because you are not currently logged in. If you want, you can try <a href="{% url 'login' %}?next={{ request.path }}">logging in</a>.</dd>
```
jwang44: One minor suggestion: Perhaps we can change the wording here. Make it clear that only contributors should log in, so that general visitors who are not contributors (i.e., who don't have an account) don't ask us about registration. Maybe change "if you want" to "if you are a contributor to the database".

jacobdgm: Thank you, good suggestion. I've added another commit addressing this.

PR #336, opened 2022-09-08
Fix bug on content-overview page

jacobdgm:
This PR fixes a bug on the content-overview page, which caused columns to be offset if a row's object didn't have a `title`, `manuscript_full_text_std_spelling`, `name` or `first_name`:

Screen Shot 2022-09-08 at 14 57 52


jacobdgm:
There's still a bug when an object without a `get_absolute_url` method (e.g. a RISM siglum) is among the 50 most recently modified items. I imagine we can ignore this, as such objects are unlikely to be edited.

jwang44 (approved):
Good catch!

PR #339, opened 2022-09-13
Add signals to precompute Chant and Source fields

### LINKED ISSUE #317: Using Django signals for pre-computing and updating expensive fields
Currently in Cantus DB, most things are computed on the fly. Examples include the chants/melodies count on the source list page and various pages that involve "the next chant". These things are quite expensive to compute and pre-computing them will save database queries and make pages load faster for users. 

We can write a script to pre-compute them and save them to the database. Then, using Django signal, we write another script that updates them whenever a relevant change is made.

jacobdgm:
This PR fixes #317, adding several functions to `main_app/signals.py` that pre-compute fields on Chant and Sequence models that are used to optimize page-load times.

It also adds a script (`populate_chant_fields.py`) that can be run when data is imported into the database from OldCantus, which opens each chant and then saves it, thus triggering the receivers in signals.py and making sure all these fields are up-to-date.

In earlier commits within this PR, functions were written to populate Chants' `next_chant` and `is_last_chant_in_feast` fields. These functions were removed in later commits, as it was discovered that they cause the above script to take an inordinately long time to run. In a future pull request, the plan is to change the docker configuration to run, at regular intervals, several existing management commands that populate these fields in a more efficient manner.

### DIFF
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -7,13 +7,39 @@ from django.db.models import Value
 from django.db.models.signals import post_save, post_delete
 from django.dispatch import receiver
 
+import re
+
 from main_app.models import Chant
 from main_app.models import Sequence
 
-
 @receiver(post_save, sender=Chant)
-def update_chant_search_vector(instance, **kwargs):
-    """When saving an instance of Chant, update its search vector field."""
+def on_chant_save(instance, **kwargs):
+    update_chant_search_vector(instance)
+    update_source_chant_count(instance)
+    update_source_melody_count(instance)
+    update_next_chant_fields(instance)
+    update_volpiano_fields(instance)
+
+@receiver(post_delete, sender=Chant)
+def on_chant_delete(instance, **kwargs):
+    update_source_chant_count(instance)
+    update_source_melody_count(instance)
+    update_next_chant_fields(instance)
+
+@receiver(post_save, sender=Sequence)
+def on_sequence_save(instance, **kwargs):
+    update_source_chant_count(instance)
+
+@receiver(post_delete, sender=Sequence)
+def on_sequence_delete(instance, **kwargs):
+    update_source_chant_count(instance)
+
+
+def update_chant_search_vector(instance):
+    """When saving an instance of Chant, update its search vector field.
+    
+    Called in on_chant_save()
+    """
     index_components = instance.index_components()
     pk = instance.pk
     search_vectors = []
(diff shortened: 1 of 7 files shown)


django/cantusdb_project/main_app/signals.py
```diff
@@ -28,32 +53,120 @@ def update_chant_search_vector(instance, **kwargs):
...
+        volpiano_intervals = "".join([str(interval) for interval in intervals])
+        return volpiano_intervals
+
+    if instance.volpiano is None:
+        return
+
+    volpiano_notes = generate_volpiano_notes(
+        instance.volpiano
+    )
+    Chant.objects.filter(id=instance.id).update(
```
jwang44: This should be `Chant.objects.get(id=instance.id)` because there would only be one chant object with this id. 
Now that I think about it, why not just do `instance.update()`?

django/cantusdb_project/main_app/signals.py
```diff
@@ -28,32 +53,120 @@ def update_chant_search_vector(instance, **kwargs):
...
+        volpiano_notes=volpiano_notes
+    )
+
+    volpiano_intervals = generate_volpiano_intervals(
+        volpiano_notes
+    )
+    Chant.objects.filter(id=instance.id).update(
```
jwang44, line 166: Same as above

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1045,16 +1045,23 @@ def test_template_used(self):
...
-
-    def test_next_chant_signal(self):
+    
+    def test_volpiano_signal(self):
         source = make_fake_source()
-        chant_1 = make_fake_chant(source=source, folio="001r", sequence_number=1)
-        response = self.client.post(
+        self.client.post(
+            reverse("chant-create", args=[source.id]), 
+            {"manuscript_full_text_std_spelling": "ut queant lactose", "folio": "001r", "sequence_number": "1", "volpiano": "9abcdefg)A-B1C2D3E4F5G67?. yiz"}
```
jwang44: Why this particular Volpiano?

django/cantusdb_project/main_app/signals.py
jacobdgm: I copied `Chant.objects.filter().update()` from elsewhere in the codebase. It turns out that if you use `.get().update()` or just `instance.update()`, python raises `AttributeError: 'Chant' object has no attribute 'update'`.

That said, this is a good place for a simplification/refactoring - you can update multiple fields at once with a single `.update()`. I'll push an update that makes this change presently.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: An excellent question! I annotated this in one of the commits I just pushed.

jacobdgm:
I just pushed a couple of additional commits - several of the tests in test_models were failing. They're fixed now.

jwang44 (approved):

PR #341, opened 2022-09-18
Merge indexer user

jwang44:
This PR implements everything related to merging the `Indexer` model into the `User` model. It is to be merged into a separate branch from `develop`, so that it doesn't contaminate the `develop` branch (which is currently serving as production) .


django/cantusdb_project/main_app/views/user.py
```diff
@@ -74,6 +76,28 @@ class UserListView(SearchableListMixin, ListView):
...
+    based on the fields defined in `search_fields` with the `icontains` lookup.
+
+    Accessed by /indexers/
+    """
+    template_name = "indexer_list.html"
+    context_object_name = "indexers"
+
+    def get_queryset(self):
+        all_users = super().get_queryset()
+        indexers = all_users.filter(show_in_list=True)
```
jacobdgm: It would be nice if we could come up with a more descriptive variable than `show_in_list`. It's succinct, but a bit vague. `published` would bring it into alignment with sources/chants/etc., but users aren't exactly published. `publicly_visible` maybe?

django/cantusdb_project/main_app/management/commands/merge_indexer_user.py
```diff
@@ -20,7 +20,10 @@ def handle(self, *args, **options):
             # do not use the existing "name" property, 
             # because for some reason, the first/last names can have trailing spaces
             # TODO: populate the full_name field and remove the "name" property
```
jacobdgm: remove this comment? (perhaps you do it in a future commit)

django/cantusdb_project/main_app/models/source.py
```diff
@@ -62,23 +62,40 @@ class Source(BaseModel):
...
-        "Indexer", related_name="entered_melody_for_sources", blank=True, null=True,
+        "Indexer", related_name="entered_melody_for_sources", blank=True
     )
-    proofreaders = models.ManyToManyField("Indexer", related_name="proofread_sources", blank=True, null=True,)
-    other_editors = models.ManyToManyField("Indexer", related_name="edited_sources", blank=True, null=True,)
+    proofreaders = models.ManyToManyField("Indexer", related_name="proofread_sources", blank=True)
+    other_editors = models.ManyToManyField("Indexer", related_name="edited_sources", blank=True)
+    
+    # replace indexer-related fields with the following user-related fields
+    inventoried_by_u = models.ManyToManyField(
```
jacobdgm: why the `_u` appended to all the field names? Is this a thing to make migrations work?

django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -210,33 +210,31 @@ <h4>{{ source.siglum }}</h4>
...
-                                            {{ editor.family_name }}</b></a><br>
+                                        <a href={{ editor.get_absolute_url }}><b>{{ editor.full_name }}</b></a><br>
                                     </li>
                                 {% endfor %}
                             </ul>
                         {% endif %}
                         {{ source.indexing_notes|default_if_none:"" }}
                         <br>
-                        Contributor: <a href="">CANTUS Database Administrator</a>
+                        Contributor: <a href=""></a>
```
jacobdgm, line 237: what is this for?

django/cantusdb_project/main_app/management/commands/merge_indexer_user.py
jwang44: This is still a TODO. I need to change everything else that depends on this `name` property. I think using the `full_name` field instead is better.

django/cantusdb_project/main_app/models/source.py
jwang44: Are you going commit by commit? These fields with `_u` at the end are for migrations purpose. For example, the `inventoried_by` field (pointing to Indexer) was used to populate the `inventoried_by_u` field (pointing to User). I remember I removed these fields in the last few commits of this PR.

django/cantusdb_project/main_app/templates/source_detail.html
jwang44: I couldn't find the `contributor` information in old Cantus json-export, put in a static string as a place holder, and forgot about it. Now I saw it again, I removed the static string so that we will know the ` contributor` field is still unknown.

django/cantusdb_project/main_app/views/user.py
jwang44: I'll explain it here and please feel free to give suggestions. All users by default have this `show_in_list` set to `False`. Those who has an Indexer record on the old Cantus will have this set to `True` so they are displayed in the indexer-list page. So this `show_in_list` field indicates whether the user was listed on old Cantus as an Indexer. Not quite the same as `published` or `unpublished`.

jwang44: Perhaps `is_indexer` or `show_as_indexer` will be a better name for the field?

jacobdgm: I like `is_indexer`!

django/cantusdb_project/main_app/models/source.py
jacobdgm: very good! I thought this (it being temporary, for the purpose of migrations) might be the case.

django/cantusdb_project/main_app/templates/source_detail.html
jacobdgm: hm... so this remains here effectively as a `TODO` for us? Do we actually want it to be visible to website users?

django/cantusdb_project/users/tests.py
```diff
@@ -29,6 +32,107 @@ def test_view(self):
...
+
+    def test_url_and_templates(self):
+        """Test the url and templates used"""
+        response = self.client.get(reverse("indexer-list"))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "base.html")
+        self.assertTemplateUsed(response, "indexer_list.html")
+
+    def test_only_public_indexer_visible(self):
+        """In the indexer list view, only public indexers (those who have at least one published source) should be visible"""
```
jacobdgm, line 58: this is not the way I would have expected things to work, but okay.

django/cantusdb_project/main_app/management/commands/sync_sources.py
```diff
@@ -268,7 +267,8 @@ def get_new_source(source_id):
     # True for published, False for unpublished
     published = status_published and source_status_published
 
-    # current editors are User, the other "people" fields are Indexer
+    # current editors are User, the other "people" fields were originally Indexer,
```
jacobdgm, line 270: this comment is probably unnecessary

django/cantusdb_project/main_app/templates/source_detail.html
jwang44: It is visible on old Cantus, but I can't find where that information is stored.

django/cantusdb_project/users/tests.py
jwang44: This is not ideal for me either, but it is how it works on old Cantus

PR #355, opened 2022-09-27
Add function to signals.py to automatically populate Feast.prefix field

jacobdgm:
This PR adds a function to `signals.py` that automatically populates feasts' `prefix` field when a feast is changed.

In addition, the `prefix` field is not set to `editable=False` - this means that the field is not displayed in the admin panel.

### DIFF
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -178,7 +178,12 @@ def update_volpiano_fields(instance):
 def update_prefix_field(instance):
     pk = instance.pk
 
-    prefix = str(instance.feast_code)[0:2]
-    instance.__class__.objects.filter(pk=pk).update(
-        prefix=prefix
-    )
+    if instance.feast_code:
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/signals.py
```diff
@@ -170,3 +175,10 @@ def generate_volpiano_intervals(volpiano_notes):
         volpiano_intervals=volpiano_intervals,
     )
 
+def update_prefix_field(instance):
+    pk = instance.pk
+
+    prefix = str(instance.feast_code)[0:2]
```
jwang44: Can you check whether this works as expected when a feast gets saved with an empty `feast_code`?

jacobdgm: this is a good point - thanks! I've added some commits, explicitly defining what to do with empty `feast_code`s in the places where we calculate `prefix`.

jwang44 (approved):

PR #383, opened 2022-10-14
Negotiate with bootstrap

### LINKED ISSUE #352: Top menu in Safari
To top menu in Safari, the menu goes beyond  the white background:

Screen Shot 2022-09-26 at 3 56 51 PM

jacobdgm:
This PR rolls bootstrap back from 5.2.1 to 4.6.2, and in so doing re-adds the spacing between the main panel and the right sidebar. It also changes the header so it stretches across the entire top of the screen (thus matching the footer), getting rid of the text-spilling-beyond-navbar-div bug (i.e. fixes #352). (I think the header would look better if it was the same width as the content + sidebar [when the window is wide, the Cantus logo ends up waaay off on the left side of the window], but this at least looks more tidy than it did)

PREVIOUSLY:
Screen Shot 2022-10-14 at 13 40 01

NOW:
Screen Shot 2022-10-14 at 13 39 40

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -117,7 +117,7 @@
     <script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v14.0" nonce="ILOxiQ1O"></script>
     <header>
         <!-- Navbar -->
-        <div class="container-fluid">
+        <div class="container-fluid" style="padding: 0px">
             <nav class="navbar navbar-expand-xl navbar-light bg-white border-top col-12">
                 <a class="navbar-brand" href="/">
                     <img src="{% static "CantusLogoSmall.gif" %}"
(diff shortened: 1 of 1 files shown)


fujinaga:
Does this work for Safari?

jwang44:
This header looks still off. It is not how it used to be. It was the same width as the content, not the footer. See attached  (staging) for the expected layout. 
Screen Shot 2022-10-14 at 21 14 40

jacobdgm:
> Does this work for Safari?

I've been testing primarily with Safari, so yes. I also tried it on Chrome and Firefox

> This header looks still off. It is not how it used to be. It was the same width as the content, not the footer. See attached (staging) for the expected layout.

Acknowledged:

> It also changes the header so it stretches across the entire top of the screen (thus matching the footer), getting rid of the text-spilling-beyond-navbar-div bug (i.e. fixes #352). (I think the header would look better if it was the same width as the content + sidebar [when the window is wide, the Cantus logo ends up waaay off on the left side of the window], but this at least looks more tidy than it did)

I settled for something that's less ideal, thinking that if we had something that at least looked fairly neat and tidy, it would be good enough for now and we could move on to improving the site's functionality. Maybe in the future we can get someone who's good at frontend to sort it out. If you (Junhao) want to spend some time figuring out how to make this work, I'd welcome that. But I don't think it's an effective use of my own time.

fujinaga:
Ok, you can close this for now.

PR #401, opened 2022-10-25
fix logic to check whether gregorien.info page is valid

jacobdgm:
This PR fixes the logic in the concordances section on the chant detail page - whereas previously (and erroneously), the gregorien.info link only displayed if the CI json-con API returned a 200 status code, it now (correctly) checks whether the relevant gregorien.info link returns a 200.

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -348,7 +348,6 @@ class ChantDetailView(DetailView):
                         </td>
                     </tr>
                 """
-                print(gregorien_response)
 
             if concordances_count:
                 context["concordances_summary"] += "</table><br>"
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -348,6 +348,7 @@ def get_context_data(self, **kwargs):
                         </td>
                     </tr>
                 """
+                print(gregorien_response)
```
jwang44: You probably want to remove this print statement.

ahankinson:
Do you need to do this on every page load? An alternative would be to have a periodic task that stores the status in the db.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: Thanks for catching this! Done!

PR #402, opened 2022-10-26
Ensure unpublished sources do not appear in user detail sources list

jacobdgm:
Fixes oversight in #399

Occasionally, an h5 will appear with no sources beneath it - this can occur when a user, for example, is listed as an "other editor" for one or more sources and every one of those sources is not published. I don't think this is a big issue, but if we'd like to ensure this doesn't happen, I can go back and filter things in the view instead of in the template.


jwang44:
This is fine for now. I'll merge it, but I suggest that this level of logic be implemented in the views instead.

PR #405, opened 2022-10-27
Cache data from remote resources

jacobdgm:
This PR adds

- two new fields: `gregorien_cached` and `ci_concordances_cached`
- two new properties:`has_gregorien_page` and `ci_concordances`, and
- two new methods: `update_gregorien` and `update_ci_concordances`

to aid in quickly rendering the chant detail page.

When a request for the chant detail page is received, the view/template make use of new `has_gregorien_page` and `ci_concordances` properties, allowing the page to be loaded without having to wait for one response from gregorien.info and another from cantusindex.org.

- Previously, in order to render the "concordances" section of the chant detail page, the chant detail view would
  - check for a 200 response from the relevant gregorien.info page before creating the link to gregorien.info, and
  - pull information on concordances from Cantus Index's `json-con` API, waiting for a response from CI before sending context to the template
- Now, the Chant and Sequence models have
  - a `has_gregorien_page` property, which stores the result of the the 200-response check, and
  - a `ci_concordances` property, which stores the data from Cantus Index about concordances for the cantus_id in question.
  - both of these properties do the following:
    1. check to see whether there is a cached value in `gregorien_cached` or `ci_concordances_cached` respectively
    2. if there isn't, fetch that information, save it to `gregorien_cached` or `ci_concordances_cached`, and then return it
    3. if there is, fetch that value, then call the `update_gregorien` method or the `update_ci_concordances` method respectively in the background (i.e. in a new thread), and immediately return the value

The upsides are:

- faster load times for chant detail pages
- if Cantus Index is having a bad day, we can still display concordances

the downsides:

- if lots of changes have been made to CI since someone last visited a particular CantusDB chant detail page, the displayed results will be out-of-date
- more data has to be stored in our database. A future optimization might be to make a model to store information on individual Cantus IDs (the concordances and the gregorien page are both specific to a single cantus ID, so if we have 50 different chants with a given Cantus ID, we end up storing up to 50 copies of the concordances)

---

As this PR adds a couple of new fields to the BaseChant model, be sure to run `makemigrations` and `migrate` after pulling!

--- 

Also, thanks is in order to @ahankinson, whose recent comment on #401 suggested these changes.

### DIFF
--- a/django/cantusdb_project/main_app/models/base_chant.py
+++ b/django/cantusdb_project/main_app/models/base_chant.py
@@ -4,6 +4,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.postgres.search import SearchVectorField
 import threading
 import requests
+import json
 
 class BaseChant(BaseModel):
     """
@@ -149,6 +150,44 @@ class BaseChant(BaseModel):
         else:
             self.gregorien_cached = False
             self.save()
+
+    @property
+    def ci_concordances(self):
+        if self.cantus_id is None:
+            return None
+        if self.ci_concordances_cached is not None:
+            t = threading.Thread(target=self.update_ci_concordances)
+            t.start()
+            return self.ci_concordances_cached
+        else: # ci_concordances_cached is None, i.e. we've not yet stored the concordances
+            # Making a request without `/refresh` at the end pulls cached concordances from CI, which is
+            # faster than calling it with `/refresh`, but also potentially out-of-date.
+            # Since we don't have a stored copy of the concordances, we fetch CI's stored copy (so
+            # we can quickly return), and then call self.update_ci_concordances to make sure our stored
+            # copy is fully up-to-date.
+            response = requests.get(
+                "http://cantusindex.org/json-con/{}".format(self.cantus_id)
+            )
+            concordances = json.loads(response.text[2:])
+            concordance_chants = [c["chant"] for c in concordances]
+            t = threading.Thread(target=self.update_ci_concordances)
+            t.start()
+            return concordance_chants
+    
+    ci_concordances_cached = models.JSONField(blank=True, null=True)
+
+    def update_ci_concordances(self):
+        # making a request with `/refresh` at the end asks CI to update its concordances
+        response = requests.get(
+            "http://cantusindex.org/json-con/{}/refresh".format(self.cantus_id)
+        )
+        concordances = json.loads(response.text[2:])
+        # response.text is an array of JSON objects. Each has a single
+        # key, ["chant"], pointing to a nested object. We only need the
+        # nested object. 
+        concordance_chants = [c["chant"] for c in concordances]
+        self.ci_concordances_cached = concordance_chants
+        self.save()
 
 
     def get_ci_url(self) -> str:
--- a/django/cantusdb_project/main_app/models/base_chant.py
+++ b/django/cantusdb_project/main_app/models/base_chant.py
@@ -2,6 +2,8 @@ from main_app.models import BaseModel
 from django.db import models
 from django.contrib.auth import get_user_model
 from django.contrib.postgres.search import SearchVectorField
+import threading
+import requests
 
 class BaseChant(BaseModel):
     """
@@ -122,6 +124,33 @@ class BaseChant(BaseModel):
     # dact = models.CharField(blank=True, null=True, max_length=64)
     # also a second differentia field
     
+    @property
+    def has_gregorien_page(self):
+        if self.gregorien_cached is not None:
+            # return previously stored value of gregorien_cached, and
+            # also update its value (in a new thread, so we don't have to wait
+            # for gregorien.info's response before returning)
+            t = threading.Thread(target=self.update_gregorien)
+            t.start()
+            return self.gregorien_cached
+        else: # gregorien_cached is None, i.e. we've never checked to see whether this chant has a gregorien.info page
+            self.update_gregorien()
+            return self.gregorien_cached
+    
+    gregorien_cached = models.BooleanField(blank=True, null=True)
+
+    def update_gregorien(self):
+        gregorien_response = requests.get(
+                "https://gregorien.info/chant/cid/{}/en".format(self.cantus_id)
+            )
+        if gregorien_response.status_code == 200:
+            self.gregorien_cached = True
+            self.save()
+        else:
+            self.gregorien_cached = False
+            self.save()
+
+
     def get_ci_url(self) -> str:
         """Construct the url to the entry in Cantus Index correponding to the chant.
 
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -220,43 +220,43 @@
                         </tr>
                     </thead>
                     <tbody>
-                        {% for chant in concordances %}
+                        {% for c in concordances %}
                             <tr>
                                 <td>
-                                    <a href="{{ chant.srclink }}" target="_blank">
-                                    {{ chant.siglum }}
+                                    <a href="{{ c.srclink }}" target="_blank">
+                                    {{ c.siglum }}
                                     </a>
                                 </td>
                                 <td>
-                                    {{ chant.folio }}
+                                    {{ c.folio }}
                                 </td>
                                 <td>
-                                    <a href="{{ chant.chantlink }}" target="_blank">
-                                        {{ chant.incipit }}
+                                    <a href="{{ c.chantlink }}" target="_blank">
+                                        {{ c.incipit }}
                                     </a>
                                 </td>
                                 <td>
-                                    {{ chant.office }}
+                                    {{ c.office }}
                                 </td>
                                 <td>
-                                    {{ chant.genre }}
+                                    {{ c.genre }}
                                 </td>
                                 <td>
-                                    {{ chant.position }}
+                                    {{ c.position }}
                                 </td>
                                 <td>
-                                    {{ chant.feast }}
+                                    {{ c.feast }}
                                 </td>
                                 <td>
-                                    {{ chant.mode }}
+                                    {{ c.mode }}
                                 </td>
                                 <td>
-                                    {% if chant.image %}
-                                        <a href="{{ chant.image }}" target="_blank">Image</a>
+                                    {% if c.image %}
+                                        <a href="{{ c.image }}" target="_blank">Image</a>
                                     {% endif %}
                                 </td>
                                 <td>
-                                    {{ chant.db }}
+                                    {{ c.db }}
                                 </td>
                             </tr>
                         {% endfor %}
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -91,30 +91,14 @@ class ChantDetailView(DetailView):
             context["syllabized_text_with_melody"] = word_zip
 
         # If chant has a cantus ID, Create table of concordances
-        if chant.cantus_id:
-            response = requests.get(
-                "http://cantusindex.org/json-con/{}".format(chant.cantus_id)
-            )
-            concordances = json.loads(response.text[2:])
-            
-            ######################################
-            ### create context["concordances"] ###
-            # (data to be unpacked in chant_detail.html to make concordances table)
-            
-            # response.text is an array of JSON objects. Each has a single
-            # key, ["chant"], pointing to a nested object. We only need the
-            # nested object. 
-            concordance_chants = [c["chant"] for c in concordances] 
-            context["concordances"] = concordance_chants
-
-            ##############################################
-            ### create context["concordances_summary"] ###
-            # (tally of how many concordances come from which databases)
+        concordances = chant.ci_concordances
+        context["concordances"] = concordances
+        if concordances:
             context["concordances_summary"] = ""
 
             # tally from all databases
             concordances_count = len(
-                concordance_chants
+                concordances
             )
             if concordances_count:
                 if concordances_count > 1:
@@ -127,7 +111,7 @@ class ChantDetailView(DetailView):
             # Cantus Database
             cd_count = len(
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "CD"
                 ]
             )
@@ -150,7 +134,7 @@ class ChantDetailView(DetailView):
             # Fontes Cantus Bohemiae
             fcb_count = len(
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "FCB"
                 ]
             )
@@ -173,7 +157,7 @@ class ChantDetailView(DetailView):
             # Medieval Music Manuscripts Online
             mmmo_count = len(
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "MMMO"
                 ]
             )
@@ -196,7 +180,7 @@ class ChantDetailView(DetailView):
             # Portuguese Early Music Database
             pem_count = len(
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "PEM"
                 ]
             )
@@ -219,7 +203,7 @@ class ChantDetailView(DetailView):
             # Spanish Early Music Manuscripts
             semm_count = len( 
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "SEMM"
                 ]
             )
@@ -242,7 +226,7 @@ class ChantDetailView(DetailView):
             # Cantus Planus in Polonia
             cpl_count = len(
                 [
-                    True for c in concordance_chants
+                    True for c in concordances
                     if c["db"] == "CPL"
                 ]
             )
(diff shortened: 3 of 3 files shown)


ahankinson:
An alternate approach, if it's helpful: Create a [Django Admin command](https://docs.djangoproject.com/en/4.1/howto/custom-management-commands/) that gets run periodically, and sets the value based on the most recent response from Gregorien / CI. 

Advantages:
 - It has zero impact on users. At this point, you're still making the first user wait for responses
 - It is always up-to-date. If a reference or concordance has been added, subsequent runs will pick up on this. 
 - Your model code is *much* simpler. You don't need to check various states, etc. -- you just need to set the field value based on the most recent run. 

The Django Admin commands basically give you a way of building custom "shell scripts" that help manage and interact with your web app. It also frees you from being bound to things needing to happen via an HTTP request -- processes can take as long as they need, and they don't tie up the web serving "side" of the application. 

I think there's also a bug here -- if the gregorien is "None" it will run the update and set either True / False. But if a new reference comes online after it's been checked, then the database will never get updated since it never calls out to update the value.

jacobdgm:
Thanks for this feedback! You're suggesting we create an admin command, and then run it periodically with cron (or a similar scheduler), yes? We currently have a few admin/management commands, which we were actually thinking of replacing:

For caching details related to the state of our database (e.g. next- and previous-chant relationships and so on), there are two reasons I think using cron is not the right approach: first, most of our objects don't regularly get changed, so we'd be running a command that does a bunch of computations in order to set the value of a bunch of fields to their previous value. Second, if someone adds or modifies an object, and then checks on something that is due to be recalculated, they will find that the recalculation hasn't happened. This could lead to people thinking that the database is not working correctly.

For caching details related to remote resources (e.g. whether or not gregorien.info has a corresponding page), I think using cron is the wrong approach for one main reason: We have 500,000 chants in the database, and I think that blitzing gregorien.info with 500,000 requests every two weeks is a bad idea.

It might be a good idea to set up a management command to run just once to initially populate the fields, so the first user to visit the page gets a faster load. I'm not sure how it would be best to implement this, though - if would be inconvenient if gregorien.info blocked my IP address after I launched a non-distributed DDoS attack while testing the management command.

jacobdgm:
I'm not sure I see your bug (the spelling-out below is mostly as an exercise to make sure *I* understand what my code is doing. But also I'm thinking of implementing something similar in other places, so better that I don't introduce a bug into a bunch of places.)

Assume `False` is currently stored in `gregorien_cached`, but that gregorien.info has recently been updated to have a page that should be linked to.
Someone requests a view which references `has_gregorien_page`. Start at line 130:
130: `gregorien_cached` is not `None`, it's `False` - proceed to line 134.
134: set up a thread to call `update_gregorien`
135: start the thread
136: return the value of `gregorien_cached`, i.e. return `False`. This is out-of-date, but that's fine - at least we returned a value quickly, and it will be up-to-date the next time someone visits the view. If the user wants up-to-date concordances info, they can reload the page.

in the meantime, a separate thread is now running, which begins at line 143.
144: we try to visit the gregorien.info page.
147: the status code is 200! Proceed to line 148.
148: set `gregorien_cached` to `True`
149: save! now the value is up-to-date for future page loads.

Unless I've missed something, this is the behavior we want - quickly return a value that was accurate the last time we used the value of `has_gregorien_page`, and in the background we make sure our cached value is up-to-date.

ahankinson:
I'm specifically suggesting this for the external resources -- how you cache internal links is up to you! This may also be down to my personal biases about these sorts of things... which I will explain below. I now see that I wasn't reading your code correctly, but that you're actually still doing a request on every page load. I was suggesting this could be avoided with a check once every few days / weeks / months. The data don't actually change that often, so checking on every user request seems like it slows down the user interaction for very little gain. (It was false last time, it's probably going to be false this time...) Nevertheless you do need to check _sometimes_, so having some process that periodically updates the site is necessary.

As for the number of requests, you could optimize that so that it's not so large; for example, you could assume that a Gregorien record will not get deleted very often, so anything with a link now will have a link later, and thus you can skip any record with a value of `True`. You could also do the opposite, and skip those with a value of `False` to see if anything has been deleted, or `None` to only get those that have not been checked yet. You can also set a rate limit (say 5 r/s) and let it run -- it will take as long as it takes, but that doesn't actually matter. 

Right now you're triggering a request on the Gregorien site for every user request on the Cantus site. With a script you can control how many requests-per-second you send to the site so that the traffic is manageable and consistent; with an on-request technique you tie their infrastructure to yours, and require that they be able to handle the same traffic "bursts" as you.  If Cantus starts getting complaints from Gregorien that it's sending too much traffic, you can adjust that in a shell script. But otherwise there's actually no recourse other than to remove the on-request method, since the other option is to limit the number of users. (I know that this is probably not a *big* deal in this case, but as a general architectural principle it's not really that great to tie sites together like that, particularly if you don't control one of them.)

All that's to say that I would rather have a shell script fail, where I can take my time fixing it and the only consequence is that some data doesn't get updated, than have user requests fail or slow down the site and I need to fix it ASAP since it affects the actual function and performance. 

---

A note on my personal biases: I'm *very* conservative about what I do, and what I don't do, during a request. When I write HTTP request handlers, I have two driving factors: How fast can it return, and how easily can it fail?

Speed is the number 1 feature that most users look for in a website -- they don't really notice if it works well, but they certainly do notice if it's slow. Calling out to external APIs at request time adds quite a bit of latency to every request, so I only do it if I absolutely need the latest answer -- and generally then I would prefer to cache a response locally and update in the backend through a separate indexing process. 

For failure, I look at it two ways. 

Exceptions in application servers, depending on how they happen, can take the application down and make everyone grumpy. I get nervous about using libraries  like `threading` at request time, since I'm never 100% sure that I know that I'm handling the exceptions the thread raises correctly, or that all the wrappers around it (WSGI, etc) will know what to do with it either. `requests.get` [can raise exceptions,](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions) which would cause the thread to stop. What happens after that ... I don't know? Will the user know what happened? Will the application keep running? Will the data be consistent? Will anyone ever know if it just keeps failing silently? 

The other failure mode is security. This can be as "simple" as not accepting unsanitized input (not a problem here) but could also require thinking about the effects of DDOS -- either targeted or inadvertent. I've seen bad crawlers get stuck on a page and do thousands of requests as they churn through all sorts of options in a form, resulting in a pretty large traffic spike. In this code, that would generate lots of threads requesting the same thing lots of times to Gregorien as well.

`json.loads` will also raise an exception if the response it gets is not well-formatted JSON. It is now, but if the API on cantus index changes and this page returns a 404, it will break the caching mechanism. (Maybe even silently -- I don't know that anything would ever appear if the thread started raising exceptions? Probably worth testing.)

jacobdgm:
Thanks for taking the time to write this up, @ahankinson! This is really useful.

> With an on-request technique you tie their infrastructure to yours, and require that they be able to handle the same traffic "bursts" as you. If Cantus starts getting complaints from Gregorien that it's sending too much traffic, you can adjust that in a shell script. But otherwise there's actually no recourse other than to remove the on-request method, since the other option is to limit the number of users. (I know that this is probably not a _big_ deal in this case, but as a general architectural principle it's not really that great to tie sites together like that, particularly if you don't control one of them.)

This is a good point. Notably, I'm pretty sure that CantusIndex, and the other sites in the Cantus Network, use the on-request method - when a request is made of `http://cantusindex.org/json-con/<cantus ID>/refresh`, (I gather that) CI sends a request to the `json-cid` API of each of the Cantus Network websites, compiles the results, and then returns. But that's maybe an approach we don't want to emulate on visible, more-likely-to-be-crawled pages.

> I would rather have a shell script fail, where I can take my time fixing it and the only consequence is that some data doesn't get updated, than have user requests fail or slow down the site and I need to fix it ASAP since it affects the actual function and performance.

This is an excellent point.

> Exceptions in application servers, depending on how they happen, can take the application down and make everyone grumpy. I get nervous about using libraries like `threading` at request time, since I'm never 100% sure that I know that I'm handling the exceptions the thread raises correctly, or that all the wrappers around it (WSGI, etc) will know what to do with it either. `requests.get` [can raise exceptions,](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions) which would cause the thread to stop. What happens after that ... I don't know? Will the user know what happened? Will the application keep running? Will the data be consistent? Will anyone ever know if it just keeps failing silently?
>
> ...
> 
> `json.loads` will also raise an exception if the response it gets is not well-formatted JSON. It is now, but if the API on cantus index changes and this page returns a 404, it will break the caching mechanism. (Maybe even silently -- I don't know that anything would ever appear if the thread started raising exceptions? Probably worth testing.)

Another excellent point.

> I've seen bad crawlers get stuck on a page and do thousands of requests as they churn through all sorts of options in a form, resulting in a pretty large traffic spike. In this code, that would generate lots of threads requesting the same thing lots of times to Gregorien as well.

This is a good point. 

---

A feature that I like of the implementation in this PR is that the cached concordances of chants whose pages are visited frequently get updated often, whereas we're not constantly updating the concordances of less-visited chants. Two alternate approaches suggest themselves that preserve this property while addressing the concerns you raised:

1. Whenever a chant's detail page is visited, set a flag on that object in the database indicating it's been visited. In our management command that gets run every so often, only update the concordances of chants whose detail pages have been visited, then set the object's flag back to unvisited.
  - downside: if a user _needs_ up-to-date concordances, they can't get them.
2. On the chant detail page, default to displaying whatever concordances we have cached. But if the url has an `?refresh-concordances=true` query parameter (or something similar), take the time to update our cached concordances before returning a page with the updated concordances. Add a link to the Concordances section with a link to the "refresh" version of the page.
  - downside: there's still the possibility of a crawler getting stuck in a loop on the "refresh" version of the page, and a bunch of requests getting sent off to other databases.
  - upside (kind of) - if an exception is raised, the failure to update the data will at least be noticeable (i.e. the "refresh" version of the page won't load), and if this is causing problems, someone will know to let us know. The non-refresh version of the page should still be accessible.

I'm leaning now towards option 2, which seems fairly transparent to users, and is similar to the behavior of other websites in the Cantus Network.

Please don't feel the need to type up a response to this unless you're feeling really motivated - this is mostly me thinking out loud, so to speak. Again, thank you!

ahankinson:
I hope this doesn't feel like badgering!

I would generally agree with option 2, but for two things:

 - My REST spidey-sense is tingling with the addition of query parameters that act as changes to state on the server. `GET` requests, and query parameters, are generally not meant as control mechanisms for server state, since including them (even relatively harmless ones) makes system security more difficult (e.g., you can generally assume that `GET` requests can be done by anyone, while state-changing requests such as `PUT`, `PATCH` or `POST`, can be secured to require authentication). **Query** parameters are called as such because they're designed to filter things, not as a way of passing actions to the server. I don't want to bore you with a long-winded diatribe, so I'll just leave it there.
 - I might suggest doing the offline caching bit first, and when you do it, print out the number of sources that have been changed in a given run. Run it a couple times manually over the span of a week or so, and see how many records are changed. Use that number to inform your decision on whether you actually need the ability to get the latest ones on-the-fly, or if so few change that waiting between runs would be fine. 

You could also potentially tie the initial cache in to the saving mechanism of the chants so that it caches the response when the record is saved as well, so your editors would be able to see the latest data. Then if you feel it's out-of-date it would just require a simple re-save of the record to fetch the latest result. Just an idea.

jacobdgm:
> I hope this doesn't feel like badgering!

Not at all - I want to get the implementation right before we go to full production mode! My "don't feel the need to type up a response" mostly me trying to respect your time. But I really appreciate the suggestions!

> * My REST spidey-sense is tingling with the addition of query parameters that act as changes to state on the server. `GET` requests, and query parameters, are generally not meant as control mechanisms for server state, since including them (even relatively harmless ones) makes system security more difficult (e.g., you can generally assume that `GET` requests can be done by anyone, while state-changing requests such as `PUT`, `PATCH` or `POST`, can be secured to require authentication). **Query** parameters are called as such because they're designed to filter things, not as a way of passing actions to the server. I don't want to bore you with a long-winded diatribe, so I'll just leave it there.

Thinking about it a bit more, I think it would be approximately as easy to set the URL up to look like `/chant/<id>/refresh` rather than `/chant/<id>?refresh-concordances=true`. This would solve the concern of non-standard use of query parameters, also bringing the implementation even more in line with other parts of Cantus Index, but it would not address the GET-requests-triggering-state-change concern.

> * I might suggest doing the offline caching bit first, and when you do it, print out the number of sources that have been changed in a given run. Run it a couple times manually over the span of a week or so, and see how many records are changed. Use that number to inform your decision on whether you actually need the ability to get the latest ones on-the-fly, or if so few change that waiting between runs would be fine.

This is a good point - it's possible this is all much ado about nothing.

> You could also potentially tie the initial cache in to the saving mechanism of the chants so that it caches the response when the record is saved as well, so your editors would be able to see the latest data. Then if you feel it's out-of-date it would just require a simple re-save of the record to fetch the latest result. Just an idea.

This raises a question best answered by the musicologists, I think. O Musicologi musicologiaeque (@annamorphism)! Do we foresee many non-indexers using the concordances, or do we expect it mostly to be people who are actively adding chants to and/or editing manuscripts?

ahankinson:
> Thinking about it a bit more, I think it would be approximately as easy to set the URL up to look like `/chant/<id>/refresh` rather than `/chant/<id>?refresh-concordances=true`

If you elevate the "verb" ("refresh") to a path in the URI, you are implicitly suggesting, in the semantics of the HTTP "Resource", that loading this URL will return a "refresh" resource, and that resource is a sub-component of `/chant/<id>`. The semantics of URLs / URIs are very specific in the [HTTP spec](https://www.rfc-editor.org/rfc/rfc9110.html#name-resources), and you need to think of the whole cycle of the request -- not only what it will trigger when you enter the URL, but what it will return when it's successful or it fails. (See also the section on ["Safe methods"](https://www.rfc-editor.org/rfc/rfc9110.html#safe.methods) in the HTTP spec as well, since it ties in to my earlier comment about security with GET / PUT / POST, etc.)

I am being a bit pedantic here, of course. Lots of people do this all the time and it's certainly a very convenient way of writing web apps. I'm mostly coming at this thinking ahead to the LinkedMusic grant, where we're going to be using these URLs as URIs in graph databases. Having to distinguish between "resource" URLs and "function" URLs isn't impossible, but it's nice to limit their use as much as possible.

> ...or do we expect it mostly to be people who are actively adding chants ...

I guess I'm suggesting that the initial status of the Gregorien concordance is set when an "indexer" saves the record, but then that's immediately available to "non-indexers" as well. So as soon as the record is published, that value will be available. Since the number of people that are editing records is going to be much fewer than those who are looking at the records, triggering an HTTP request on the chant save would seem like a good place to set the initial state of the cache. Then it's down to the periodic refreshes to keep those initial values up-to-date.

jacobdgm:
https://www.rfc-editor.org/rfc/rfc9110.html#section-9.2.1 says that its
> definition of safe methods does not prevent an implementation from including behavior that is potentially harmful, that is not entirely read-only, or that causes side effects while invoking a safe method. What is important, however, is that the client did not request that additional behavior and cannot be held accountable for it. For example, most servers append request information to access log files at the completion of every response, regardless of the method, and that is considered safe even though the log storage might become full and cause the server to fail.

If we're caching things related to remote resources upon receiving a GET request, does this not fall within "not entirely read-only, [and/or] caus[ing] side effects"? But if we were to include `/refresh` or `?refresh-concordances=true`, it would contradict the specification that "the client... not request that additional behavior and [not] be held accountable for it".

Your point about `chant/<id>/refresh` suggesting that we're returning a "refresh" resource is a good point. So yes, let's not do that.

> `json.loads` will also raise an exception if the response it gets is not well-formatted JSON. It is now, but if the API on cantus index changes and this page returns a 404, it will break the caching mechanism. (Maybe even silently -- I don't know that anything would ever appear if the thread started raising exceptions? Probably worth testing.)

I just ran a small test - an exception in a thread just cause just that thread to stop, without causing exceptions in other processes. Details on the exceptions are printed to the terminal. It might still be worth testing this in a running django project, but it looks very likely that if `json.loads` tries to digest some poorly-formed JSON,
- it won't take the server down with it
- anything following the exception in the threadified function simply won't be executed, and
- details of the exception should appear in the terminal of anyone connected to the django container

Perhaps if we set up a system to log any errors we hit (I'm a bit confused why we don't have any kind of logging set up already) it would be enough to reassure ourselves that updating-our-cached-values-asynchronously-via-threading will not lead us down a path of the caching process failing completely silently.

ahankinson:
Looking at the examples they give, I think the intention is to highlight that some processes may make changes on the server -- such as writing to a log file, or charging an accounting department. The key phrase is "that the client did not request that additional behavior and cannot be held accountable for it" -- I take this to mean that any *specific* requests for behaviours (like the client calling "refresh" to get a refreshed record) falls outside of these exceptions; the client *did* request it, and they *can* be held accountable for it.

ahankinson:
> Details on the exceptions are printed to the terminal. It might still be worth testing this in a running django project ...  should appear in the terminal of anyone connected to the django container

Thinking about this some more, I think the problem with that is that it's only available if a human is connected... You can't catch the exception and send back a status code indicating failure, since the thread is just thrown away on exception. So effectively it will only be noticed if someone connects and happens to see it. I think this means that it can fail silently, and until someone figures out that the IDs aren't updating (days? weeks? months?), only then would it be fixed.

PR #407, opened 2022-10-28
replace contact form with link to email Debra

jacobdgm:
This PR prevents people from using the contact form, and just provides a link to email Debra. This way, no one will press "Submit" and see a "Success" message after no email has been sent, which is the current state of affairs.

### DIFF
--- a/django/cantusdb_project/main_app/templates/contact_form.html
+++ b/django/cantusdb_project/main_app/templates/contact_form.html
@@ -8,58 +8,8 @@
     </object>
 
     <h3>Contact</h3>
-    {% if messages %}
-        {% for message in messages %}
-            {% if 'success' in message.tags %}
-                <div class="alert alert-success alert-dismissible">{{ message }}</div>
-            {% elif 'error' in message.tags %}
-                <div class="alert alert-danger alert-dismissible">{{ message }}</div>
-            {% endif %}
-        {% endfor %}
-    {% endif %}
-
-    <form action="" method="post">
-        {% csrf_token %}
-        {{ form.non_field_errors }}
-        <div class="form-group col-lg-2">
-            {{ form.name.errors }}
-            <label for="{{ form.name.id_for_label }}"><small>Your name</small></label>
-            {{ form.name }}
-        </div>
-        <div class="form-group col-lg-2">
-            {{ form.sender_email.errors }}
-            <label for="{{ form.sender_email.id_for_label }}"><small>Your e-mail address</small></label>
-            {{ form.sender_email }}
-        </div>
-        <div class="form-group col-lg-2">
-            {{ form.subject.errors }}
-            <label for="{{ form.subject.id_for_label }}"><small>Subject</small></label>
-            {{ form.subject }}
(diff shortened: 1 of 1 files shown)


annamorphism:
I worry a little that this will cause bot-spam to Debra's email--but maybe she gets that anyway.

jacobdgm:
@annamorphism yeah, that occurred to me while I've been thinking about the contact page over the past few days. For what it's worth, I have links like this on my personal website with my address and I've not yet noticed spam as a big issue, and Debra has her email address displayed on her profile on uwaterloo.ca anyways.

fujinaga:
> I worry a little that this will cause bot-spam to Debra's email--but maybe she gets that anyway.

Good point. We should have [at] email address.

fujinaga:
@jacobdgm please use “[at]” not “@“.

jacobdgm:
I can do this - I'll roll it into #408. I suppose I should get rid of the `<a href="mailto:...` as well? - If I was writing a spambot crawler, that's the first thing I'd look for.

(I still think this approach looks messy and unprofessional, but if we think this is the best solution, I'll set it up.)

jacobdgm:
I guess I'm not entirely clear what the bigger picture for this is: is displaying Debra's email address on the contact page meant to be a temporary measure, which will be replaced once we get automated emails set up? Or is it meant to be a permanent change?

My opinion is that **permanently doing away with the Contact form** would work better.

Advantages:
- the contact form is the only place where we need to send automated emails. If we permanently got rid of the contact form, we wouldn't need to worry about maintaining an email server.
- It's more transparent to the user - they can style their message if they see fit, and once it's sent, it will show up in their sent messages folder. Personally, I've always found "contact us" forms to feel like black boxes, where you just cross your fingers and hope that the message you sent into the aether actually finds the intended recipient. If I send the email myself, I'm more confident that it's actually going where I mean it to.

Disadvantages:
- Either we sign Debra up to receive some amount of spam, or we have to display a messy-looking obfuscated email address.

Considering the **set-up-the-contact-form-eventually appoach:**

Advantages:
- Our skill-testing question cuts down on some of the spam, and we don't need to display someone's email address

Disadvantages:
- we always need to ensure our automated email system is functioning (e.g. if the contact form is broken, how will we ever find out?)

One step we might take is to email Debra and ask whether she's alright having a link to email her directly that is displayed on the CantusDB contact page, like the one that's on her uwaterloo.ca profile page. She may also have a better sense of how UW's spam filter performs (I don't really remember details of this back from my undergrad...)

So which approach do we prefer? If we're taking the set-up-the-contact-form-eventually path, I'll push the obfuscated email address change to this branch and merge it directly. If we're permanently doing away with the contact form, I'll not apply the [at] for @ substitution in #408 .

fujinaga:
Please contact Debra to see what she prefers.

jacobdgm:
Debra is good with us posting just her email address rather than creating a full form, though she's given us an @dal.ca address in place of the @uwaterloo.ca one. #408 has been updated to incorporate this.

PR #408, opened 2022-10-28
remove contact form and replace with link to email Debra

jacobdgm:
This PR properly implements #407, removing the old code to used to render and process the form. Instead of a contact form with a "Submit" button etc., there is now a link to email Debra.


django/cantusdb_project/main_app/views/views.py
```diff
@@ -255,67 +254,18 @@ def csv_export(request, source_id):
...
 
     Args:
         request (request): The request
 
     Returns:
-        HttpResponse: Render the contact form
+        HttpResponse: Render the contact page
     """
-    # submitted = False
-    if request.method == "GET":
```
dchiller, line 269: @jacobdgm 

I don't understand what these skill_testing_questions are...was the contact form being used for skill testing?

dchiller: It makes sense to me that you wouldn't need this any more, but just checking that there are no dependencies on this functionality.

dchiller (approved):

django/cantusdb_project/main_app/views/views.py
jacobdgm: Thanks for checking about this - this is the solution (i.e. fill in the correct letter or the form submit doesn't work) OldCantus found to the problem most other sites solve using Captcha/Recaptcha. I'm confident it's not used elsewhere.

PR #409, opened 2022-11-01
write tests for get_next_chant(), and fix related bug

jacobdgm:
This PR adds several tests for the Chant model's `.get_next_chant()` method.

In writing them, I found a bug in the implementation of `.get_next_chant()`, which this PR also fixes:
- In situations where one or several pages of the manuscript have been lost, this is usually indicated by a "LACUNA" chant record, with a `sequence_number` of `99` on the last folio before the lacuna. `.get_next_chant` now accounts for gaps in the numbering such as this.

This bug fix is necessary for a future to the `populate_next_chant_fields` admin command, PR forthcoming in the next couple of days.

The code for `.get_next_chant` is starting to become a bit complicated, so it should probably be refactored sometime soon.

### DIFF
--- a/django/cantusdb_project/main_app/models/chant.py
+++ b/django/cantusdb_project/main_app/models/chant.py
@@ -104,20 +104,34 @@ class Chant(BaseChant):
                 folio=self.folio,
                 sequence_number=self.sequence_number + 1,
             )
-        except Chant.DoesNotExist: # i.e. it's the last chant on the folio
-            chants_next_folio = Chant.objects.filter(
-                source=self.source, folio=get_next_folio(self.folio)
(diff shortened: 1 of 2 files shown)


dchiller (approved):
Sent Jacob a comment about creation of potentially unnecessary fake chants in the test functions, but those may be intentional.

jacobdgm:
I think what I was thinking was that there may be weird interactions if there is only one chant on a folio, and the folio doesn't have a set of chants with sequential `sequence_number`s. But I should have made `for` loops to make the intention clearer. In any case, I don't think it's necessary, so I just got rid of the extraneous additional chants - it's simpler/cleaner this way.

PR #411, opened 2022-11-04
Improve Chant Index view

jacobdgm:
This PR makes two changes to the Chant Index view:

First, it changes the value displayed in the siglum column - whereas we had previously been using chants' `siglum` fields, we now use chants' sources' `siglum` fields, which should be standardized and generally more up-to-date.

Second, it changes the sorting of the records on the page. Whereas previously, chants had been ordered by `id`, they are now ordered first by `folio`, and then by `sequence_number`.


dchiller:
This looks good (and makes sense) to me!

Issue #421, opened 2022-11-14
OldCantus has several flavours of chant-search page


jacobdgm:
On OldCantus, the default behavior upon visiting the chant-search page is evidently to search for the specified keyword in only the incipit field, not in the full-text field.

[https://cantus.uwaterloo.ca/search](https://cantus.uwaterloo.ca/search):
![Screen Shot 2022-11-14 at 14 45 23](https://user-images.githubusercontent.com/58090591/201751505-87af4f1a-53fd-4c44-91c3-9d4115f232cc.png)
[https://cantus.uwaterloo.ca/search?op=starts&t=apud&genre=All&field_position_value=&cid=&mode=&feast=&volpiano=All&field_differentia_new_value_op=contains&field_differentia_new_value=](https://cantus.uwaterloo.ca/search?op=starts&t=apud&genre=All&field_position_value=&cid=&mode=&feast=&volpiano=All&field_differentia_new_value_op=contains&field_differentia_new_value=):
![Screen Shot 2022-11-14 at 14 46 36](https://user-images.githubusercontent.com/58090591/201751714-b0d26b7b-c121-4b99-a853-a9d06d720647.png)

Clicking on "Search "apud" in full texts" takes you to a page with a different number of search results
[https://cantus.uwaterloo.ca/search-all?op=contains&t=apud&genre=All&field_position_value=&cid=&mode=&feast=&volpiano=All&field_differentia_new_value_op=contains&field_differentia_new_value=](https://cantus.uwaterloo.ca/search-all?op=contains&t=apud&genre=All&field_position_value=&cid=&mode=&feast=&volpiano=All&field_differentia_new_value_op=contains&field_differentia_new_value=):
![Screen Shot 2022-11-14 at 14 48 17](https://user-images.githubusercontent.com/58090591/201752037-16b387d9-1da6-45b3-bb1f-51b3f47a17b3.png)
(notice how the url has changed from cantus.uwaterloo.ca/search to cantus.uwaterloo.ca/search-all)

If the link at the top is to be believed, searching by incipit is faster than searching by full text. Go figure.

On the main search page (but not on the search-all page), there's a link to a "printer-friendly version", which seems to be be plain html with minimal CSS styling
[...]

PR #423, opened 2022-11-16
Add link to search cantusindex.org to chant search page

jacobdgm:
This PR goes partway towards fixing #421, adding a link to the chant search page taking you to the corresponding search page on cantusindex.org.

Along the way, a missing pluralization template tag was added, the formatting was improved (previously, the pagination blurb had been on the same line as the name of the source when the template was being used to render the ChantSearchManuscript view) and an unnecessary print statement was removed.

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -709,6 +709,19 @@ class ChantSearchView(ListView):
             
         context["url_with_search_params"] = url_with_search_params
 
+        if search_keyword:
+            cantus_index_url = "http://cantusindex.org/search?t={k}".format(k=search_keyword)
+            cantus_index_inner_text = "Search <b>{k}</b> on CantusIndex.org".format(k=search_keyword)
+            cantus_index_title_text = "Search {k} on CantusIndex.org".format(k=search_keyword)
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -708,7 +708,19 @@ def get_context_data(self, **kwargs) -> dict:
             url_with_search_params = current_url + "?"
             
         context["url_with_search_params"] = url_with_search_params
-        print(self.request)
+
+        if search_keyword:
+            cantus_index_url = "http://cantusindex.org/search?t={k}".format(k=search_keyword)
```
dchiller: It could be I'm not understanding the purpose of the changes here, but I'm wondering why you are building html strings here? Could they not be included in an appropriate template?

jacobdgm: This is a good point – for some reason, I was worried about applying the right formatting (i.e. bolding of the search keyword) in the template, so I decided to build the string in the view. But now that I think a little more about it, we have everything we need to do this all in the template. Let me make this change.

jacobdgm: done! thank you for this suggestion!

dchiller: cool, I think this is cleaner and more logical

dchiller (approved):

PR #424, opened 2022-11-16
Add printer-friendly chant search view

jacobdgm:
This PR goes partway towards fixing #421, adding a printer-friendly version of the chant search page. Some notes:

- the existing `chant_search.html` template was duplicated and adapted to create this page. The `ChantSearchView` in `views/chant.py` was similarly copied and adapted
- Some differences in behavior between New and OldCantus:
  - in addition to preserving search parameters, as OldCantus does, this implementation preserves ordering parameters as well. In addition, the column headers can be clicked on to re-order the page just as in the main chant search view.
    - but switching from the printer-friendly page to the normal page and back does not preserve pagination query parameters (this is a similarity between Old and New Cantus)
  - column headers are clickable in this implementation (they weren't in OldCantus's), and display the small arrow icon that indicates what field the results are being ordered by as well as the direction of their ordering
- at some point in the process, I evidently told Serenade to format `urls.py`, so a bunch of lines have been tweaked so as to standardize them. There shouldn't be any substantial difference, only reallocated whitespace

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_search.html
+++ b/django/cantusdb_project/main_app/templates/chant_search.html
@@ -12,26 +12,26 @@
     {% if source %}
         <b>Searching in source: <a href="{{ source.get_absolute_url }}" target="_blank">{{ source.title }}</a></b>
     {% else %}
-        <p>
-            <small>
-                » <a href="{{ printer_friendly_url }}">Printer-friendly version</a>
-        <p>
-            <b>Searching in source: <a href="{{ source.get_absolute_url }}" target="_blank">{{ source.title }}</a></b>
-        </p>
-    {% elif keyword %}
-        <p>
-            <small>
-                » <a href="http://cantusindex.org/search?t={{ keyword }}" title="Search {{ keyword }} on CantusIndex.org" target="_blank">
-                    Search <b>{{ keyword }}</b> on CantusIndex.org
-                </a>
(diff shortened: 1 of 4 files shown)


dchiller (approved):
@jacobdgm The amount of repeated code is setting off a warning lights in my head. Does so much need to be duplicated between the views?

jacobdgm:
yeah, the code duplication is not ideal. The main issue in my head is that the printer-friendly chant search view needs to have none of the bells and whistles present in our `base.html` - the nav items at the top, the footer, the bootstrap formatting, etc. I haven't thought about it too deeply, and perhaps there is an elegant way to accomplish this, but I feel like if one of our views inherits from the other, and only one of them extends `base.html`, any such implementation will be not just complex but complicated. (i.e. the downside of this implementation is all the duplicated code, but the upside is it's uncomplicated)

Let me think about it for a little while - perhaps there is a good way to implement this while avoiding lots of code duplication. An abstract chant-search template that all chant-search templates inherit from?

ahankinson:
Usually you implement page printing with a `print` CSS media query:

https://sympli.io/blog/a-quick-guide-to-css-for-printable-webpages

jacobdgm:
wow, this is a _much_ more elegant solution than both what I came up with and what's currently on OldCantus. Thank you!

PR #427, opened 2022-11-17
Implement sorting in ChantSearchMS view

jacobdgm:
In implementing sorting on the normal ChantSearchView, it caused knock-on effects for the ChantSearchMSView, which shares a template with the normal ChantSearchView. This PR add logic to the ChantSearchMSView allowing the results to be sorted just as on the normal ChantSearchView.

Initially, I attempted to do this by creating a BaseChantSearchView for ChantSearchView and ChantSearchMSView to inherit from; it caused failing tests and error messages, so I went for the copy-and-paste approach. Eventually, this might be a good thing to go back and refactor.


dchiller (approved):
Agreed -- this seems a candidate for further refactoring to reduce risk of unintentional divergence between the two views

Issue #444, opened 2022-12-01
On article detail page, add sidebar with 5 most recent news stories


jacobdgm:
On OldCantus, on every news article page, there is a sidebar listing the 5 most recent articles ([example](https://cantus.uwaterloo.ca/article/710905)). NewCantus does not currently have this sidebar ([example](http://206.12.88.113/article/32)), but probably should.

dchiller:
I was just about to create a duplicate of this...and have one addition (I don't know if this is a new or same issue):

on NewCantus, the fact that the search input field is in the same "box" as the article is odd and different than OldCantus. I think as a user I'd expect a search box in the NewCantus position to search the contents of the article, and not the entire database.

OldCantus:
<img width="1024" alt="image" src="https://github.com/DDMAL/CantusDB/assets/11023634/9fcb9eb8-2906-4868-8ff6-733ebe1f0d76">


NewCantus:

<img width="1129" alt="image" src="https://github.com/DDMAL/CantusDB/assets/11023634/0ec28916-ec0e-411c-b620-06f809d61ca6">

jacobdgm:
Huh - in OldCantus, when a user is logged in, this news articles sidebar isn't visible. NewCantus just uses the logged-in layout whether or not a user is logged in.

OldCantus, logged in:
![Screenshot 2023-06-01 at 2 08 21 PM](https://github.com/DDMAL/CantusDB/assets/58090591/443e620c-d958-4353-88f6-eaf38e3217fc)

We should fix this eventually, but it's still low priority.

PR #446, opened 2022-12-01
Ensure all pages on the site print nicely

jacobdgm:
This PR adds CSS to make all pages of the website display nicely when being printed. This obviates the need for a separate printer-friendly chant search view (see #421).

Lots of files have been changed in this PR, so here's a summary of what has happened:
- CSS has been added to `/templates/base.html` which hides the header, footer, sidebar and global search bar when a page is being printed
- classes have been added to relevant HTML elements in many of the templates:
  - in views with a sidebar, a `sidebar` class has been added to that `div`
  - in views with a search bar:
    - if the search bar is within a sidebar, this was previously implemented with `{% include "global_search_bar.html" %}`, without a containing element. In such instances, this has been wrapped in a `div` with `class="search-bar"`
    - if the search bar is not within a sidebar, it was previously contained in an `object` element. In such instances, a class of `search-bar` has been added to the object element.
    - all instances of `{% include "global_search_bar.html" %}` had a `<!-- global search bar-->` comment above them. With the addition of the `search-bar` class, this comment began to seem superfluous, so I removed it everywhere I saw it.
- also, the article detail and list pages did not include the global search bar. It has been added to both views.
  - The ordering of the elements in the article detail template has also been adjusted slightly to facilitate this change.

All views have been tested in Safari, and some views have been tested in Chrome and Firefox. Some screenshots of Safari, and of the .pdf files generated by Safari's print dialog, are included below for reference:

**Page with a sidebar, default and print:**
with sidebar
with sidebar print

**Page without a sidebar, default and print:**
without sidebar
without sidebar print

**Article detail page, before and after:**
article detail before

article detail after

### DIFF
--- a/django/cantusdb_project/articles/templates/article_detail.html
+++ b/django/cantusdb_project/articles/templates/article_detail.html
@@ -1,11 +1,14 @@
 {% extends "base.html" %}
 {% block content %}
-<div class="mr-3 p-3 col-md-9 bg-white rounded">
+<div class="mr-3 p-3 col-md-12 bg-white rounded">
+    <object align="right" class="search-bar">
+        {% include "global_search_bar.html" %}
+    </object>
+    <h3>
+        {{article.title}}
+    </h3>
     <div class="row">
         <div class="col">
-            <h3>
-                {{article.title}}
-            </h3>
             <div class="container">
                 <div class="container text-wrap">
                     <small>Submitted by <a href="{{ article.author.get_absolute_url }}">{{ article.author }}</a> on {{ article.date_created|date:"D, m/d/Y - H:i" }}</small>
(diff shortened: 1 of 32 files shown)


jacobdgm:
The current version of the article detail view feels a bit inelegant - the text spans too much of the page. Once this is merged, I'll attend to #444 which should help things somewhat.

dchiller (approved):

jacobdgm:
Just doing a bit of testing that I should've done before submitting the PR, I discovered that the side bar and search bar were not labelled (and thus not hidden while printing) on the homepage, so there is an extra commit correcting this oversight. I also simplified the style tag on base.html by combining a couple of the css selectors. Should be good to go now - going to go ahead and merge.

PR #447, opened 2022-12-01
on chant detail page, ensure sequence_number displays even when it is 0

### LINKED ISSUE #440: "Sequence" disappears for value 0
For fragments, we sometimes have the "Sequence" value as 0, meaning it is an incomplete item that begins on the previous (missing) page. This doesn't show up in New Cantus though:
image
 cf oldCantus:
image

I also think it would be better if the folio/sequence items displayed next to each other, as they do in OldCantus. Typically you need both bits of information at once.

jacobdgm:
fixes #440

Also ensures Folio and Sequence columns stay close to each other.

Also changes a couple of other places where there was `if sequence_number` check was occurring.

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1071,7 +1071,7 @@ class ChantCreateView(LoginRequiredMixin, UserPassesTestMixin, CreateView):
             }
         latest_folio = latest_chant.folio if latest_chant.folio else "001r"
         latest_feast = latest_chant.feast.id if latest_chant.feast else ""
-        latest_seq = latest_chant.sequence_number if latest_chant.sequence_number else 0
+        latest_seq = latest_chant.sequence_number if latest_chant.sequence_number is not None else 0 # "is not None": sequence_number is sometimes 0
         latest_image = latest_chant.image_link if latest_chant.image_link else ""
         return {
             "folio": latest_folio,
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1071,7 +1071,7 @@ def get_initial(self):
             }
         latest_folio = latest_chant.folio if latest_chant.folio else "001r"
         latest_feast = latest_chant.feast.id if latest_chant.feast else ""
-        latest_seq = latest_chant.sequence_number if latest_chant.sequence_number else 0
+        latest_seq = latest_chant.sequence_number if latest_chant.sequence_number is not None else 0 # "is not None": sequence_number is sometimes 0
```
dchiller: Slightly confused about this line....but probably because I am not familiar with the database.

`latest_chant` will either have non-zero positive `sequence_number` (this is sort of the "normal case") or it will have `sequence_number = 0` in the case where we have the fragment situation mentioned in #440. Will it ever be `None`? 

If not, then this is fine, but if it could be, then this seems incorrect?

jacobdgm: Yes, there are situations where a `sequence_number` has not been provided, so its value is `None`.

Functionally, both the original version of 1074 and the changed version of 1074 do the same thing, but this version (I hope!) makes it a tad clearer that in situations where `latest_chant.sequence_number` is falsy, we're dealing with two conceptually distinct cases:

- if `latest_chant.sequence_number` is `0`, it's an actual/valid sequence number, so `latest_seq` can literally become `latest_chant.sequence_number`
- if `latest_chant.sequence_number` is `None`, it's not an actual sequence number, so we set `latest_seq` to a placeholder value of `0`
- In both cases, the result is the same: `latest_seq` is now `0`

But the comment at the end of the line actually muddies the water rather than clarifying things I think. I should just remove it. Thanks for making note of this!

dchiller (approved):

PR #454, opened 2022-12-02
Mark required fields on Source Create page with asterisks

### LINKED ISSUE #449: required fields
On OldCantus, to create a source you need both a manuscript ID and a siglum (fields marked with asterisk) otherwise it won't create the source. 
NewCantus has no asterisks on these fields, and was quite happy to let me make sources with no siglum (though it does tell me to fill out an ID field if I try to submit without it.)

On the chant level, Folio and Sequence seem to be required fields (they are not on OldCantus!) but are not marked as such with asterisks, either.

jacobdgm:
fixes #449 

This PR sets Sources' `siglum` fields to required (as they are on OldCantus) and adds asterisks to both mandatory fields (`siglum` and `title`) on the source-create page indicating this.

Along the way, I needed to update some tests to accommodate these changes, and I used this opportunity to do some refactoring (i.e. mostly a lot of turning `Source.objects.create(...)` into `make_fake_source(...)`)

### DIFF
--- a/django/cantusdb_project/main_app/models/source.py
+++ b/django/cantusdb_project/main_app/models/source.py
@@ -33,8 +33,8 @@ class Source(BaseModel):
     # it is a human-readable ID for a source
     siglum = models.CharField(
         max_length=63, 
-        null=True, 
-        blank=True,
+        null=False, 
+        blank=False,
         help_text="RISM-style siglum + Shelf-mark (e.g. GB-Ob 202).",
     )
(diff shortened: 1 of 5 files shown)


jacobdgm:
UPDATE: asterisks added and tests updated - we're good to go!

Wait! don't approve this yet, as some tests are failing - I need to update the tests and add asterisks to the source-edit page as well.

dchiller (approved):

PR #468, opened 2022-12-09
replace `.sequence` and `.sequence_number` with `.s_sequence` and `.c_sequence` respectively

### LINKED ISSUE #455: sequence numbers not editable
When investigating the folios and sequence fields for #453 , I found that sequence numbers populate automatically, which we want: 
image

But if one tries to edit this to, say, sequence 102: 
image

there is a success message, but the sequence number stays firmly at 100:
image

### LINKED ISSUE #463: Chants should use `sequence_number`, but ChantEdit and ChantProofread views set chants' `sequence`
Our `Chant` and `Sequence` models both inherit from `BaseChant`, so they both have a `sequence_number` field and a `sequence` field. `sequence_number` is supposed to be used only for `Chant`s, and `sequence` is supposed to be used only for `Sequence`s. The ChantCreate page properly sets chants' `sequence_number`, but the ChantEdit and ChantProofread pages set chants' `sequence` instead.

So, when a user loads a chant's Edit page, the `sequence` field is not populated, as the chant has a `sequence_number` but no `sequence`. When a user changes a `sequence` in the ChantEdit page and then visits the chant's Detail page, their edit will appear to have had no effect, since we don't ever try to display the chant's `sequence`.

jacobdgm:
fixes #455, fixes #463

Previously, we had a confusing situation where `Sequence` and `Chant` both inherited from `BaseChant` (to combine querysets in certain places, we needed to ensure they had the same fields), but `Chant` had a `.sequence_number` property (which only ever had integers, so it was an IntegerField) and `Sequence` had a `.sequence` field (a StringField, since they sometimes included letters as well as numerals). This led to a great deal of confusion, and bugs such as #455. This PR goes through the codebase, replacing all instances of `.sequence_number` with `.c_sequence` (i.e. "chant sequence" and all instances of `.sequence` with `.s_sequence` (i.e. "sequence sequence"), variable names chosen to be concise and unambiguous.

Once this is merged, it will require `makemigrations` and `migrate`, and the confirmations that the fields for both models have indeed been substituted. When I ran this locally, Django suggested the correct substitutions, so I don't expect this not to go smoothly when we're applying this update to our servers.

Rather than simply performing a straight swap, I substituted the correct new property in places where the incorrect old property was used (such that this PR fixes #455). I also fixed the CSV export API and added a test to ensure that it is properly outputting rows for sequences, instead of only for chants (whereas OldCantus just outputs a CSV header with no rows for sources containing Sequences with no chants, NewCantus has output data on Sequences in a source, but with an empty `"sequence"` column up until now). I've added comments in the "Files changed" section wherever I made such corrections.


django/cantusdb_project/main_app/forms.py
```diff
@@ -231,7 +231,7 @@ class Meta:
             "volpiano",
             "marginalia",
             "folio",
-            "sequence",
+            "c_sequence",
```
jacobdgm, line 234: `"sequence"` was incorrect, this substitution corrects the error

django/cantusdb_project/main_app/forms.py
```diff
@@ -252,7 +252,7 @@ class Meta:
             "volpiano": VolpianoAreaWidget(),
             "marginalia": TextInputWidget(),
             "folio": TextInputWidget(),
-            "sequence": TextInputWidget(),
+            "c_sequence": TextInputWidget(),
```
jacobdgm, line 255: `"sequence"` was incorrect, this substitution corrects the error

django/cantusdb_project/main_app/forms.py
```diff
@@ -288,7 +288,7 @@ class Meta:
             "volpiano",
             "marginalia",
             "folio",
-            "sequence",
+            "c_sequence",
```
jacobdgm, line 291: `"sequence"` was incorrect, this substitution corrects the error

django/cantusdb_project/main_app/forms.py
```diff
@@ -318,7 +318,7 @@ class Meta:
             "volpiano": VolpianoAreaWidget(),
             "marginalia": TextInputWidget(),
             "folio": TextInputWidget(),
-            "sequence": TextInputWidget(),
+            "c_sequence": TextInputWidget(),
```
jacobdgm, line 321: `"sequence"` was incorrect, this substitution corrects the error

django/cantusdb_project/main_app/management/commands/sync_sequences.py
```diff
@@ -76,9 +76,9 @@ def get_new_sequence(seq_id):
         folio = None
 
     try:
-        sequence_number = json_response["field_sequence_text"]["und"][0]["value"]
+        s_sequence = json_response["field_sequence_text"]["und"][0]["value"]
```
jacobdgm, line 79: `"sequence_number"` was incorrect, this substitution corrects the error (also line 81)

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1318,9 +1318,9 @@ def get_context_data(self, **kwargs):
 
         # 4064 is the id for the sequence database
         if source.segment.id == 4064:
-            queryset = source.sequence_set.order_by("folio", "sequence_number")
+            queryset = source.sequence_set.order_by("folio", "s_sequence")
         else:
-            queryset = source.chant_set.order_by("folio", "sequence")
+            queryset = source.chant_set.order_by("folio", "c_sequence")
```
jacobdgm, line 1323: `"sequence_number"` and `"sequence"` were incorrect, these substitutions correct the error

dchiller (approved):

PR #481, opened 2022-12-16
Remove all code related to chant-delete view

jacobdgm:
A while ago, Jiwoo set up a chant-delete view to facilitate testing during development. Now that we are spending less time implementing features related to chant editing, this view is less useful, and is mostly a way for users to delete things from the database, either by accident or if they become disgruntled for some reason (and if we make a mistake somehow and this view becomes accessible to the general public, this could be a significant problem). We don't need it anymore, so let's get rid of it.


annamorphism:
won't some users (proofreaders, for example) occasionally want to delete things, though--chants that accidentally got put in in duplicate, for example? or is this something separate from that?

jacobdgm:
No, they shouldn't need to. This is not a feature of OldCantus as far as I know - it's something Jiwoo set up to facilitate development.

My understanding is that if someone finds a chant that needs to be deleted, they contact Debra (or someone else with admin privileges), who can then delete it using the admin page.

jacobdgm:
I suppose we could check before deleting it, but we should put some thought into how we want it to be implemented it if we decide to keep it around (since it amounts to a new feature that no one, to my knowledge, has asked for). It had a link from the chant edit page (but not from the proofread page), and could be accessed by replacing `chant` with `chant-delete` in the URL of any chant detail page.

My nightmare is that we change something with our permissions settings in the future, and some random person on the internet discovers they aren't actually shown a 403 error when they add `-delete` to a URL, and go about deleting chants at random before someone realizes what's happening. If we don't realize for a while, this could mean a lot of lost work if we have to roll the database back by several months.

dchiller (approved):
This looks good to me from a code perspective, so I'm approving! 

It doesn't sound like this was developed with the idea that it would be on the production site, so it seems like a good call to revisit if that's a desired feature, rather than to just include this. Maybe this code serves as the basis, but it sounds like we'd want to have additional safeguards.

Issue #508, opened 2023-01-11
On Chant Edit pages, update the melody-with-text preview in real time


jacobdgm:
In an email from Lucia Denk to Junhao,

> Jennifer and I thought of a possible new feature/update for Cantus Database which we'd like to ask you about. Currently, whenever we edit the Volpiano transcription of a chant in Cantus, the display in Volpiano does not include the text underlay, just the melody. As a result, making changes to the melody/fixing note errors can be difficult, since in the current editing display it's harder to locate exactly where in the transcription changes have to be made. We were wondering, therefore, if there might be a way to change the layout for the Volpiano editor to include the text, so as to make it easier to edit melodic transcriptions? Curious to know your thoughts on this...thank you!

Junhao responded,

> If I understand correctly, you would like to see the text under melody in the volpiano editor, so that the change in text-melody alignment can be inspected as you edit the melody. Is that correct?
> 
> Implementation-wise, this sounds similar to what we already have for displaying text-melody alignment on the chant detail page, so it might be helpful to reuse some code from there.

and forwarded it to me.

Lucia responded,

> Yes, that's correct! Being able to view the text-melody alignment in the Volpiano editor while correcting the melodies would really help expedite the editing process.

I responded,

> So this would be on both the edit-chant page (where you can edit just a single chant) and on the edit-volpiano page (where there are links to quickly switch between editing any of the chants on a particular folio), correct?
> 
> To have the text appear beneath the volpiano _inside_ the volpiano input field would be very difficult. But on those pages currently, there is already a melody-with-text preview at the bottom of the page, beneath the "Indexing Notes" field. It would be easy to move the melody-with-text preview from where it currently is to directly beneath the volpiano field. From there, I could set it up to automatically regenerate the preview whenever a change is made to either the volpiano field or the ms-spelling field - would this be enough to streamline the process?

UPDATE: Jennifer responded,

> Yes – that would help things enormously. The other thing that would help, which may be too difficult, would be for the layout to be the same. Currently you might have, say, 50 pitches on the first staff in the Volpiano field and 40 pitches in the Volpiano + text display, so it makes it really tricky to ensure you’re actually in the right place in the editor.

annamorphism:
hear hear, I would love this feature!
Automatic regeneration would certainly be a big help. But part of the problem is that in a very long chant it's often hard to find the spot you need to be somewhere in the middle--if there were some way of having both the with and without-text version have the same line breaks, for example, that would also be a big help.

PR #525, opened 2023-01-18
Add melody-with-text preview to Chant Edit, Proofread pages

### LINKED ISSUE #507: OldCantus displays preview of melody with text on Chant Editing pages
visible on https://cantus.uwaterloo.ca/edit-volpiano/699578 (our Source Edit Chants page) https://cantus.uwaterloo.ca/node/702097/edit (our Edit Chant page), beneath Indexing Notes.

For some reason we don't have this implemented yet.

jacobdgm:
Fixes #507

Whereas on OldCantus, the preview is at the bottom of the page near the "Save" button, this PR adds the preview beneath the Volpiano field, where it will be most useful.

The preview is only displayed once the chant is saved, i.e. this PR makes no effort to update the preview in real time (see #508)

This is how it looks on the Chant Edit page:
Screen Shot 2023-01-18 at 11 32 19 AM
And here it is on the Chant Proofread page:
Screen Shot 2023-01-18 at 11 32 03 AM

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_edit.html
+++ b/django/cantusdb_project/main_app/templates/chant_edit.html
@@ -79,6 +79,28 @@
                         </div>
                     </div>
 
+                    {% if chant.volpiano %}
+                    <div class="form-row">
+                        <div class="form-group m-1 col-lg-12">
+                            <small>Preview of melody with text:</small>
+                            {% if chant.manuscript_syllabized_full_text %}
+                                <p><small>Syllabification is based on saved syllabized text.</small></p>
+                            {% endif %}
+                            <dd>
+                                {% for zip in syllabized_text_with_melody %}
+                                    {% for syl_mel, syl_text in zip %}
+                                        <span style="float: left">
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/templates/chant_edit.html
```diff
@@ -79,6 +79,28 @@
                         </div>
                     </div>
 
+                    {% if chant.volpiano %}
```
dchiller, line 82: As I understand it, you are moving this from below, right? So shouldn't there be an equivalent deletion starting at line 191?

jacobdgm: Nope, no duplication! I can see where the confusion arose - I copied this code (and the new code in chant_proofread.html) from the Chant Detail view. Up until now, NewCantus entirely lacked a text-and-melody preview on the Chant Edit and Chant Proofread pages - it was only present in the Chant Detail view. OldCantus, in contrast, had a preview near the bottom of its Chant Edit page. So now NewCantus's Chant Edit page has a text-and-melody preview, as it should, but it's in a more useful place on the page than it was in OldCantus's.

PR #533, opened 2023-01-30
Fix failing ChantSearchView tests

jacobdgm:
Evidently, I forgot to run the test suite before opening #529... Since the Chant Search View only returns a handful of columns necessary for rendering the search page now (rather than entire Chant objects), a number of tests which used `AssertIs` began failing. With this PR, all tests now pass.

I also took the opportunity to improve a few of the tests - instead of simply checking that chants that should be found by a particular query are found, several of the tests now include dummy chants that _shouldn't_ be found by a query, and checks that these chants are _not_ in fact returned in the context for the view.


jacobdgm:
Turns out these tests aren't sufficient - #529 actually broke the search page, as the data on the related objects is not actually being rendered by the template. I'll open a new PR soon reverting the changes and adding better tests.

PR #537, opened 2023-02-02
add `time_function` function for optimization testing

jacobdgm:
A useful function for testing for bottlenecks and verifying that optimizations are actually optimizing - putting it here so it'll always be around. If you notice any `@time_function`s left over from testing in future PRs, please tell me off as you would for extraneous `print` statements.

### DIFF
--- a/django/cantusdb_project/time_function.py
+++ b/django/cantusdb_project/time_function.py
@@ -1,0 +1,12 @@
+from functools import wraps
+from time import time
+
+def time_function(fn):
+    @wraps(fn)
+    def measure_time(*args, **kwargs):
+        t1 = time()
+        result = fn(*args, **kwargs)
+        t2 = time()
+        print(f"@time_function: {fn.__name__} took {t2 - t1:.3f}s.")
+        return result
+    return measure_time


dchiller (approved):
Haha! Will do.

fujinaga:
@jacobdgm Are you using Python's logger module? You can change the logging levels and you don't need to remove any code.

ahankinson:
To follow up on @fujinaga's suggestion about the logging modules, a few tips:

- You can set some custom logging levels pretty easily. If you're running in development, for example, you can set your default log level to DEBUG, whereas in production you can set it to ERROR. Here's a snippet of how we have logging configured in one of our projects: https://github.com/rism-digital/muscatplus_server/blob/develop/search_server/server.py#L64-L72
- When you do use logging, make sure you use the old-style percent string encoding, rather than the newer `f` strings... This is because with the older style, if the log level is not set to emit a message (e.g., you are running at ERROR and it's being asked to emit a DEBUG message), then the string won't actually get evaluated with percent-encoding.

So do this:

`log.debug("The value of the variable foo is %s", foo)`

not this:

`log.debug(f"The value of the variable foo is {foo}")`

and definitely not this:

`log.debug("The value of the variable foo is %s" % foo)`

See here: https://medium.com/flowe-ita/logging-should-be-lazy-bc6ac9816906

- Here's a slightly better decorator for timing functions: https://github.com/rism-digital/muscatplus_indexer/blob/develop/indexer/helpers/utilities.py#L17-L39

The small advantages of this one are: 
1. It uses the `timeit` module instead of `time`, which is [more suitable for timing execution time](https://stackoverflow.com/a/17579466) rather than wall time
2. It uses the Python logging system
3. It also gives the function name, and the time in hh:mm:ss, in a nicely formatted message

PR #539, opened 2023-02-06
Ensure Chant Search View has complete suite of tests

jacobdgm:
This PR improves our test suite for the Chant Search View, ensuring that the right information shows up where it should in each row of the results table.

These changes also necessitated some changes to our "make_fake_{}" functions. Previously, `\n`s would occasionally appear in fake texts. In places where this was causing problems, I switched from using `make_fake_text` to `faker.sentence`, which eliminates the possibility of new lines being generated.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -807,7 +807,6 @@ class ChantSearchViewTest(TestCase):
         office = make_fake_office()
         office_name = office.name
         office_description = office.description
-        print("office_description:", office_description)
         url = office.get_absolute_url()
         fulltext = "manuscript full text"
         search_term = "full"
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -749,6 +749,237 @@ def test_keyword_search_contains(self):
...
+        search_term = "full"
+        chant = make_fake_chant(
+            source=source,
+            manuscript_full_text_std_spelling=fulltext
+        )
+        response = self.client.get(
+            reverse("chant-search"), {"keyword": search_term, "op": "contains"}
+        )
+        html = str(response.content)
+        self.assertIn(siglum, html)
```
dchiller, line 766: Would these first two `.assertIn` statements be covered by the 3rd? 

I see this similar repetition in other tests too, so I'm assuming it's intentional....so just curious what this does...

jacobdgm: Yes, they would be.

In writing these tests, I've been following a pattern of testing for one or a few specific details, and then testing that they all come together correctly. Though I hadn't fully thought this out when I was writing them, I think the reason I did it is to communicate intent to future developers. The first two `assertIn`s are my way of saying, "yes, we really do expect that both the siglum and the URL are present and correct", while the third one is just there to make sure everything is fitting together properly. I'm thinking of these tests, I suppose, as a kind of documentation - as indications/hints about what the important elements are in each of our views and functions.

Another way of putting it: if, in the future, someone tweaks the relevant `<a>` tag in the template, I expect them to adjust the final `assert` to reflect the change, but I also expect them to make sure they know what they're doing before they delete either of the first two `assert`s.

Do you think this approach makes sense?

dchiller: Thanks for explaining! This is a thoughtful approach...and you've consistently applied it, so I think it works. I hadn't thought about tests as (among other things) an additional form of documentation ~ I like that concept.

dchiller (approved):

django/cantusdb_project/main_app/tests/test_views.py
dchiller: A maybe-related question I have is the reasoning behind creating minimal fake objects for each test. For example, to test the feast and office columns, you create a fake chant with a fake feast and a fake chant with a fake office, respectively, but minimal other attributes. It seems like another approach would be to create a fake chant with both a fake feast and fake office and test both columns against that one chant, testing all those features with multiple assert statements. Are you trying to stick with the minimum object that will pass the test in each case so that that's clear to other's looking at the code (ie. a  similar "documentary" purpose" as above)?

I'm mostly asking because in reviewing this I realize I have no idea what's going on with Cantus Ultimus's tests//haven't touched them since I joined the lab (I think maybe Eli did something with them early in summer 2022?) so I'm now curious about the way you're thinking about them and think it might be good for me to get some sense of what's going on over at CU.

jacobdgm: This is a good point, and one I haven't thought about very deeply! One thing that comes to mind - not sure if this is exactly what you're thinking of - is that if we test, say, filtering of search results by multiple fields at the same time, the number of tests we write becomes an exercise in combinatorics.

But in any case, it would probably be worth having a couple of tests at the bottom of each test class that build one or several objects with as many attributes as we expect to see "in the real world" for which we test that things come together properly, without aiming to be exhaustive.

jacobdgm: > I realize I have no idea what's going on with Cantus Ultimus's tests//haven't touched them since I joined the lab

Spending some time to make sure the test suite is up-to-date and covers everything important is something I can hardly recommend highly enough! For me, not only do I find that good tests alleviate most of my concerns that I've introduced a subtle bug someplace whenever I make a change, but I also find that writing tests can clarify what a new function/feature will do.

If you need any help with this, or have any questions, let me know: I enjoy writing tests, and would be happy to help wherever I can!

PR #541, opened 2023-02-08
add `.absolute_url` field to all models

jacobdgm:
This PR adds a new field, `.absolute_url`, to all (non-User) models. In doing so, it clears the way for implementing the change in #529 - this way, all the data needed to render the template will be present in the database, and we will no longer need to load Python objects into memory in order to call their `.get_absolute_url` methods.

The PR adds a new hook to `signals.py`, which ensures that whenever an object is saved, its `absolute_url` field is up-to-date (we don't expect objects' `id`s to change often (perhaps while syncing data with OldCantus?), but this ensures that nothing breaks if it occurs). In order to populate these fields initially, a new management command, `save_all_objects`, has been added to load and save all objects in the database, thus triggering the new `on_object_save` hook. _This command will need to be run when these changes are added to our staging and production servers._

If I got all my `git checkout`s correct, these changes pass all tests introduced in #539.

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/save_all_objects.py
+++ b/django/cantusdb_project/main_app/management/commands/save_all_objects.py
@@ -13,6 +13,10 @@ from main_app.models import (
     Source,
 )
 
+# Run with `python manage.py save_all_objects`.
+# This command runs through all objects in the database
+# and saves them, thus triggering any on_save hooks in signals.py.
+
 
 class Command(BaseCommand):
     def add_arguments(self, parser):
--- a/django/cantusdb_project/main_app/management/commands/save_all_objects.py
+++ b/django/cantusdb_project/main_app/management/commands/save_all_objects.py
@@ -1,0 +1,43 @@
+from django.core.management.base import BaseCommand
+from main_app.models import (
+    Century,
+    Chant,
+    Feast,
+    Genre,
+    Notation,
+    Office,
+    Provenance,
+    RismSiglum,
+    Segment,
+    Sequence,
+    Source,
+)
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        pass
+
+    def handle(self, *args, **options):
+        for cls in (
+            Century,
+            Chant,
+            Feast,
+            Genre,
+            Notation,
+            Office,
+            Provenance,
+            RismSiglum,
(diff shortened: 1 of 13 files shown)


jacobdgm:
hold off on approving this for a moment - all the tests are passing, but when I view it in a browser, there's still a problem with some of the links on the chant search page...

Update - the changes to the models and template seem to be fine as far as I can tell, but the script to save all the objects appears not to have hit all the objects, for reasons I don't fully understand. I made some changes, and we're running it again.

jacobdgm:
@dchiller, can you take a look at the `save_all_objects` script? I tried to run it yesterday, and it seems not to have touched certain chants. Out of the ~500,000 chants, there were at least 73851 for which their `absolute_url` was `None`. I tried running it again today (my computer fell asleep several times yesterday as the script ran into the evening, so I thought that might have been the culprit), and after a second pass, the number of chants without `absolute_url`s is now 33425. It's skipping chants at random, and I don't understand why.

Similarly, out of our 17,000 sequence objects, 6192 had been missed in the first pass, and 2185 remained after the second pass.

If you can't see what's going wrong, I may try to rewrite the script to simply update the specific field rather than loading and saving the entire chant object. But I would like to figure out what's going wrong!

ahankinson:
Could you explain a bit more about the motivation for this?

Django can construct a url for any route given only a PK. This means that you just need to get a single field from the db to construct a URL. Moreover, if you have a set of FKs you can also constuct URLs for them too.

So I guess I’m missing the motivation for why you’re storing the full URL string in the database as well. You say something about not needing to load the database objects in memory, but you’ll need to do that if you want to retrieve the URL from there.

jacobdgm:
> Django can construct a url for any route given only a PK.

Aha! Thank you for this! I can replace `{{ chant.get_absolute_url }}` with `{% url 'chant-detail' chant.id %}` in all templates to link to a chant detail page, and adapt this to link to feast, source, etc. detail pages as necessary. A much better solution - again, thank you!

I should get this set up everywhere in the codebase, and then get rid of our models' `.get_absolute_url` method, so future developers aren't misled as I was - I spent a while trying to figure out whether I could annotate querysets using `django.urls.reverse`, but now it's clear that there's a much simpler way.

PR #551, opened 2023-02-10
Full Index view etc.: Replace instances of "index" with "inventory"

### LINKED ISSUE #550: Tweaks to Index page
In an email from Debra:

> One hopefully quick thing: Can we change "CANTUS Manuscript Index" at the top of the screen to "Cantus Manuscript Inventory"?
> 
> The listing of records just below could also say "inventory" instead of "index" (if that's a single change in the code that's easy to do).

re: http://206.12.93.196/index/?source=669163

jacobdgm:
fixes #550 

I also took this opportunity, since I was looking at it, to clean up the sidebars on several of our views. It's always looked cluttered to me; now all the links are on their own lines, the "»"s are applied with a CSS pseudoclass, and I got to remove a bunch of `<br>`s and `&nbsp`s from the templates.

Before (Source Detail page):
Screen Shot 2023-02-10 at 4 27 47 PM

Now:
Screen Shot 2023-02-10 at 4 28 28 PM

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -268,7 +268,7 @@
 
                 <h4>List of melodies</h4>
                 <span id="melodyLoadingPrompt" style="display: none; color: #922"><b>Loading melodies...</b></span>
-                <a id="melodyButton" href="#" onclick="loadMelodies('{{ chant.cantus_id }}'); return false;">» Display the melodies connected with this chant</a>
+                <a id="melodyButton" href="#" onclick="loadMelodies('{{ chant.cantus_id }}'); return false;">Display the melodies connected with this chant</a>
                 <div id="melodyDiv"></div>        
             {% endif %}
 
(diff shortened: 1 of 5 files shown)


django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -268,7 +268,7 @@ <h4 id="concordances">List of concordances</h4>
 
                 <h4>List of melodies</h4>
                 <span id="melodyLoadingPrompt" style="display: none; color: #922"><b>Loading melodies...</b></span>
-                <a id="melodyButton" href="#" onclick="loadMelodies('{{ chant.cantus_id }}'); return false;">» Display the melodies connected with this chant</a>
+                <a id="melodyButton" href="#" onclick="loadMelodies('{{ chant.cantus_id }}'); return false;">Display the melodies connected with this chant</a>
```
dchiller, line 271: I notice that the new class ("guillemet") is not applied here. This also isn't included in the screenshots, so of course your intention could be to not have the double "greater than sign" (lol, what is the English name for these?) here. So I'm approving since you will know whether this was oversight or intentional! :)

dchiller (approved):

django/cantusdb_project/main_app/templates/chant_detail.html
jacobdgm: I think its's "guillemet" even in English - at least, wikipedia provides a few anglicized pronunciations of it.

I consciously did not apply `class="guillemet"` here because this is in the main panel of the page and not in the sidebar. These show up in a couple of other places and I'm not sure whether we'll want them all to be `display: block`.

But in general, I'm not sure why OldCantus uses guillemets for some links and not others. It doesn't seem to indicate that they're external web pages. So if we were to adopt a consistent approach, I'm not sure what the correct consistent approach would be...

PR #566, opened 2023-02-22
In templates, replace instances of .get_absolute_url with url tags

### LINKED ISSUE #478: On Chant Detail page, in the Source Navigation sidebar, indicate current chant
Currently, on the chant detail page, there is a sidebar that lists the chants on the current folio as well as on nearby folios. In the sidebar, there is no indication of which chant is currently being displayed:
Screen Shot 2022-12-15 at 1 19 53 PM

It might be helpful if we visually indicated somehow the chant that is currently being displayed.
Possible options:
1. Bold it
2. make it not be a link
3. add a little arrow pointing to the chant
4. a combination of 2. and 3., or maybe all three of the options above.

For example, here's a quick mockup of option 4:

Screen Shot 2022-12-15 at 1 34 09 PM

### LINKED ISSUE #543: Wherever we display `feast.name`, add a `feast.description` title attribute
for mouseover. Simple to implement. Not at all urgent.

jacobdgm:
Mainly as a step toward implementing more effective querying in the Chant Search View and other views, this PR replaces almost all instances of `.get_absolute_url` in all templates with `{% url %}` tags.

I also had to go through and add some `{% if chant.feast %}` (or .office, .source, .genre, etc.) when rendering links, to prevent the template from generating an error when it tried to get the `id` of a nonexistent feast/office/etc. Before changing how links are encoded, these would have generated empty links (equivalent to `<a href="">`).

As I was working in the views, it was not difficult to implement some other small changes we've been thinking of for a while. These include adding title attributes giving more information about feasts, genres and offices when they are moused over (i.e. fixes #543), bolding the current chant on the Chant detail page (i.e. fixes #478), and clarifying templates by improving indentation in some places.

(also - oops, it looks as though the change to `requirements.txt` from #560 got rolled up into this PR. I can remove this if you think it's best)

### DIFF
--- a/django/cantusdb_project/requirements.txt
+++ b/django/cantusdb_project/requirements.txt
@@ -13,6 +13,7 @@ Django==4.1.7
 django-autocomplete-light==3.5.1
 django-extra-views==0.13.0
 django-quill-editor==0.1.40
+django_debug_toolbar==3.8.1
 Faker==4.1.0
 gunicorn==20.0.4
 idna==2.10
--- a/django/cantusdb_project/align_text_mel.py
+++ b/django/cantusdb_project/align_text_mel.py
@@ -155,7 +155,6 @@ def syllabize_melody(volpiano):
         # remove the trailing "--" (added in previous line) from the last syllable
         syls[-1] = syls[-1][:-2]
         syls_melody.append(syls)
-    # print(syls_melody)
     return syls_melody
 
 
--- a/django/cantusdb_project/articles/templates/article_detail.html
+++ b/django/cantusdb_project/articles/templates/article_detail.html
@@ -1,11 +1,12 @@
 {% extends "base.html" %}
 {% block content %}
+<title>{{ article.title }} | Cantus Manuscript Database</title>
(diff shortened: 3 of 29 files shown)


django/cantusdb_project/requirements.txt
```diff
@@ -13,6 +13,7 @@ Django==4.1.7
 django-autocomplete-light==3.5.1
 django-extra-views==0.13.0
 django-quill-editor==0.1.40
+django_debug_toolbar==3.8.1
```
dchiller, line 16: I think this is fine...

dchiller (approved):
Nice use of the django template tags -- I agree that this is a good change. 

It looks like you may have also changed some html formatting (which looks good imho) but just wanted to make sure you were aware.

django/cantusdb_project/requirements.txt
jacobdgm: > changed some html formatting

i.e. mostly splitting lines down into nested tags on different lines?

dchiller: exactly.

PR #568, opened 2023-02-28
Improve Chant Search View

jacobdgm:
This PR implements several changes to the Chant Search View in order to improve execution time:
- when a queryset containing Chants and a queryset containing Sequences are combined using `.union()`, we now pass an `all=True` argument to prevent Postgres from checking the resulting queryset to ensure all items are `DISTINCT` from each other
- we now apply `.values()` to fetch only the values needed for rendering the template
- these changes were applied to both ChantSearchView and ChantSearchMSView, since they both use the same template

Some other related changes occurred:
- updated tests for Chant Search View
- updated tests for Chant Search MS View
  - Also wrote a bunch of new tests in ChantSearchMSViewTest, mostly copied from ChantSearchViewTest, ensuring that the values/links/etc. in the cells of the table render correctly. Since we previously weren't testing this, the Chant Search MS page was initially broken by the changes in the template, and the all tests were passing even when they shouldn't have been. So it's a good thing these tests are now here!
- we no longer use `keyword_search`, previously used in both views mentioned above, so I was able to delete the function.
- some slight refactoring of how we handle query parameters, mostly to make the code more self-documenting

I tried to replace the `{% if chant.search_vector %}` field with an annotation, but I kept running into an error, `ORDER BY term does not match any column in the result set.` This will be a nice thing to refactor eventually, but it's not urgent, so I'll try to figure this out another time. Since we no longer use `keyword_search`, once this is done, we can remove the `search_vector` from our Chant/Sequence model.

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_search.html
+++ b/django/cantusdb_project/main_app/templates/chant_search.html
@@ -209,8 +209,8 @@
                     {% for chant in chants %}
                         <tr>
                             <td class="text-wrap" style="text-align:left">
-                                {% if chant.source %}
-                                    <a href="{% url 'source-detail' chant.source.id %}" title="{{ chant.source.title }}">{{ chant.source.siglum }}</a>
+                                {% if chant.source__id %}
+                                    <a href="{% url 'source-detail' chant.source__id %}" title="{{ chant.source__title }}">{{ chant.source__siglum }}</a>
                                 {% endif %}
                             </td>
                             <td class="text-wrap" style="text-align:center">
@@ -233,25 +233,25 @@
                                 </p>           
                             </td>
                             <td class="text-wrap" style="text-align:center">
-                                {% if chant.feast %}
-                                    <a href="{% url 'feast-detail' chant.feast.id %}" title="{{ chant.feast.description }}">{{ chant.feast.name|default:"" }}</a>
+                                {% if chant.feast__id %}
+                                    <a href="{% url 'feast-detail' chant.feast__id %}" title="{{ chant.feast__description }}">{{ chant.feast__name|default:"" }}</a>
                                 {% endif %}
                             </td>
                             <td class="text-wrap" style="text-align:center">
-                                {% if chant.office %}
-                                    <a href="{% url 'office-detail' chant.office.id %}" title="{{ chant.office.description }}">{{ chant.office.name|default:"" }}</a>
+                                {% if chant.office__id %}
+                                    <a href="{% url 'office-detail' chant.office__id %}" title="{{ chant.office__description }}">{{ chant.office__name|default:"" }}</a>
                                 {% endif %}
                             </td>
                             <td class="text-wrap" style="text-align:center">
-                                {% if chant.genre %}
-                                    <a href="{% url 'genre-detail' chant.genre.id %}" title="{{ chant.genre.description }}">{{ chant.genre.name|default:"" }}</a>
+                                {% if chant.genre__id %}
+                                    <a href="{% url 'genre-detail' chant.genre__id %}" title="{{ chant.genre__description }}">{{ chant.genre__name|default:"" }}</a>
                                 {% endif %}
                             </td>
                             <td class="text-wrap" style="text-align:center">
                                 {{ chant.position|default:"" }}
                             </td>
                             <td class="text-wrap" style="text-align:center">
-                                <a href="{{ chant.get_ci_url }}" target="_blank">{{chant.cantus_id|default:"" }}</a>
+                                <a href="http://cantusindex.org/id/{{ chant.cantus_id }}" target="_blank">{{chant.cantus_id|default:"" }}</a>
                             </td>
                             <td class="text-wrap" style="text-align:center">
                                 {{ chant.mode|default:"" }}
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -818,8 +818,12 @@ class ChantSearchView(ListView):
                 keyword = self.request.GET.get("keyword")
                 # the operation parameter can be "contains" or "starts_with"
                 if self.request.GET.get("op") == "contains":
-                    chant_set = keyword_search(chant_set, keyword)
-                    sequence_set = keyword_search(sequence_set, keyword)
+                    chant_set = chant_set.filter(
+                        manuscript_full_text_std_spelling__icontains=keyword
+                    )
+                    sequence_set = sequence_set.filter(
+                        manuscript_full_text_std_spelling__icontains=keyword
+                    )
                 else:
                     chant_set = chant_set.filter(incipit__istartswith=keyword)
                     sequence_set = sequence_set.filter(incipit__istartswith=keyword)
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -722,8 +722,6 @@ class ChantSearchView(ListView):
         return context
 
     def get_queryset(self) -> QuerySet:
-        CHANT_SEARCH_TEMPLATE_VALUES # in order to fetch only those values necessary
-                                     # for rendering chant_search.html template
         # Create a Q object to filter the QuerySet of Chants
         q_obj_filter = Q()
         display_unpublished = self.request.user.is_authenticated
@@ -975,8 +973,6 @@ class ChantSearchMSView(ListView):
         return context
 
     def get_queryset(self) -> QuerySet:
-        CHANT_SEARCH_TEMPLATE_VALUES # in order to fetch only those values necessary
-                                     # for rendering chant_search.html template
         # Create a Q object to filter the QuerySet of Chants
         q_obj_filter = Q()
         # If the "apply" button hasn't been clicked, return empty queryset
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -829,7 +829,7 @@ class ChantSearchView(ListView):
                     sequence_set = sequence_set.filter(incipit__istartswith=keyword)
 
             # once unioned, the queryset cannot be filtered/annotated anymore, so we put union to the last
-            queryset = chant_set.union(sequence_set)
+            queryset = chant_set.union(sequence_set, all=True)
             queryset = queryset.order_by(order, "id")
 
         return queryset
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -992,6 +992,34 @@ class ChantSearchMSView(ListView):
         return context
 
     def get_queryset(self) -> QuerySet:
+        template_values = (
+            # fetch only those database columns necessary for rendering
+            # the template
+            "id",
+            "folio",
+            "search_vector",
+            "incipit",
+            "manuscript_full_text_std_spelling",
+            "position",
+            "cantus_id",
+            "mode",
+            "manuscript_full_text",
+            "volpiano",
+            "image_link",
+
+            "source__id",
+            "source__title",
+            "source__siglum",
+            "feast__id",
+            "feast__description",
+            "feast__name",
+            "office__id",
+            "office__description",
+            "office__name",
+            "genre__id",
+            "genre__description",
+            "genre__name",
+        )
         # Create a Q object to filter the QuerySet of Chants
         q_obj_filter = Q()
         # If the "apply" button hasn't been clicked, return empty queryset
@@ -1058,14 +1086,31 @@ class ChantSearchMSView(ListView):
         )
         # Filter the QuerySet with Q object
         queryset = queryset.filter(q_obj_filter)
+        # Fetch only the values necessary for rendering the template
+        queryset = queryset.values(*template_values)
         # Finally, do keyword searching over the QuerySet
         if self.request.GET.get("keyword"):
             keyword = self.request.GET.get("keyword")
+            operation = self.request.GET.get("op")
             # the operation parameter can be "contains" or "starts_with"
-            if self.request.GET.get("op") == "contains":
-                queryset = keyword_search(queryset, keyword)
-            else:
-                queryset = queryset.filter(incipit__istartswith=keyword)
+            if operation == "contains":
+                ms_spelling_filter = Q(
+                    manuscript_full_text__icontains=keyword
+                )
+                std_spelling_filter = Q(
+                    manuscript_full_text_std_spelling__icontains=keyword
+                )
+                keyword_filter = ms_spelling_filter | std_spelling_filter
+                queryset.filter(keyword_filter)
+            elif operation == "starts_with":
+                ms_spelling_filter = Q(
+                    manuscript_full_text__istartswith=keyword
+                )
+                std_spelling_filter = Q(
+                    manuscript_full_text_std_spelling__istartswith=keyword
+                )
+                keyword_filter = ms_spelling_filter | std_spelling_filter
+                queryset.filter(keyword_filter)
         # ordering with the folio string gives wrong order
         # old cantus is also not strictly ordered by folio (there are outliers)
         # so we order by id for now, which is the order that the chants are entered into the DB
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -797,33 +797,34 @@ class ChantSearchView(ListView):
                 # as a substring
                 feasts = Feast.objects.filter(name__icontains=feast)
                 q_obj_filter &= Q(feast__in=feasts)
-            order: str = 'siglum'
             order_get_param: Optional[str] = self.request.GET.get('order')
             sort_get_param: Optional[str] = self.request.GET.get('sort')
-            if order_get_param is not None:
-                if order_get_param == 'siglum':
-                    pass # this is the default
-                elif order_get_param == 'incipit':
-                    order = 'incipit'
-                elif order_get_param == 'office':
-                    order = 'office'
-                elif order_get_param == 'genre':
-                    order = 'genre'
-                elif order_get_param == 'cantus_id':
-                    order = 'cantus_id'
-                elif order_get_param == 'mode':
-                    order = 'mode'
-                elif order_get_param == 'has_fulltext':
-                    order = 'manuscript_full_text'
-                elif order_get_param == 'has_melody':
-                    order = 'volpiano'
-                elif order_get_param == 'has_image':
-                    order = 'image_link'
+            
+            order_param_options = (
+                'incipit',
+                'office',
+                'genre',
+                'cantus_id',
+                'mode',
+                'has_fulltext',
+                'has_melody',
+                'has_image',
+            )
+            if order_get_param in order_param_options:
+                if order_get_param == "has_fulltext":
+                    order = "manuscript_full_text"
+                elif order_get_param == "has_melody":
+                    order = "volpiano"
+                elif order_get_param == "has_image":
+                    order = "image_link"
+                else:
+                    order = order_get_param
+            else:
+                order = 'siglum'
 
-            if sort_get_param and sort_get_param == "asc":
-                pass # this is the default
-            elif sort_get_param and sort_get_param == 'desc':
-                order = "-" + order
+            # sort values: "asc" and "desc". Default is "asc"
+            if sort_get_param and sort_get_param == 'desc':
+                order = f"-{order}"
 
             if not display_unpublished:
                 chant_set: QuerySet = Chant.objects.filter(source__published=True)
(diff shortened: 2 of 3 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -716,6 +692,32 @@ def get_context_data(self, **kwargs) -> dict:
         return context
 
     def get_queryset(self) -> QuerySet:
+        values_for_rendering_template = (
```
ahankinson: Typically for static lists like this I define them somewhere external to the method itself, since there is no need to re-allocate the list every time the method is called. In itself, it would have a negligible effect on performance or memory usage, but I also find that just thinking in that way helps me to make lots of coding decisions that, in combination, makes a difference.

I think this list is also repeated below as well, so that also makes it a target for refactoring.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -813,19 +829,36 @@ def get_queryset(self) -> QuerySet:
...
+            sequence_set = sequence_set.values(*values_for_rendering_template)
             # Finally, do keyword searching over the querySet
             if self.request.GET.get("keyword"):
                 keyword = self.request.GET.get("keyword")
-                # the operation parameter can be "contains" or "starts_with"
-                if self.request.GET.get("op") == "contains":
-                    chant_set = keyword_search(chant_set, keyword)
-                    sequence_set = keyword_search(sequence_set, keyword)
+                operation = self.request.GET.get("op")
+                if operation == "contains":
```
ahankinson: This is a case where type annotations would help understand a bit better... 

```
operation: Optional[str] = self.request.GET.get("op")
if operation and operation == "contains":
    ....
```

IOW, you should check the `None` case as well, since if `op` isn't in your request parameters, `GET.get("op")` will return `None`.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -813,19 +829,36 @@ def get_queryset(self) -> QuerySet:
...
+                    std_spelling_filter = Q(
+                        manuscript_full_text_std_spelling__icontains=keyword
+                    )
+                    keyword_filter = ms_spelling_filter | std_spelling_filter
+                    chant_set = chant_set.filter(keyword_filter)
+                    sequence_set = sequence_set.filter(keyword_filter)
                 else:
-                    chant_set = chant_set.filter(incipit__istartswith=keyword)
-                    sequence_set = sequence_set.filter(incipit__istartswith=keyword)
+                    ms_spelling_filter = Q(
```
ahankinson, line 856: There's a lot of duplicate code between the two cases of the if statement.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I like this suggestion - let me make this change

django/cantusdb_project/main_app/views/chant.py
jacobdgm: good point - thank you!

django/cantusdb_project/main_app/views/chant.py
jacobdgm: > This list is also repeated below as well

Yes, it's repeated exactly, because both views use the same template. I thought of putting the list at the top of the file to avoid this duplication, but I expect to be making similar-but-different lists for other views (e.g. the FullIndexView needs a similar update). Better, I think, to put them in a consistent place near to where they're needed.

> somewhere external to the method itself, since there is no need to re-allocate the list every time the method is called

Django docs [suggest](https://docs.djangoproject.com/en/4.1/topics/class-based-views/intro/#using-class-based-views) that class-based views are "instantiated for each request dispatched to [them]", so defining this list in the `get_context_data` method will take just as much time/memory as if it's defined as a class variable. And since each view will have its own unique list of values, I think it makes most sense to define the list somewhere in the class.

Does this reasoning make sense? Is there a way to approach this I'm not thinking of?

ahankinson: For lists that are almost similar, I typically define a core set of fields common to both, and then append the specific differences.

```
FOO = ["1", "2", "3"]
SPECIAL_FOO = [*FOO, "5"]
SPECIAL_BAR = [*FOO, "4"]
```

Often, I'll define these sorts of things in the top level of the file directly. That way they only get allocated once when the application is loaded. 

Again, the difference in computing resources for this specific instance is miniscule, but it helps to develop patterns like this as a way of working.

jacobdgm: Sorry, could have been clearer - "similar" in concept, but potentially quite different in terms of the values in the list.

Having sat with the idea for a while, I think a reasonable implementation is to include a list (or tuple, as the case may be) at the top of the file for the specific template .html file, and we can add additional similar-in-concept lists when we do similar things to different views. I'll make this change - thanks for pointing this out!

dchiller:
What is a chant's `search_vector`?

dchiller (approved):

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -692,32 +722,8 @@ def get_context_data(self, **kwargs) -> dict:
...
-            "feast__description",
-            "feast__name",
-            "office__id",
-            "office__description",
-            "office__name",
-            "genre__id",
-            "genre__description",
-            "genre__name",
-        )
+        global CHANT_SEARCH_TEMPLATE_VALUES # in order to fetch only those values necessary
```
ahankinson: I don't think you need to use `global` here... It's already in the outer scope.

jacobdgm:
> What is a chant's `search_vector`?

A very good question, and one I've spent several months wondering about and then being confused by. I think I've got it mostly figured out now.

As I understand it: when Postgres is doing a keyword search, it will add an annotation to each record in the table, which takes several of the record's values and uses them to calculate a number for ranking the results in order of relevance. Our BaseChant model has a `search_vector` field to store this information, and it's calculated whenever a Chant (but not a Sequence) is saved/updated, so that Postgres can just use this stored value instead of needing to calculate it during the query evaluation.

As currently calculated, the `search_vector` field includes the chant's incipit, MS spelling and standard spelling fulltexts, and its source's title as heavily weighted components of the score, and the names of its genre, feast and office as lightly weighted components.

Since we want to search for a keyword in only (as I understand it) the two fulltext fields, and since we have the Chant Search page set up to be sortable by clicking on the column headings, we have no need for these ranking features nor for the searching in the incipit, source title, office name fields etc.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: you're quite right - thanks

jacobdgm:
Upon noting that many chants have only an incipit, and neither fulltext field (see #574), I added an extra commit to ensure that such chants also appear among search results.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -716,6 +722,8 @@ def get_context_data(self, **kwargs) -> dict:
         return context
 
     def get_queryset(self) -> QuerySet:
+        CHANT_SEARCH_TEMPLATE_VALUES # in order to fetch only those values necessary
```
ahankinson: This does nothing, AFAIK. Remove.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -933,6 +975,8 @@ def get_context_data(self, **kwargs):
...
```
ahankinson: Same here.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -726,18 +734,32 @@ def get_queryset(self) -> QuerySet:
...
+                    *CHANT_SEARCH_TEMPLATE_VALUES
+                )
+                sequence_set = sequence_set.filter(
+                    q_obj_filter
+                ).values(
+                    *CHANT_SEARCH_TEMPLATE_VALUES
+                )
+                queryset = chant_set.union(sequence_set, all=True)
 
         else:
```
ahankinson, line 767: Github won't let me comment on the lines below, but this line:

```
            if not self.request.GET:
                return Chant.objects.none()
```

It seems like that can go outside of this `if` statement? Since the outer if statement is:

`if self.request.GET.get("search_bar"):`

Then you can return `Chant.objects.none()` before the if statement even happens.

Unless I'm missing something?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -776,56 +798,76 @@ def get_queryset(self) -> QuerySet:
...
-            else:
-                order = 'siglum'
-            if self.request.GET.get('sort'):
-                if self.request.GET.get('sort') == "asc":
-                    order = order
-                elif self.request.GET.get('sort') == 'desc':
-                    order = "-" + order
+
+            if sort_get_param and sort_get_param == "asc":
+                pass # this is the default
```
ahankinson: This is a no-op, so you do not need this clause.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -776,56 +798,76 @@ def get_queryset(self) -> QuerySet:
...
                 q_obj_filter &= Q(feast__in=feasts)
-            if self.request.GET.get('order'):
-                if self.request.GET.get('order') == 'siglum':
-                    order = 'siglum'
-                elif self.request.GET.get('order') == 'incipit':
+            order: str = 'siglum'
+            order_get_param: Optional[str] = self.request.GET.get('order')
+            sort_get_param: Optional[str] = self.request.GET.get('sort')
+            if order_get_param is not None:
+                if order_get_param == 'siglum':
```
ahankinson: You don't need this clause. (`pass` is, to me, code smell. It's useful while developing, but really shouldn't make its way into production code except in very limited circumstances.)

django/cantusdb_project/main_app/templates/chant_search.html
```diff
@@ -209,8 +209,8 @@ <h3>Search Chants</h3>
                     {% for chant in chants %}
                         <tr>
                             <td class="text-wrap" style="text-align:left">
-                                {% if chant.source %}
-                                    <a href="{% url 'source-detail' chant.source.id %}" title="{{ chant.source.title }}">{{ chant.source.siglum }}</a>
+                                {% if chant.source__id %}
```
ahankinson, line 212: I couldn't tell, but is this lookup optimized? I'm trying to remember if this is the case, but generally when doing double-underscore lookups, I like to make sure the target table (`source`, in this case) is added to a `select_related` so that it doesn't have to do the traversal.

(I can't actually test this with data, since I don't have a DB dump, so I'm just trying to go from memory)

django/cantusdb_project/main_app/views/chant.py
jacobdgm: good grief! Yes, I'll remove it.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: Very good point - moving this to the top of the method.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: for this and `if sort_get_param == "asc"` below, I think I was following the "explicit is better than implicit" dictum - I thought I was being helpful for future developers who will see a `?order=siglum` in the links on the template, try to find where this case is treated in the view, and would otherwise get confused when they can't find it. You have a better sense of this than I do, though, so happy to defer to you if you think these no-ops should just be skipped.

Alternatively, would skipping the initial `order: str = 'siglum'`, and setting things up along the lines of
```python
if order_get_param == 'a bunch of options here':
    order = 'one of the many options'
#...
elif order_get_param == 'siglum':
    order = 'siglum'
else:
    order = 'siglum'
```
work better?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -776,56 +798,76 @@ def get_queryset(self) -> QuerySet:
                 # as a substring
                 feasts = Feast.objects.filter(name__icontains=feast)
                 q_obj_filter &= Q(feast__in=feasts)
-            if self.request.GET.get('order'):
-                if self.request.GET.get('order') == 'siglum':
-                    order = 'siglum'
-                elif self.request.GET.get('order') == 'incipit':
+            order: str = 'siglum'
```
ahankinson: I don't think this is used?

django/cantusdb_project/main_app/views/chant.py
ahankinson: You set `order` above anyway, so it's `"siglum"` by default unless it gets changed.

I like the `else:` better, though, since it's more explicit about behaviour.

ahankinson: if `order_get_param` doesn't match anything else in the `if` statement, it's `"siglum"`. So:

```
elif order_get_param == 'siglum':
    order = 'siglum'
else:
    order = 'siglum'
```

is redundant. You don't need the first case, since it would match on the `else` by default anyway.

django/cantusdb_project/main_app/templates/chant_search.html
jacobdgm: > is this lookup optimized?

I'm not sure what you mean.

I did some experiments with this, and couldn't get `select_related` or `prefetch_related` to play nicely with `values`:

If I specify `'source'` among the arguments to `.values`, I'll have access to the source's pk (i.e. the value in the 'source' column) but not to the source object itself. If I use `source` in the template after specifying it in the `values`, I'm worried future developers will think we're working with the source object rather than with its pk. Hence me using `source__id` in both places instead of `source`.

Since we're already fetching `source__title`, `source__siglum`, etc., postgres will already be joining the two tables under the hood, so I don't think fetching `source__id` will cause it to run any slower (i.e. as far as I can tell, it's just the difference between `SELECT chant.source AS 'source' FROM chant SOMEJOIN source ON chant.source = source.id` and `SELECT source.id AS 'source__id' FROM chant SOMEJOIN source ON chant.source = source.id`).

Does this address your question?

ahankinson: I think I would need the data as well to test it. Let's say my spidey-senses were tingling, but I can't verify one way or the other. 

Note that `.values()` returns a [QuerySet of dictionaries](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#values) which is probably why you couldn't get it to work properly. I'm not sure what problem you're trying to solve with using `.values()`, but maybe `.only()` [is useful too](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#only)? This would return your results as objects, so it means you don't dead-end yourself into a list of dictionaries, and can keep spiralling down the ORM rabbit hole. :-D

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -776,56 +798,76 @@ def get_queryset(self) -> QuerySet:
...
+            if order_get_param is not None:
+                if order_get_param == 'siglum':
+                    pass # this is the default
+                elif order_get_param == 'incipit':
                     order = 'incipit'
-                elif self.request.GET.get('order') == 'office':
+                elif order_get_param == 'office':
                     order = 'office'
-                elif self.request.GET.get('order') == 'genre':
+                elif order_get_param == 'genre':
```
ahankinson: Looking a bit more closely at this... this whole block can be removed and mostly replaced with a couple simple checks.

1. Check if the incoming value is in a list of acceptable values
1a. Check the special cases where the incoming value isn't the value we want
1b. Otherwise, set the order value to the incoming value
2. If it isn't in a list of acceptable values, all other values get set to "siglum" (default), including no value.

```py
order_get_param: Optional[str] = self.request.GET.get("order")
if order_get_param in ('incipit', 'office', 'genre', 'cantus_id', 'mode', 'has_fulltext', 'has_melody', 'has_image'):
    if order_get_param == "has_fulltext":
         order = "manuscript_full_text"
    elif order_get_param == "has_melody":
         order = "volpiano"
    elif order_get_param == "has_image":
         order = "image_link"
    else:
         order = order_get_param
else:
    order = 'siglum'
```

django/cantusdb_project/main_app/views/chant.py
ahankinson:
```py
if sort_get_param and sort_get_param == 'desc':
    order = f"-{order}"
```

This has the exact same effect.

django/cantusdb_project/main_app/views/chant.py
ahankinson: >  I thought I was being helpful for future developers who will see a ?order=siglum in the links on the template, try to find where this case is treated in the view, and would otherwise get confused when they can't find it. 

My sense is that this can be captured in a comment, rather than needing to capture it in code. Something like:

"# sort values: Allows 'asc' or 'desc'. Default if not specified is 'asc'."

django/cantusdb_project/main_app/templates/chant_search.html
jacobdgm: ah! So yes, I learned through trial and error that `.values` gives me a bunch of dictionaries. It would be worth me running some experiments to see whether `.only` is as performant as `.values`, as I do like the idea of getting rid of all the double underscores in the template. Thanks for this hint!

django/cantusdb_project/main_app/views/chant.py
jacobdgm: Fair enough - I'll make this change in both places (or perhaps go with your 'looking a bit more closely' suggestion below). I like the idea of self-documenting code, but occasionally go further than necessary with it evidently. Thanks!

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I like this suggestion.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: Curious, regarding the `if` clause itself: Had I not known from comments on past issues that we're accounting for the case where `sort_get_param` is `None`, I would have been confused by this. I can imagine two other ways to write this:
```python
if sort_get_param == 'desc': # shorter and simpler, since None == 'desc' evaluates to False anyways
```
and
```python
if sort_get_param is not None and sort_get_param == 'desc': # more explicit, since we really care about None-ness rather than truthiness
```
Is there a reason you chose a middle ground here? Is this just convention?

ahankinson: Just convention. You rarely find:

```py
if foo is not None:
```

and favour:

```py
if foo:
    # foo is truthy.
```

Unless you're *explicitly* checking against possibly other falsey values:

```py
foo: Optional[bool] = ???  # foo can be None | True | False
if foo is not None:
    # explicit None check makes sure you don't accidentally also capture the case where `foo` is False
```

ahankinson: As for your first option, you can do that too -- sure. You just need to watch out in cases like this:

```py
if foo.get("baz") == "bar":
```

since if `foo` is None, you'll crash and burn. But it's not the case here.

jacobdgm:
I pushed a new commit adding some tweaks based on @ahankinson's feedback. I'll run some experiments with `.only` and `.select_related` etc., and make further changes if it runs just as fast as `.values`.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: with [1ae779a](https://github.com/DDMAL/CantusDB/pull/568/commits/1ae779a97e266e0276e57a3a6ebe36f5a2611d29), this has been replaced by an `else` statement.

jacobdgm:
Running some experiments with this, I'm unable to make `.select_related` and `.only` work together nicely together - it seems I'm not able to traverse related models' fields in the template. This may still be good to look into in more detail - **suggesting that we merge this as-is, and hold `.select_related` and `.only` in mind as a potential thing to refactor** in the future.

(For what it's worth, the `.select_related`+`.only` version ran no faster than the `.values` version, even when objects' names/urls/etc. were not being rendered. So it doesn't appear that switching away from `.values` would further improve page load times - the benefit will mostly be to make the template look a bit cleaner by swapping out the `__`s for `.`s.)

django/cantusdb_project/main_app/templates/chant_search.html
jacobdgm: an update: I was not immediately able to make `select_related` work nicely with `only` - I wasn't able to use the ORM as I expected in the template, so values weren't rendering, and even with many things not rendered, the page didn't load any faster than when I used `.values`. When I asked it to count queries, DjDT counted just as many queries in both versions of the code.

ahankinson: > DjDT counted just as many queries in both versions of the code.

out of curiosity, how many was it?

jacobdgm: 11 queries both times! Our `base.html` template that all the other templates extend causes a handful of queries to run every time a page loads, so while I didn't look in detail, this is likely 10 queries for rendering the footer and 1 for the search results.

jacobdgm:
Seeing as no new objections have been raised in the last 96 hours, and seeing as I'd like to get these changes up on staging in order to see how they behave on the DRA servers, I'm going to go ahead and merge this. There's still room for improvement, especially in the style/formatting department, but this should be good enough for the time being.

PR #576, opened 2023-03-03
Prevent UnboundLocalErrors from occurring

### LINKED ISSUE #574: Chant Edit/Proofread pages - UnboundLocalError
discovered at http://206.12.88.113/edit-chants/123704?pk=316285&folio=003r. Doesn't seem to affect that chant's Detail page, but it is also present at http://206.12.88.113/proofread-chant/123704?pk=316285&folio=003r.

jacobdgm:
This PR provides a default syllabized text in situations where a chant has its `volpiano` set but has no full text (or incipit). It fixes #574, on Chant Proofread and Edit pages, and also prevents a similar error on Chant Detail pages. It also includes some new tests for these views, which only check that the response to a GET request has a 200 code.

Some implementation details:
- On Chant Detail page, we check for MS spelling full text, then standardized spelling full text, then incipit. Finding none of these, we render the volpiano with no text beneath
- On the Chant Edit/Proofread pages, we check all of the fields listed above. If we find none of them, we display a message explaining that there is no saved full text (hopefully this will be helpful for transcribers? Particularly since sometimes, the standardized spelling field will automatically be populated using text pulled (I think) from CI, so this lets them know that that text is not currently saved in the database). See screenshot below:

Screen Shot 2023-03-03 at 3 09 46 PM

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1601,10 +1601,13 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
         chant = self.get_object()
 
         # Preview of melody and text:
-        # in the old CantusDB,
+        # in OldCantus,
         # 'manuscript_syllabized_full_text' exists => preview constructed from 'manuscript_syllabized_full_text'
         # no 'manuscript_syllabized_full_text', but 'manuscript_full_text' exists => preview constructed from 'manuscript_full_text'
         # no 'manuscript_syllabized_full_text' and no 'manuscript_full_text' => preview constructed from 'manuscript_full_text_std_spelling'
+        # to this we add:
+        # no full text of any kind => preview constructed from `incipit`
+        # none of the above => show message explaining why melody preview has no text
 
         if chant.volpiano:
             syls_melody = syllabize_melody(chant.volpiano)
@@ -1621,6 +1624,11 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
             elif chant.manuscript_full_text_std_spelling:
                 syls_text = syllabize_text(chant.manuscript_full_text_std_spelling, pre_syllabized=False)
                 syls_text, syls_melody = postprocess(syls_text, syls_melody)
+            elif chant.incipit:
+                syls_text = syllabize_text(chant.incipit, pre_syllabized=False)
+                syls_text, syls_melody = postprocess(syls_text, syls_melody)
+            else:
+                syls_text = [["(no saved Full Text or Incipit)"]]
 
             word_zip = align(syls_text, syls_melody)
             context["syllabized_text_with_melody"] = word_zip
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -86,9 +86,11 @@ class ChantDetailView(DetailView):
                     chant.manuscript_full_text, pre_syllabized=False
                 )
                 syls_text, syls_melody = postprocess(syls_text, syls_melody)
-            else:
+            elif chant.incipit:
                 syls_text = syllabize_text(chant.incipit, pre_syllabized=False)
                 syls_text, syls_melody = postprocess(syls_text, syls_melody)
+            else:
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -86,9 +86,11 @@ def get_context_data(self, **kwargs):
                     chant.manuscript_full_text, pre_syllabized=False
                 )
                 syls_text, syls_melody = postprocess(syls_text, syls_melody)
-            else:
+            elif chant.incipit:
                 syls_text = syllabize_text(chant.incipit, pre_syllabized=False)
                 syls_text, syls_melody = postprocess(syls_text, syls_melody)
+            else:
+                syls_text = [[""]]
```
dchiller, line 99: It doesn't look like you are checking standard spelling in this view (ChantDetail)...is that right?

annamorphism:
> 
> Screen Shot 2023-03-03 at 3 09 46 PM
Is there a way to do this without the whitespace in the staff?

jacobdgm:
Not really. I could break the error message up into separate words/syllables, which would likely cut down on any gaps, but I can't guess at what the volpano for a given chant will be, so this could cause other problems.

@annamorphism, do you think it would be better without the error message?

django/cantusdb_project/main_app/views/chant.py
jacobdgm: huh - my understanding was that the standard spelling fulltext was not really meant to be displayed with volpiano, but I suppose that is true of the incipit as well. Let me add this - good suggestion!

dchiller: Ok, I think I was responding to:

>  we check for MS spelling full text, then standardized spelling full text, then incipit. Finding none of these, we render the volpiano with no text beneath

and then didn't see it, so sorry if I misunderstood and it's not necessary!

dchiller:
Could you do something like pass a boolean in the context data that is True/False based on whether an error message should display, and then do an if tag with the html for the error message in the template?

And I guess in this case, then, just pass in `""` as the syllables.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I'm a bit confused. I think you identified an actual oversight, and I fixed it in [this commit](https://github.com/DDMAL/CantusDB/pull/576/commits/334573d42cc6d4b7acf2242522a8c0c50cec712f) - is this not the case?

jacobdgm:
> Could you do something like pass a boolean in the context data that is True/False based on whether an error message should display, and then do an if tag with the html for the error message in the template?

So yes, but I think we're getting into enhancement territory here - is this error message actually a thing that anyone wants, or is it just added/unnecessary complication? I added the error message like this because it was simple to implement, but I don't see a reason that we should go out of our way to add and polish a new feature that no one has asked for.

I'm inclined to leave it as-is (i.e. with the simple error message and the gap in the staff), unless either of you strongly feel that it would be better to just pass empty syllables, in which case I'll remove the error message. This will only display on the chant-edit and chant-proofread pages and nowhere outward-facing, so I don't think the stakes are particularly high.

dchiller:
Makes sense. I don't love how the addition of the text makes it look like something unrelated to the error message is broken or a display is malfunctioning. And so I'd say: get rid of the `UnboundLocalError` and just display no text, and if a user-facing message is important, that's an enhancement.

But, you and Anna would know better than I whether the message is critical.

django/cantusdb_project/main_app/views/chant.py
dchiller: No you are correct, you fixed it. I was just explaining why I made the suggestion in the first place (that there seemed to be a discrepancy between your description above and the code), but that if you didn't need to check standardized spelling fulltest, then you didn't need to check standardized spelling fulltext...

so we should consider this, from my perspective, resolved!

jacobdgm:
Pushed a new commit to remove the error message! Now it checks syllabized fulltext, MS spelling fulltext, standard spelling fulltext, and incipit, tries to render the first of those it can find, and if it finds none of them, displays just the volpiano with no text underneath.

dchiller (approved):

PR #582, opened 2023-03-14
Century Detail page: exclude links to source detail pages that user can't visit

### LINKED ISSUE #581: On Century Detail pages, sometimes unpublished sources are listed
Visiting century/3863 while not logged in, there's a link to source/672452. When I click on it, I get a 403 Access Denied error. We need to ensure that links to sources that are inaccessible to a user are never displayed.

This bug occurs on both staging and production.

jacobdgm:
This PR fixes #581.

In all, it:
- Changes `CenturyDetailView` to exclude private sources when the user is not logged in, and tweaks `century_detail.html` to accommodate these changes.
- Adds several tests to `CenturyDetailViewTest` to ensure that all the right sources, and none of the wrong sources, are being displayed on the Century Detail page
- tweaks `make_fake_source`, adding a `century` parameter, to accommodate the new tests just mentioned
  - in doing so, I took the opportunity to reorder some of the parameters/logic so that they're consistent, and add type annotations.

### DIFF
--- a/django/cantusdb_project/main_app/tests/make_fakes.py
+++ b/django/cantusdb_project/main_app/tests/make_fakes.py
@@ -266,25 +266,25 @@ def make_fake_sequence(source=None, title=None, incipit=None, cantus_id=None) ->
 
 
 def make_fake_source(
-        published:Optional[bool] = None,
-        title:Optional[str] = None,
-        segment_name:Optional[str] = None,
-        segment:Optional[Segment] = None,
-        siglum:Optional[str] = None,
-        rism_siglum:Optional[RismSiglum] = None,
-        description:Optional[str] = None,
-        summary:Optional[str] = None,
-        provenance:Optional[Provenance] = None,
-        century:Optional[Century] = None,
-        full_source:Optional[bool] = None,
-        indexing_notes:Optional[str] = None,
+        published: bool = True,
+        title: Optional[str] = None,
+        segment_name: Optional[str] = None,
+        segment: Optional[Segment] = None,
+        siglum: Optional[str] = None,
+        rism_siglum: Optional[RismSiglum] = None,
+        description: Optional[str] = None,
+        summary: Optional[str] = None,
+        provenance: Optional[Provenance] = None,
+        century: Optional[Century] = None,
+        full_source: bool = True,
+        indexing_notes: Optional[str] = None,
     ) -> Source:
     """Generates a fake Source object."""
     # The cursus_choices and source_status_choices lists in Source are lists of
     # tuples and we only need the first element of each tuple
 
-    if published is None:
-        published = True
+    # if published...
+    #     published already defaults to True
     if title is None:
         title = faker.sentence()
     if segment_name is None:
@@ -309,8 +309,8 @@ def make_fake_source(
         provenance = make_fake_provenance()
     if century is None:
         century = make_fake_century()
-    if full_source is None:
-        full_source = True
+    # if full_source...
+    #     full_source already defaults to True
     if indexing_notes is None:
         indexing_notes = make_fake_text(
             max_size=MAX_NUMBER_TEXT_CHARS,
@@ -329,6 +329,7 @@ def make_fake_source(
         description=description,
         summary=summary,
         provenance=provenance,
+        # century: ManyToManyField, must be set below
         full_source=full_source,
         indexing_notes=indexing_notes,
         provenance_notes=make_fake_text(
--- a/django/cantusdb_project/main_app/tests/make_fakes.py
+++ b/django/cantusdb_project/main_app/tests/make_fakes.py
@@ -15,6 +15,8 @@ from main_app.models import Sequence
 from main_app.models import Source
 from django.contrib.auth import get_user_model
 
+from typing import Optional
+
 # Max values for two types of Char fields
 LONG_CHAR_FIELD_MAX = 255
 SHORT_CHAR_FIELD_MAX = 63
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -263,28 +265,56 @@ def make_fake_sequence(source=None, title=None, incipit=None, cantus_id=None) ->
     return sequence
 
 
-def make_fake_source(published=None, title=None, segment=None, segment_name=None, siglum=None, description=None, summary=None, provenance=None, full_source=None, rism_siglum=None, indexing_notes=None) -> Source:
+def make_fake_source(
+        published:Optional[bool]=None,
```
ahankinson: If you don't have your editor configured to highlight the formatting problems here, you should do that. The "canonical" way of typing methods is, e.g.,:

`published: Optional[bool] = None,`

That is, space between the name and the type, and spaces around the `=`. (NB: This is different from the non-typed way: `published=None`.)

https://peps.python.org/pep-0526/#preferred-coding-style-for-variable-annotations

django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -263,28 +265,56 @@ def make_fake_sequence(source=None, title=None, incipit=None, cantus_id=None) ->
...
+        provenance:Optional[Provenance]=None,
+        century:Optional[Century]=None,
+        full_source:Optional[bool]=None,
+        indexing_notes:Optional[str]=None,
+    ) -> Source:
     """Generates a fake Source object."""
     # The cursus_choices and source_status_choices lists in Source are lists of
     # tuples and we only need the first element of each tuple
 
+    if published is None:
```
ahankinson: It also seems your use of `Optional` and default arguments is overlapping. In this case, you're using `published=None`, but then immediately capturing that and setting it to `True`. You can get rid of this by just doing:

`def make_fake_source(published: bool = False ...)`

Which will do exactly the same thing, but without needing the Optional, nor this conversion check. I think this holds for the other boolean fields as well.

django/cantusdb_project/main_app/views/century.py
```diff
@@ -1,7 +1,20 @@
 from django.views.generic import DetailView
-from main_app.models import Century
+from main_app.models import Century, Source
+from typing import Any, Dict
```
ahankinson: Note that if you're using Python >= 3.9, the use of `Dict` from typing has been deprecated. You should just use the built-in `dict`, e.g., `def foo() -> dict[str, Any]`

https://docs.python.org/3/library/typing.html#typing.Dict

django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -263,28 +265,56 @@ def make_fake_sequence(source=None, title=None, incipit=None, cantus_id=None) ->
...
         summary = make_fake_text(
-            max_size=MAX_NUMBER_TEXT_CHARS, min_size=MIN_NUMBER_TEXT_CHARS
+            max_size=MAX_NUMBER_TEXT_CHARS,
+            min_size=MIN_NUMBER_TEXT_CHARS,
+        )
+    if provenance is None:
+        provenance = make_fake_provenance()
+    if century is None:
+        century = make_fake_century()
+    if full_source is None:
```
dchiller: Did we just not create fake data for some of these attributes (eg. `full_source`, `provenance`) before?

django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -293,34 +323,34 @@ def make_fake_source(published=None, title=None, segment=None, segment_name=None
...
         ),
         image_link=faker.image_url(),
-        indexing_notes=indexing_notes,
         indexing_date=make_fake_text(
             max_size=MAX_NUMBER_TEXT_CHARS, min_size=MIN_NUMBER_TEXT_CHARS
         ),
         json_info=None,
     )
-    source.century.set([make_fake_century()])
+    source.century.set([century])
```
dchiller, line 354: Is there a logic behind having some of these things set via `.set()` and some set during the instantiation of the `Source` object? 

It seems that these are the many-to-many fields -- I guess this is to make sure that the associated fake objects are created (ie. that `make_fake_notation()`) is called?

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: Not that I can see? Unless I'm missing something - can you give me a line number?

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: Initially, I tried passing `century=century` as an argument to `Source.objects.create()`, and python gave me error messages, so I followed the pattern I saw in lines 323-326 (pre-modification line numbers). I think you can't just specify a single century because it's a many-to-many field.

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: thanks - I'll update this.

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: I get what you're saying. I like the way things are currently because all the parameters are present, in the same order - the parameters declared at the top of the function, the series of `if` statements here, and the arguments to `Chant.objects.create()` - and they're all handled in the same way. As well, I figure tests are a place where I'd tend to prioritize clarity/consistency over efficiency.

With that said, I could make this nearly as clear by replacing this `if` block with a comment (a change that will, at the very least, prevent future developers from thinking "why was this not done with a default argument in the first place?"). (I should also add placeholder comments for `century`, `provenance`, etc. among the `.create()` arguments below.) So sure, I can make this change.

django/cantusdb_project/main_app/tests/make_fakes.py
dchiller: I mean: you are adding the functionality that the default is that fake data is created....before this PR is would just be `None`

So, no line number because it didn't exist previously.

django/cantusdb_project/main_app/tests/make_fakes.py
ahankinson: It's not so much about efficiency as about data correctness. `published` is a boolean field in the database, so it should only ever be a boolean in your code. Making it an optional boolean means there are then three states for that variable, one of which is not actually a valid value.

django/cantusdb_project/main_app/tests/make_fakes.py
dchiller: Ok -- I think what I'm reacting to here mostly is the sense that there are three places/ways where fake data is being created:
1. Before the instantiation of the source, where argument values are checked, roughly lines 286 - 317 (for example, how `rism_siglum` is created).
2. In the actual instantiation expression of the `Source` object, lines 323 - 352 (for example, `date` or `liturgical_occasions`)
3. After the instantiation of the Source, lines 353 - 356 (for example, `notation`).

It sounds like `.set()` is necessary for those many-to-many attributes (maybe a comment about this is warranted), but I think all the fake data could be created all in one place and that this would make it easier to follow and/or troubleshoot/modify later.

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: ah, I understand what you mean now. Looking through, it appears you're correct - previously, some of these fields would just get set to `None` if no argument was provided for that parameter.

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: I added some comments in the changes I pushed a few minutes ago.

> I think all the fake data could be created all in one place and that this would make it easier to follow and/or troubleshoot/modify later.

I can make this change if you'd like. I think it makes sense as it currently is - we handle all the optional parameters by making fake data/objects if certain parameters haven't been supplied, but otherwise, we make the fake objects/data at the moment they're needed, and no earlier.

Maybe this is symptomatic of these `make_fake_<something>` functions being initially defined without any parameters, and slowly accumulating new parameters where required as the test suite evolved. I think this is true of all the `make_fake`s, and I'm not sure that it's worth refactoring all/any of them.

dchiller: > Maybe this is symptomatic of these `make_fake_<something>` functions being initially defined without any parameters, and slowly accumulating new parameters where required as the test suite evolved. I think this is true of all the `make_fake`s, and I'm not sure that it's worth refactoring all/any of them.

Yes, this has the whiff of that about it. Not ideal, but also, as you say, not worth it to change if this is the way things are.

dchiller (approved):

PR #585, opened 2023-03-14
Full Index/Inventory page - for sources in bower segment, fix broken urls and fix ordering of sequences

### LINKED ISSUE #583: Full Inventory Page, Bower segment: sequences out of order
discovered on staging, http://206.12.88.113/index/?source=123906. Seems to be caused by some folios not having leading zeroes. Sequences, as far as I understand, can be ordered by `s_sequence` alone, rather than by `folio` and then by `c_sequence` (sequences don't have a c_sequence - only chants do)

### LINKED ISSUE #584: Full Inventory Page, Bower Segment: links to sequences broken
discovered on staging, http://206.12.88.113/index/?source=123906.

In the template, we have
```html
<a href="{% url 'chant-detail' chant.id %}" target="_blank">
```
But since the source is from the Bower segment and contains sequences, not chants, django tries to point to the URL of a (nonexistent) chant. We need to check whether the "chant" is a chant or a sequence, and specify `'chant-detail'` or `'sequence-detail'` as necessary.

This could be happening on other pages as well, where we replaced `chant.get_absolute_url` with `url 'chant-detail' chant.id`.

jacobdgm:
Several things are happening in this PR:
- in the Full Index/Inventory view, we annotate the chant set indicating whether each object is a Chant or a Sequence
- In the template, we use these annotations to ensure we generate URLs to the right detail page (chant-detail or sequence-detail) (i.e. fixes #584)
- these annotations are also used to decide between displaying an object's `c_sequence` or `s_sequence`.
- While working in the Full Index view, I took the opportunity to fix the ordering of sequences (for chants, `c_sequence` indicates position/ordering on the folio, whereas for sequences, `s_sequence` indicates its ordering in an entire manuscript) (i.e. fixes #583)

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -5,7 +5,7 @@ import json
 import threading
 from django.contrib import messages
 from django.contrib.postgres.search import SearchQuery, SearchRank
-from django.db.models import F, Q, QuerySet
+from django.db.models import F, Q, QuerySet, Value
 from django.shortcuts import get_object_or_404
 from django.urls import reverse
 from django.views.generic import (
@@ -1389,9 +1389,17 @@ class ChantIndexView(TemplateView):
 
         # 4064 is the id for the sequence database
         if source.segment.id == 4064:
-            queryset = source.sequence_set.order_by("folio", "s_sequence")
+            queryset = source.sequence_set.annotate(
+                is_sequence=Value(True)
+            ).annotate(
+                is_chant=Value(False)
+            ).order_by("folio", "s_sequence")
         else:
-            queryset = source.chant_set.order_by("folio", "c_sequence")
+            queryset = source.chant_set.annotate(
+                is_sequence=Value(False)
+            ).annotate(
+                is_chant=Value(True)
+            ).order_by("folio", "c_sequence")
 
         context["source"] = source
         context["chants"] = queryset
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1393,7 +1393,7 @@ class ChantIndexView(TemplateView):
                 is_sequence=Value(True)
             ).annotate(
                 is_chant=Value(False)
-            ).order_by("folio", "s_sequence")
+            ).order_by("s_sequence")
         else:
             queryset = source.chant_set.annotate(
                 is_sequence=Value(False)
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1390,15 +1390,11 @@ class ChantIndexView(TemplateView):
         # 4064 is the id for the sequence database
         if source.segment.id == 4064:
             queryset = source.sequence_set.annotate(
-                is_sequence=Value(True)
-            ).annotate(
-                is_chant=Value(False)
+                record_type=Value("sequence")
             ).order_by("s_sequence")
         else:
             queryset = source.chant_set.annotate(
-                is_sequence=Value(False)
-            ).annotate(
-                is_chant=Value(True)
+                record_type=Value("chant")
             ).order_by("folio", "c_sequence")
 
         context["source"] = source
--- a/django/cantusdb_project/main_app/templates/full_index.html
+++ b/django/cantusdb_project/main_app/templates/full_index.html
@@ -54,9 +54,9 @@
                     <td>{{ chant.marginalia|default_if_none:"" }}</td>
                     <td>{{ chant.folio|default_if_none:"" }}</td>
                     <td>
-                        {% if chant.is_chant %}
+                        {% if chant.record_type == "chant" %}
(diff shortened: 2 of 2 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1389,9 +1389,17 @@ def get_context_data(self, **kwargs):
 
         # 4064 is the id for the sequence database
         if source.segment.id == 4064:
-            queryset = source.sequence_set.order_by("folio", "s_sequence")
+            queryset = source.sequence_set.annotate(
+                is_sequence=Value(True)
```
ahankinson: Why not a single variable, `record_type='chant' | 'sequence'`.

(I'm a big proponent of data modelling where you "make impossible states impossible". In this case, you would never have `is_chant = True` and `is_sequence = True` since that would be an error. So model your data to make this state impossible!)

If you're interested: 

https://www.youtube.com/watch?v=IcgmSRJHu_8

https://kentcdodds.com/blog/make-impossible-states-impossible

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1389,9 +1389,17 @@ def get_context_data(self, **kwargs):
...
+            ).annotate(
+                is_chant=Value(False)
+            ).order_by("s_sequence")
```
dchiller, line 1394: I see how this fixes the issue in #583. 

For chants, might you run into this same problem, caused by the way folios are sorted without the leading 0? I see Andrew's comment about natural sorting on that issue. 

Probably out of scope for this PR, but I'm curious if that's an issue.

jacobdgm: This is a good point. I kind of see this as more of a data issue, since folios are (I think!) supposed to be of the form `dddl` anyways, where each `d` is a numerical digit and `l` is a letter.

ahankinson: Having worked with manuscripts for a while now, the one thing I'm certain of is that there is no consistency. 

 - Some MSS have +999 pages, so three digits are not enough.
 - Some folios are inserts or flyouts, so they may have an additional qualifiers (a/b) to indicate that they're different. 
 - Many books use pagination instead of foliation (p1, p2, p3 instead of 001r, 001v, 002r)
 - Many books were re-ordered, sometimes even recently, so they might have two folio numbers -- the original and the new. 
 - Many books were re-foliated. This may occur if someone mis-counted and skipped a folio, so someone had to come along later and fix this. Or double-counted.
 - Some books have inserts that are numbered outside of the "main" numbering / foliation, e.g., i, ii, iii, iv, ix, etc. 
 - Some books have been split in two, and the foliation continues in the new book, so they can start at f333r, etc.
 - Some books have been joined into one, so they can have their foliation re-started halfway through!

Many of these may contradict the editorial standards of Cantus, but it's not necessarily a given that all MSS will behave the same way. So I would be careful about trying to standardize their form too much...

ahankinson: In fact, the Salzinnes is an example of a book where the folio numbering re-starts, and has additional qualifiers on the numbering scheme:

folio A22r: https://cantus.simssa.ca/manuscript/25/?folio%5B0%5D=A22r&pageAlias=Folio%20A22r

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I like this suggestion - thanks. Let me make this change.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: So all of this amounts to _folio numbering systems are too variable to try to standardize, so what we have in this PR is fine and let's not try to further tweak the ordering of results_, correct? No further action necessary?

dchiller: Approving, because I agree that we don't want to re-do sorting given these various manuscript-by-manuscript idiosyncrasies. 

I guess I just want to clarify what this PR does and does not "solve." Because of all that's mentioned above about folio naming/numbering, items on the full index page may end up out of order. Since `s_sequence` numbers work the way they do (by manuscript), we can use them with this minor change to correctly order the index page from here on. Because `c_sequence` numbers work differently (by folio), ensuring "proper" ordering would require a more complex change and the decision is not to do that at this point (or maybe ever). Right?

dchiller (approved):

django/cantusdb_project/main_app/views/chant.py
jacobdgm: > I guess I just want to clarify what this PR does and does not "solve." Because of all that's mentioned above about folio naming/numbering, items on the full index page may end up out of order. Since `s_sequence` numbers work the way they do (by manuscript), we can use them with this minor change to correctly order the index page from here on. Because `c_sequence` numbers work differently (by folio), ensuring "proper" ordering would require a more complex change and the decision is not to do that at this point (or maybe ever). Right?

All this is correct!

PR #590, opened 2023-03-17
docker-compose: change postgres version to postgres:14-alpine

jacobdgm:
`postgres:alpine` used to indicate Postgres 14 in dockerhub, now it indicates Postgres 15. Specifying `postgres:14-alpine` should resolve the issue of the two versions of the database being incompatible.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - django
 
   postgres:
-    image: postgres:alpine
+    image: postgres:14-alpine
     env_file: ./config/envs/dev_env
     volumes:
       - postgres_data:/var/lib/postgresql/data/


dchiller (approved):

jacobdgm:
putting this comment here so that I remember where to start once I begin work next week:

Since we didn't specify a version in the dockerfile, it appears that we've been running Postgres 14 on the staging server while I've been running Postgres 15 on my local machine for the past several months (i.e. when I tried to run these changes locally, python/django complained about not being able to open a database created in Postgres 15). In order to bring my local development environment in line with what we're running on staging/production, I'll need to do a fresh setup of CantusDB locally.

When we do the final data sync, we should see what we can do about bringing Postgres as up as possible to date.

jacobdgm:
welp, just checked and production is running postgres 12.3 (i.e. the image has an ID of 17150f4321a3), so this too will need to be updated.

jacobdgm:
trying to keep things straight...
I've got a partially-working copy of CantusDB running locally, running `psql --version` = `14.7`.
Staging is currently running `psql --version` = `14.7` too.
Production is indeed running `psql --version` = `12.3`.

We need to figure out how to update Production while minimizing the risk of losing data.

jacobdgm:
Phew! Having used a copy of the staging database to restore my local setup, I can confirm that `python manage.py runserver` leads to a properly functioning version of the website, and all tests pass. Going to merge this. At last!

Issue #607, opened 2023-04-12
Errors while running test_views


jacobdgm:
The last time I tried to run our test suite, I got a bunch of errors.

Here's the traceback for two of the failing tests; the other failures are similar to one or the other of these:

```python
======================================================================
ERROR: test_provide_suggested_fulltext (main_app.tests.test_views.SourceEditChantsViewTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 445, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 440, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/local/lib/python3.9/http/client.py", line 1377, in getresponse
    response.begin()
  File "/usr/local/lib/python3.9/http/client.py", line 320, in begin
    version, status, reason = self._read_status()
  File "/usr/local/lib/python3.9/http/client.py", line 289, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/requests/adapters.py", line 440, in send
    resp = conn.urlopen(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 755, in urlopen
    retries = retries.increment(
  File "/usr/local/lib/python3.9/site-packages/urllib3/util/retry.py", line 532, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/usr/local/lib/python3.9/site-packages/urllib3/packages/six.py", line 769, in reraise
    raise value.with_traceback(tb)
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 445, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 440, in _make_request
[...]

Issue #587, opened 2023-03-14
pages in the admin area very slow to load


jacobdgm:
related to #208, I tried to upgrade @annamorphism's permissions through the admin page on staging, and got a variety of 50x errors (severe enough that it prevented me from updating the permissions) when I tried to visit the admin user detail page. Similar results occurred when I tried to do the same thing on production. Debra/others are going to need to change objects via the admin panel, so this needs to be sorted out sooner rather than later.

ahankinson:
Usually this is caused by lots of SQL queries for complex data models.

Admin models have a `get_queryset` method you can override and add in `select_related` which can help massively.

See, for example: https://github.com/diamm/diamm/blob/master/diamm/admin/data/item.py#L139-L142

jacobdgm:
Thanks for these pointers, @ahankinson!

Some general comments, for helping me/a future developer unpick this: Testing locally, I'm able to load Anna's page, but it takes ~1min (as compared to the page for my own account, which takes a second or two). The main difference between our accounts is that Anna has dozens of sources for which she is listed as a current editor.

We already have some customization of the User Change page, to allow listing of sources, so I'll do some reading of docs and some experimenting to try to sort this out.

jacobdgm:
> Usually this is caused by lots of SQL queries for complex data models.

You don't say...
![Screenshot 2023-03-15 at 2 41 44 PM](https://user-images.githubusercontent.com/58090591/225412096-dad59da1-4695-4971-9272-c9a84fe98ebe.png)

Glancing through the queries, it seems that in generating the list of User-Source relations, we're getting lots of similar/duplicate queries to the RismSiglum table, since sources' `rism_siglum` is being used to generate their `__str__`.

ahankinson:
Yeah, you should be able to clear that up with a `select_related`. You can do through relationships there as well with the double-underscores -- `select_related("source__rism_siglum")` might help?

jacobdgm:
running some experiments locally in the Admin Area using DjDT to establish a baseline and identify the areas that are particularly problematic:
- Articles
  - Articles
    - "Select article to change" page: 5 queries, 12.63ms
    - "Change article" page: 5 queries, 6.65ms
- Authentication and Authorization
  - Groups
    - "Select group to change": 5, 13.28
    - "Change group": 6, 10.30
- Flat Pages
  - Flat Pages
    - "Select flat page to change": 6, 9.10
    - "Change flat page": 6, 8.99
- Main App
[...]

PR #608, opened 2023-04-12
Change string representation of Source model

jacobdgm:
This PR tweaks the string representation of Source objects in our database, switching from using a source's `rism_siglum` to its `siglum`. This is good for two reasons:
- the `siglum` (really the siglum for an institution + shelfmark) has more information than the `rism_siglum`, which only identifies the institution the MS is held at - we should really have been displaying it anyways
- the `siglum` is a plain old field in the Source table in our database, whereas rism_siglum is a foreign key. Testing locally, using the non-foreign-key field improves page load times on Administrate User page by about a factor of 3. (The Administrate User page is still much slower than it should be - this is more of an attempt to sidestep the problems going on on that page, rather than than a proper f\ix for #587)

Once we get #607 sorted, we can test that this hasn't broken anything and it should be good to merge.


jacobdgm:
Tests pass! This PR's ready for review!

lucasmarchd01 (approved):

PR #619, opened 2023-04-27
Replace instances of http with https where possible

jacobdgm:
This PR updates requests to Cantus Index's APIs, changing the URL to use `https://` instead of `http://`.

Along the way
- a number of other URLs in the views and templates were updated to use `https` as well (though some of the sites we use/link to are not set up to use https, so several instances of `http://` remain).
- I added `timeout=5` arguments to all our `request.get()` calls - pylint was complaining about this, and looking into it, it makes sense that we shouldn't want any of our pages to hang indefinitely if the request we're waiting for never arrives. (I didn't bother trying to update it in the place where we make a request in a separate thread)
- I updated the formatting in several places - some lists of arguments have been split up onto multiple lines, and some instances of `"string".format()` have been replaced with `f"string"`s


lucasmarchd01 (commented):
Please review the conflicts in  django/cantusdb_project/main_app/views/chant.py

jacobdgm:
> Please review the conflicts in django/cantusdb_project/main_app/views/chant.py

Done! Thanks!

dchiller (approved):

PR #622, opened 2023-05-01
Update all files to conform to black

jacobdgm:
This PR is the result of running `black .` in the root directory. I'm a bit worried that formatting of comments etc. may be messed up in a few places by having blindly run the script, but this is probably preferable to spending several days manually checking all of the script's changes.


django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -77,7 +66,10 @@ def test_syllabify_word(self):
             ("triumphas", ["tri", "um", "phas"]),
             ("phylosophia", ["phy", "lo", "so", "phi", "a"]),
             # fl
-            ("flores", ["flo", "res"],),
+            (
+                "flores",
+                ["flo", "res"],
```
jacobdgm, line 125: This confused me for a while, but it appears that the trailing comma after `["flo", "res"]` caused black to split this onto multiple lines whereas the previous tuples were kept on one line. Working to standardize this currently.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -555,17 +563,19 @@ def change_password(request):
...
     else:
         form = PasswordChangeForm(request.user)
         if request.user.changed_initial_password == False:
             messages.warning(
                 request,
-                "The current password was assigned to you by default and is unsecure. Please make sure to change it for security purposes."
+                (
+                    "The current password was assigned to you by default and is unsecure. "
+                    "Please make sure to change it for security purposes."
+                ),
```
jacobdgm, line 575: @dchiller or @lucasmarchd01, please verify that I've done this correctly!

dchiller: Oops....what am I verifying?

jacobdgm (commented):
I've gone through all the changes black made, and fixed the places where there were inconsistencies and/or where black moved comments from useful places into non-useful places.

django/cantusdb_project/main_app/views/views.py
dchiller: It is!

dchiller (approved):

PR #627, opened 2023-05-09
Hide unnecessary fields in admin area - fixes #604

### LINKED ISSUE #604: Hide certain fields in admin area
Several of our models' fields are designed to improve database operation, or to preserve an archive of past information, and should not be edited by users. These fields include `json_info` (present in several models - i.e. the response from OldCantus from when we last synced the data), `next_chant` (Chant, Sequence), `is_last_chant_in_feast` (Chant, Sequence) `number_of_chants` and `number_of_melodies` (Source); there may be others.

We can `exclude` these fields in our Admin configuration - this shouldn't be too hard to fix.

lucasmarchd01:
Hide unnecessary fields in admin area

- `BaseModelAdmin` class contains updated list of fields to ignore
- `ChantAdmin` excludes `s_sequence`, `SequenceAdmin` excludes `c_sequence`
- `base_chant` includes verbose name argument to Hide unnecessary fields in admin area

Fixes #604

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -6,7 +6,7 @@ from main_app.models import *
 
 class BaseModelAdmin(admin.ModelAdmin):
     # these fields should not be editable
-    exclude = ("created_by", "last_updated_by")
+    exclude = ("created_by", "last_updated_by", "json_info", "col1", "col2", "col3", "next_chant", "number_of_chants", "number_of_melodies")
 
     # if an object is created in the admin interface, assign the user to the created_by field
     # else if an object is updated in the admin interface, assign the user to the last_updated_by field
@@ -23,8 +23,7 @@ class CenturyAdmin(BaseModelAdmin):
 
 
 class ChantAdmin(BaseModelAdmin):
-    pass
-
+    exclude = BaseModelAdmin.exclude + ("s_sequence",)
 
 class FeastAdmin(BaseModelAdmin):
     pass
@@ -55,8 +54,7 @@ class SegmentAdmin(BaseModelAdmin):
 
 
 class SequenceAdmin(BaseModelAdmin):
-    pass
-
+    exclude = BaseModelAdmin.exclude + ("c_sequence",)
 
 class SourceAdmin(BaseModelAdmin):
     # from the Django docs:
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/admin.py
```diff
@@ -6,7 +6,7 @@
 
 class BaseModelAdmin(admin.ModelAdmin):
     # these fields should not be editable
-    exclude = ("created_by", "last_updated_by")
+    exclude = ("created_by", "last_updated_by", "json_info", "col1", "col2", "col3", "next_chant", "number_of_chants", "number_of_melodies")
```
jacobdgm: the `"colN"`s will need to be displayed and editable on SequenceAdmin pages. These should just be excluded from ChantAdmin pages.

django/cantusdb_project/main_app/admin.py
```diff
@@ -23,8 +23,7 @@ class CenturyAdmin(BaseModelAdmin):
 
 
 class ChantAdmin(BaseModelAdmin):
-    pass
-
+    exclude = BaseModelAdmin.exclude + ("s_sequence",)
```
jacobdgm: Since `ChantAdmin` already inherits `exclude` from `BaseModelAdmin`, I don't have a clear sense of whether this line will lead to unexpected effects.

A safer approach might be to define `EXCLUDE = ["created_by", "last_updated_by", "json_info"]` at the top of the file, and then in each `ModelAdmin`, add additional elements: i.e. here, it would look like `exclude = EXCLUDE + ["col1", "col2", "col3", "next_chant", "s_sequence"]`.

django/cantusdb_project/main_app/models/base_chant.py
```diff
@@ -30,9 +30,9 @@ class Meta:
     )
     # The "s_sequence" char field, used for Sequences, is used to indicate the relative positions of sequences on the page.
     # It sometimes features non-numeric characters and leading zeroes, so it's a CharField.
-    s_sequence = models.CharField(blank=True, null=True, max_length=255)
+    s_sequence = models.CharField("Sequence", blank=True, null=True, max_length=255)
```
jacobdgm, line 33: This is good!

jacobdgm (changes requested):

django/cantusdb_project/main_app/admin.py
jacobdgm: we probably also need to add `is_last_chant_in_feast` to this list, for both Chants and Sequences

jacobdgm (approved):
Solid!

PR #646, opened 2023-05-18
Wrap external API calls in `try/except` blocks

### LINKED ISSUE #628: SSLError preventing Chant Detail page from loading
Testing locally, I just experienced an SSLError that prevented a Chant Detail page from loading (i.e. I got a debug screen instead of the page I was expecting). When I tried to visit the CantusIndex API specified in the error message directly in my browser (https://cantusindex.org/json-con/201925), I got a useful response, and subsequent reloads of the Chant Detail page worked successfully (in hindsight, I should have taken a screenshot or copied the error message before trying to reload).

This is the same type of error that seems to be causing #607.

This could probably be solved by surrounding the `requests.get` calls in our views in `try/except` blocks, providing alternative empty values in case CantusIndex fails to return a useful response.

jacobdgm:
This PR ensures that exceptions raised by calls to external APIs are caught in all our views. This ensures that if an error occurs while making a request to another website, or if the website returns an unexpected response, we can supply a plausible value to functions downstream rather than the entire page not loading. Fixes #628.

Along the way, several related changes occurred:
- Whenever we catch an exception while trying to fetch/process information from an external API, a brief error message is printed to the terminal. Eventually, once we set up logging, these print statements can be sent to our logs instead.
- A short error message to be displayed on the Chant Detail page instead of the concordances has been added, when we encounter an error while fetching these.
- The concordances summary table on the Chant Detail page has been refactored, so that the bulk of the generation occurs in the template (rather than in python in the view, as had previously occurred)
- Black decided to reformat a couple of things

Credit to @rebmizrahi, who contributed code during several pair programming sessions, especially for the refactoring of the Concordances Summary.

## Screenshots
Concordances summary table, before the changes in this PR (screenshot taken on [staging](http://206.12.88.113/chant/689869)):
previously

Concordances summary table, following the changes in this PR (should be the same):
updated

The error message shown when concordances fail to load (note that "Refresh concordances" shows up at the bottom of the full table of concordances, even if concordances load properly):
error message

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -4,6 +4,7 @@ from main_app.models import *
 # these fields should not be editable by all classes
 EXCLUDE = ("created_by", "last_updated_by", "json_info")
 
+
 class BaseModelAdmin(admin.ModelAdmin):
     exclude = EXCLUDE
 
@@ -22,7 +23,15 @@ class CenturyAdmin(BaseModelAdmin):
 
 
 class ChantAdmin(BaseModelAdmin):
-    exclude = EXCLUDE + ("col1", "col2", "col3", "next_chant", "s_sequence", "is_last_chant_in_feast")
+    exclude = EXCLUDE + (
+        "col1",
+        "col2",
+        "col3",
+        "next_chant",
+        "s_sequence",
+        "is_last_chant_in_feast",
+    )
+
 
 class FeastAdmin(BaseModelAdmin):
     pass
@@ -54,6 +63,7 @@ class SegmentAdmin(BaseModelAdmin):
 
 class SequenceAdmin(BaseModelAdmin):
     exclude = EXCLUDE + ("c_sequence", "next_chant", "is_last_chant_in_feast")
+
 
 class SourceAdmin(BaseModelAdmin):
     # from the Django docs:
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -23,6 +23,9 @@ class CenturyAdmin(BaseModelAdmin):
 
 
 class ChantAdmin(BaseModelAdmin):
+    list_display = ("incipit", "siglum", "genre")
+    search_fields = ("title", "incipit", "cantus_id")
+    list_filter = ("genre",)
     exclude = EXCLUDE + (
         "col1",
         "col2",
--- a/django/cantusdb_project/main_app/models/base_chant.py
+++ b/django/cantusdb_project/main_app/models/base_chant.py
@@ -32,8 +32,8 @@ class BaseChant(BaseModel):
     # It sometimes features non-numeric characters and leading zeroes, so it's a CharField.
     s_sequence = models.CharField("Sequence", blank=True, null=True, max_length=255)
     # The "c_sequence" integer field, used for Chants, similarly indicates the relative positions of chants on the page
-    c_sequence = models.PositiveIntegerField("Sequence",
-        help_text='Each folio starts with "1"', null=True, blank=True
+    c_sequence = models.PositiveIntegerField(
+        "Sequence", help_text='Each folio starts with "1"', null=True, blank=True
     )
     genre = models.ForeignKey("Genre", blank=True, null=True, on_delete=models.PROTECT)
     rubrics = models.CharField(blank=True, null=True, max_length=255)
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -205,8 +205,38 @@
             </dl>
             {% if chant.cantus_id %}
                 <h4 id="concordances">List of concordances</h4>
-                <small>{{ concordances_summary | safe }}</small>
-                {% if concordances %}
+                {% if concordances_loaded_successfully %}
+                    <small>
+                        Displaying <b>{{ concordances_count }}</b> concordance{{ concordances_count|pluralize }} from the following database{{ concordances_databases|pluralize }}
+                        (Cantus ID <b><a href="http://cantusindex.org/id/{{ chant.cantus_id }}" target="_blank" title="{{ chant.cantus_id }} on Cantus Index">{{ chant.cantus_id }}</a></b>):
+                        <table id="concordancesSummary" style="margin-bottom: 10px">
+                            <tbody>
+                                {% for database in concordances_databases %}
+                                    <tr>
+                                        <td>
+                                            <a href="{{ database.base_url }}" target="_blank">
+                                                {{ database.name }}
+                                            </a>
+                                            {% if database.initialism %}
+                                                ({{ database.initialism }})
+                                            {% endif %}
+                                        </td>
+                                        <td>
+                                            {% if database.alternate_results_count_text %}
+                                                <a href="{{ database.results_url }}" target="_blank" class="guillemet">
+                                                    {{ database.alternate_results_count_text }}
+                                                </a>
+                                            {% else %}
+                                                <a href="{{ database.results_url }}" target="_blank"><b>
+                                                    {{ database.results_count }}</b> chant{{ database.results_count|pluralize }}
+                                                </a>
+                                            {% endif %}
+                                        </td>
+                                    </tr>
+                                {% endfor %}
+                            </tbody>
+                        </table>
+                    </small>
                     <table id="concordanceTable" class="table table-bordered table-sm small" style="table-layout: fixed; width: 100%;">
                         <thead>
                             <tr>
@@ -265,6 +295,10 @@
                             {% endfor %}
                         </tbody>
                     </table>
+                {% else %}
+                    <p>Cantus Database encountered an error while loading concordances
+                        (Cantus ID <b><a href="http://cantusindex.org/id/{{ chant.cantus_id }}" target="_blank" title="{{ chant.cantus_id }} on Cantus Index">{{ chant.cantus_id }}</a></b>).
+                    </p>
                 {% endif %}
                 <p>
                     <a href="#concordances" onclick="window.location.href = '#concordances'; window.location.reload(true);">Refresh concordances</a><br>
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -206,65 +206,69 @@
             {% if chant.cantus_id %}
                 <h4 id="concordances">List of concordances</h4>
                 <small>{{ concordances_summary | safe }}</small>
-                <table id="concordanceTable" class="table table-bordered table-sm small" style="table-layout: fixed; width: 100%;">
-                    <thead>
-                        <tr>
-                            <th scope="col" class="text-wrap" style="width:15%" title="Siglum">Siglum</th>
-                            <th scope="col" class="text-wrap" style="width:6%" title="Folio">Folio</th>
-                            <th scope="col" class="text-wrap" style="width:23%" title="Incipit">Incipit</th>
-                            <th scope="col" class="text-wrap" style="width:4%" title="Office"></th>
-                            <th scope="col" class="text-wrap" style="width:4%" title="Genre"></th>
-                            <th scope="col" class="text-wrap" style="width:4%" title="Position"></th>
-                            <th scope="col" class="text-wrap" style="width:20%" title="Feast">Feast</th>
-                            <th scope="col" class="text-wrap" style="width:7%" title="Mode">Mode</th>
-                            <th scope="col" class="text-wrap" style="width:8%" title="Image Link">Image</th>
-                            <th scope="col" class="text-wrap" style="width:9%" title="Database">DB</th>
-                        </tr>
-                    </thead>
-                    <tbody>
-                        {% for chant in concordances %}
+                {% if concordances %}
+                    <table id="concordanceTable" class="table table-bordered table-sm small" style="table-layout: fixed; width: 100%;">
+                        <thead>
                             <tr>
-                                <td>
-                                    <a href="{{ chant.srclink }}" target="_blank">
-                                    {{ chant.siglum }}
-                                    </a>
-                                </td>
-                                <td>
-                                    {{ chant.folio }}
-                                </td>
-                                <td>
-                                    <a href="{{ chant.chantlink }}" target="_blank">
-                                        {{ chant.incipit }}
-                                    </a>
-                                </td>
-                                <td>
-                                    {{ chant.office }}
-                                </td>
-                                <td>
-                                    {{ chant.genre }}
-                                </td>
-                                <td>
-                                    {{ chant.position }}
-                                </td>
-                                <td>
-                                    {{ chant.feast }}
-                                </td>
-                                <td>
-                                    {{ chant.mode }}
-                                </td>
-                                <td>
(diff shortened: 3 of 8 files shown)


jacobdgm:
Also: update regarding `test_views` - just ran the test suite and we had 2 failures and 0 errors! making progress!

update: just it again after making some changes - we're back to 2 failures, 3 errors...

update to the update: while refactoring, I forgot to copy and paste some of the necessary logic from certain new `except` blocks. This is fixed with commit [c5c493c](https://github.com/DDMAL/CantusDB/pull/646/commits/c5c493caf8deb51328fdacf026626549b7d8ced1)!

django/cantusdb_project/main_app/admin.py
lucasmarchd01, line 1: I'm pretty sure these changes are already published on the develop branch from a previous merge. I'm not sure if this will cause conflicts with the develop branch since it should be the same but if it does you may need to pull the latest changes first.

django/cantusdb_project/main_app/models/base_chant.py
lucasmarchd01, line 1: Same as above

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -358,12 +358,45 @@ def get_context_data(self, **kwargs):
...
+                    cantus_id (string): a Cantus ID
+
+                Returns:
+                    None
+                """
+                try:
+                    requests.get(
+                        url=f"https://cantusindex.org/json-con/{chant.cantus_id}/refresh",
+                        timeout=1,
+                        verify=False,  # since we're not doing anything with the data from
```
dchiller: I get why we don't *need* to have a valid SSL, but what's the advantage of not doing the verification?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -112,11 +113,26 @@ def get_context_data(self, **kwargs):
...
-                f"https://cantusindex.org/json-con/{chant.cantus_id}",
-                timeout=5,
-            )
-            concordances = json.loads(response.text[2:])
+            try:
+                response = requests.get(
+                    f"https://cantusindex.org/json-con/{chant.cantus_id}",
+                    timeout=5,
+                )
+                concordances = json.loads(response.text[2:])
```
dchiller: Took me testing this to understand why the string slicing is happening....so maybe a good spot for a comment?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1139,45 +1097,74 @@ def get_suggested_chants(self):
...
-            genre_name = chant_dict["genre"]
-            genre_id = Genre.objects.get(name=genre_name).id
-            chant_dict["genre_id"] = genre_id
-            return chant_dict
+            try:
+                response = requests.get(
+                    f"https://cantusindex.org/json-cid/{cantus_id}",
+                    timeout=5,
+                )
+                assert response.status_code == 200
```
dchiller: I'm not sure it's worth refactoring these and the other try/except wrappers, but I don't love the bundling of all these possible exceptions together...if you ever were trying to use these to figure out what happened: did the server not respond? Was the result mal-formed? You're also doing a database query of your own database within the try...

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1312,11 +1299,20 @@ def get_context_data(self, **kwargs):
                 "ghisp": "All",
                 "page": page,
             }
-            page = requests.get(
-                "https://cantusindex.org/search",
-                params=p,
-                timeout=5,
-            )
+            try:
```
dchiller, line 1312: Since I don't see a change to this template, I'm curious what the user will see if this View fails?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1632,15 +1628,25 @@ def get_chants_with_folios(chants_in_feast):
...
-            request = requests.get(
-                f"https://cantusindex.org/json-cid/{cantus_id}",
-                timeout=5,
-            )
-            request_text = json.loads(request.text[2:])
-            if request_text:
-                context["suggested_fulltext"] = request_text[0]["fulltext"]
-            else:
-                context["suggested_fulltext"] = ""
+            try:
```
dchiller, line 1641: Same as above....do we need to put something in this template that shows up in case suggested_fulltext = "" because of the failure?

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -205,66 +205,104 @@ <h3>{{ chant.incipit }}</h3>
...
-                            <th scope="col" class="text-wrap" style="width:4%" title="Position"></th>
-                            <th scope="col" class="text-wrap" style="width:20%" title="Feast">Feast</th>
-                            <th scope="col" class="text-wrap" style="width:7%" title="Mode">Mode</th>
-                            <th scope="col" class="text-wrap" style="width:8%" title="Image Link">Image</th>
-                            <th scope="col" class="text-wrap" style="width:9%" title="Database">DB</th>
-                        </tr>
-                    </thead>
-                    <tbody>
-                        {% for chant in concordances %}
+                {% if concordances_loaded_successfully %}
```
dchiller, line 208: Yay! This is much better than before!

django/cantusdb_project/main_app/admin.py
dchiller: These must be different changes because the changes you are seeing are with the current `develop` branch.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: Ah, there's a comment in one other place in the code about this. I can copy it

django/cantusdb_project/main_app/views/chant.py
jacobdgm: This way, we're less likely to have it raise an SSLError or an SSLEOFError? But I guess if we're `except Exception`ing, this line is unnecessary...

django/cantusdb_project/main_app/views/chant.py
jacobdgm: This is a good point... let me think about this

django/cantusdb_project/main_app/views/chant.py
jacobdgm: In the situation where we hit an SSLError or Timeout on the first page, `if len(cantus_id) == 0` will be True, so `context["results"]` will be set to `[["No results", "No results", "No results"]]`. If we hit an error on subsequent pages, we just `break` out of the `for` loop, and we only display the results we've parsed so far in previous passes through the `for`.

Do you think this needs a comment?

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I think the template will look for "suggested_fulltext" in the context, not find it, and display nothing? I guess it could be better if we spell this out explicitly rather than handling it implicitly as is the case now.

jacobdgm: never mind, we already set `context["suggested_fulltext"]` five lines up. i don't think any changes are necessary here.

jacobdgm: I think if we don't have a suggested fulltext, we just don't suggest a fulltext (i.e. our suggestion should be `""`). It would get in people's ways if we added a different message.

dchiller: I guess the reason I ask is if the editor could just refresh the page and get a suggested text (because CantusIndex decided to work this time) is it worth telling them that?

Not a big deal, and I guess this keeps it agreeing with current Cantus

django/cantusdb_project/main_app/views/chant.py
dchiller: No, I think fine for the same reason below.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I've refactored this, and split it up into a few different `except` cases - the difference is only in the error message printed to the console, but things should be a little bit clearer now.

django/cantusdb_project/main_app/views/chant.py
jacobdgm: This is done!

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I've removed this argument - good suggestion!

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -358,12 +296,42 @@ def get_context_data(self, **kwargs):
             # https://cantusindex.org/json-con/{cantus_id}/refresh in a new thread,
             # thus ensuring the current page doesn't take a long time to load, while also
             # ensuring that concordances are up-to-date for future page loads.
+
+            def refresh_ci_json_con_api(cantus_id: str) -> None:
```
ahankinson: Is this a nested function?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -128,228 +156,138 @@ def get_context_data(self, **kwargs):
...
-                    <tr>
-                        <td>
-                            <a href="http://fragmenta.zti.hu/en/" target="_blank">Fragmenta Manuscriptorum Musicalium Hungariae</a> (FRH)
-                        </td>
-                        <td>
-                            <a href="http://fragmenta.zti.hu/en/id/{chant.cantus_id}" target="_blank">{chant_tally}</a>
-                        </td>
-                    </tr>
-                """
+            context["concordances_databases"] = list(
```
ahankinson: You can also do `[d["results_count"] for d in concordances_databases if "results_count" in d and d["results_count"] > 0]`. This has a couple advantages: You don't need to cast it to a list; if "results_count" isn't in d then it won't fail, and it makes the 'falsyness' of 0 explicit, rather than implicit.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -128,228 +156,138 @@ def get_context_data(self, **kwargs):
...
+                    "while making a request to,"
+                    f"https://gregorien.info/chant/cid/{chant.cantus_id}/en:",
+                    exc,
+                )
 
-            if concordances_count:
-                context["concordances_summary"] += "</table><br>"
+            # tally from all databases
+            concordances_count = len(concordance_chants)
+            context["concordances_count"] = concordances_count
```
ahankinson:
```suggestion
            context["concordances_count"] = len(concordance_chants)
```

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1139,45 +1107,84 @@ def get_suggested_chants(self):
...
+                Timeout,
+            ) as exc:
+                print(  # eventually, we should log this rather than printing it to the console
+                    "Encountered an error in",
+                    "ChantCreateView.get_suggested_chants.make_suggested_chant_dict",
+                    f"while making a request to https://cantusindex.org/json-cid/{cantus_id}:",
+                    exc,
+                )
+                cid_dict = {}
+            except (
```
ahankinson: The JSON Loading should be contained in it's own try/except. You should be very minimal in what goes in the body of the `try` and not do too many things in one try statement.

"Additionally, for all try/except clauses, limit the try clause to the absolute minimum amount of code necessary. Again, this avoids masking bugs"

https://peps.python.org/pep-0008/

django/cantusdb_project/main_app/views/chant.py
ahankinson: I find the `assert` to be weird. You might consider instead using the context manager version of the `requests` get... 

```
with requests.Session() as s:
    s.get(...url...)
```

https://requests.readthedocs.io/en/latest/user/advanced/#session-objects

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -128,228 +156,138 @@ def get_context_data(self, **kwargs):
...
+            concordances_databases = [
+                {
+                    "name": "Cantus Database",
+                    "initialism": "CD",
+                    "base_url": "/",
+                    "results_url": f"{reverse('chant-by-cantus-id', args=[chant.cantus_id])}",
+                    "results_count": len(
+                        [True for c in concordance_chants if c["db"] == "CD"]
+                    ),
+                    "alternate_results_count_text": None,
```
ahankinson: I think you can just omit this from the dictionary altogether, since it is usually `None`, and only set it in the places where you need it...

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -128,228 +156,138 @@ def get_context_data(self, **kwargs):
...
+            #   Displaying N concordances from the following databases:
+            #   Cantus Database (CD) - N chants
+            #   Fontes Cantus Bohemiae (FCB) - N chants
+            #   etc...
+            #   Gregorien.info - VIEW AT GREGORIEN.INFO
+            concordances_databases = [
+                {
+                    "name": "Cantus Database",
+                    "initialism": "CD",
+                    "base_url": "/",
```
ahankinson, line 276: Use the `reverse()` method?

ahankinson (approved):

django/cantusdb_project/main_app/views/chant.py
jacobdgm: I like this suggestion.

But, we don't have the homepage listed in any of our `urls.py`s. Several of us spent a while trying to figure out where the `/` path is being defined, plugging different plausible arguments (and lack of arguments) into `reverse()`, and our best guess is that the `/` path is maybe being defined/generated in the flatpages app.

If you know off the top of your head how to make this work, I'll happily make this change, but it's looking like more trouble than it's worth at the present moment.

ahankinson: `python manage.py show_urls`

You need the `django-extensions` module installed and configured.

jacobdgm:
Re-requesting reviews - I just finished a bunch of refactoring, which mostly involved moving nested functions to the top of the file, and also breaking up several `try/except` blocks so that we `try` just one or two lines of code at a time rather than a whole bunch all at once.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -95,12 +215,40 @@ def get_context_data(self, **kwargs):
             context["syllabized_text_with_melody"] = text_and_mel
 
         # If chant has a cantus ID, Create table of concordances
+        context["concordances_loaded_successfully"] = True
         if chant.cantus_id:
```
ahankinson, line 219: This `if` statement is still really long, so it's hard to see what it's doing. Any chance you could break this out into a separate function?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1619,18 +1638,37 @@ def get_chants_with_folios(chants_in_feast):
...
             cantus_id = current_chant.cantus_id
 
-            request = requests.get(
-                f"https://cantusindex.org/json-cid/{cantus_id}",
-                timeout=5,
-            )
-            request_text = json.loads(request.text[2:])
-            if request_text:
-                context["suggested_fulltext"] = request_text[0]["fulltext"]
+            try:
```
lucasmarchd01, line 1641: I see this try/except code repeated a few times, I wonder if it would be possible/worth putting this in a separate function

lucasmarchd01 (approved):

PR #647, opened 2023-05-18
Auto-populate Syllabized Full Text field

### LINKED ISSUE #610: On Edit Syllabification page, auto-populate Syllabized Full Text field
As noted by @annamorphism on #606, On the Edit Syllabification page, where there is no existing Syllabized Full Text, we should auto-populate this field on the page to give editors a starting point in creating/modifying the Syllabized Full Text. We can do this by syllabizing the MS Spelling Full Text field (or the STD Spelling field, or the Incipit, depending on what's available) using `syllabize_text`.

OldCantus already has this feature, so we should have it too.

lucasmarchd01:
This PR addressed issue #610 where the "Edit Syllabification" view shows an empty field for the syllabized text. The modifications allow this field to be populated by the syllabification generated from the full text (or other resources available if `manuscript_full_text` is not provided)

- add `get_initial()` function to `ChantEditSyllabificationView` to initialize `manuscript_syllabized_full_text` 
- modify `chant.py` by adding method `syllabize_text_and_melody` (used for populating the volpiano section)
- modify `chant.py` by adding method to `syllabize_text_to_string` (used for populating the manuscript_syllabized_full_text field)
- remove redundant code in chant views
- fixes #610 

Example syllabification:

image

### DIFF
--- a/django/cantusdb_project/align_text_mel.py
+++ b/django/cantusdb_project/align_text_mel.py
@@ -327,7 +327,7 @@ def syllabize_text_and_melody(text, pre_syllabized, melody):
 
     Args:
         text (str): The input text to be syllabized.
-        pre_syllabized (list): A list of pre-syllabized words corresponding to the input text.
+        pre_syllabized (bool): A list of pre-syllabized words corresponding to the input text.
         melody (list): A list representing the melody.
     Returns:
         list: A list of tuples containing the aligned words and syllables.
--- a/django/cantusdb_project/align_text_mel.py
+++ b/django/cantusdb_project/align_text_mel.py
@@ -314,3 +314,40 @@ def align(syls_text, syls_melody):
         list_of_zips.append(word_zip)
 
     return list_of_zips
+
+
+def syllabize_text_and_melody(text, pre_syllabized, melody):
+    """Syllabizes the given text and melody, aligns the syllables, and returns a zip of words and syllables.
+
+    This function takes a text, pre-syllabized text, and a melody as input. It performs syllabization on the text
+    and melody separately based on the volpiano protocols. The syllabized text and melody are then postprocessed,
+    and syllables of the text and melody are aligned to create a zip
+    of words and syllables.
+
+
+    Args:
+        text (str): The input text to be syllabized.
+        pre_syllabized (list): A list of pre-syllabized words corresponding to the input text.
+        melody (list): A list representing the melody.
+    Returns:
+        list: A list of tuples containing the aligned words and syllables.
+    """
+    syls_melody = syllabize_melody(melody)
+    syls_text = syllabize_text(text, pre_syllabized=pre_syllabized)
+    syls_text, syls_melody = postprocess(syls_text, syls_melody)
+    word_zip = align(syls_text, syls_melody)
+    return word_zip
+
+
+def syllabize_text_to_string(text, pre_syllabized):
+    """Syllabizes the given text using pre-syllabized information and returns a human-readable string.
+
+    Args:
+        text (str): The input text to be syllabized.
+        pre_syllabized (bool): Indicates whether the input text is already pre-syllabized.
+    Returns:
+        str: The syllabized text as a human-readable string.
+    """
+    syls_text = syllabize_text(text, pre_syllabized)
+    human_readable_text = " ".join(["".join(word) for word in syls_text])
+    return human_readable_text
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -22,7 +22,6 @@ class CenturyAdmin(BaseModelAdmin):
     pass
 
 
-@admin.register(Chant)
 class ChantAdmin(BaseModelAdmin):
     list_display = ("incipit", "siglum", "genre")
     search_fields = ("title", "incipit", "cantus_id")
@@ -86,6 +85,7 @@ class SourceAdmin(BaseModelAdmin):
 
 
 admin.site.register(Century, CenturyAdmin)
+admin.site.register(Chant, ChantAdmin)
 admin.site.register(Feast, FeastAdmin)
 admin.site.register(Genre, GenreAdmin)
 admin.site.register(Notation, NotationAdmin)
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -22,7 +22,11 @@ class CenturyAdmin(BaseModelAdmin):
     pass
 
 
+@admin.register(Chant)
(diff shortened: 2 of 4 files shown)


django/cantusdb_project/main_app/models/chant.py
```diff
@@ -148,3 +149,62 @@ def get_next_folio(folio):
...
+
+    def get_syllabized_text(self):
+        """Returns the syllabized text of a chant
+
+        Returns:
+            String: A String with words separated by spaces and syllables separated by "-"
+
+        """
+        if self.manuscript_syllabized_full_text:
+            print("hello")
```
dchiller: Probably not intended to be here still :)

django/cantusdb_project/main_app/models/chant.py
```diff
@@ -148,3 +149,62 @@ def get_next_folio(folio):
...
+            elif self.incipit:
+                syls_text = syllabize_text(self.incipit, pre_syllabized=False)
+                syls_text, syls_melody = postprocess(syls_text, syls_melody)
+            else:
+                syls_text = [[""]]
+
+            word_zip = align(syls_text, syls_melody)
+            return word_zip
+
+    def get_syllabized_text(self):
```
dchiller: There is a lot of repeat between this function and `get_syllabized_text_and_melody`. Perhaps these can be reworked in some way.

django/cantusdb_project/main_app/models/chant.py
```diff
@@ -148,3 +149,62 @@ def get_next_folio(folio):
             next_chant = None
 
         return next_chant
+
+    def get_syllabized_text_and_melody(self):
+        """Returns the syllabized text with the melody of a chant. If volpiano is not established return None
```
dchiller: Maybe say here what we mean by "syllabized text"? An short example?

dchiller:
@lucasmarchd01 

Just fyi, wondering if you know about linking Pull Requests and issues (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

django/cantusdb_project/main_app/models/chant.py
lucasmarchd01: Removed, oops!

django/cantusdb_project/main_app/models/chant.py
lucasmarchd01: Now joined into one function to avoid repetitive code

django/cantusdb_project/main_app/models/chant.py
lucasmarchd01: Added more description to the docstring

django/cantusdb_project/main_app/models/chant.py
```diff
@@ -148,3 +149,49 @@ def get_next_folio(folio):
             next_chant = None
 
         return next_chant
+
+    def get_syllabized_text_and_melody(self, text_only=False):
+        """Returns the syllabized text with the melody of a chant. Syllabized text splits the words from the
+         full text into its syllable components based on the volpiano protocols
+         (ex: Ocundare plebs fidelis cuius pater => I-O-cun-da-re plebs fi-de-lis cu-ius pa-ter)
```
dchiller: Oops, I think the `I-` at the beginning of your example is a typo?

dchiller (approved):

django/cantusdb_project/main_app/models/chant.py
annamorphism: the word actually is Iocundare, and *should* be syllabized Io-cun-da-re, though at the moment it automatically gets parsed as I-o-cun-da-re....maybe not the best example!

lucasmarchd01: I'll pick a different example, thanks!

django/cantusdb_project/main_app/admin.py
lucasmarchd01, line 1: Along the way, Andrew made these changes to admin.py that adds a search bar to the admin chant page

django/cantusdb_project/main_app/models/chant.py
```diff
@@ -1,5 +1,6 @@
 from django.db.models.query import QuerySet
 from main_app.models.base_chant import BaseChant
+from align_text_mel import *
```
dchiller: I think it would be better to be specific here -- are there lots of functions you are using?

django/cantusdb_project/align_text_mel.py
```diff
@@ -314,3 +314,40 @@ def align(syls_text, syls_melody):
...
+
+    This function takes a text, pre-syllabized text, and a melody as input. It performs syllabization on the text
+    and melody separately based on the volpiano protocols. The syllabized text and melody are then postprocessed,
+    and syllables of the text and melody are aligned to create a zip
+    of words and syllables.
+
+
+    Args:
+        text (str): The input text to be syllabized.
+        pre_syllabized (list): A list of pre-syllabized words corresponding to the input text.
```
dchiller: I think this should be boolean? It looks like you are passing booleans below.

dchiller (commented):
Refactor looks very nice!

django/cantusdb_project/main_app/admin.py
```diff
@@ -22,7 +22,11 @@ class CenturyAdmin(BaseModelAdmin):
     pass
 
 
+@admin.register(Chant)
```
jacobdgm: Is there a reason we're registering this area of the admin differently than the others?

lucasmarchd01: This was a change that Andrew made, I don't think there is any real difference so we can change it back if you think that's best.

jacobdgm (approved):
This is a great improvement over the code we previously had!

Approving, as all of this looks good, but leaving a couple of suggestions for possible changes if you want:
- I'm not sure about `get_most_descriptive_text` - what does it mean for text to be "descriptive"? Having had a weekend to think about it, maybe something along the lines of `get_best_text_for_syllabizing` would be clearest. At the very least, if I had questions about what "descriptive" means, I would look first at the docstring, where I find no clarification - maybe this would be a good spot for a small change explaining why we have this particular "chain of command".
- there are a couple of spots where you call `syllabize_text_to_string(some_text, some_boolean_variable)`. I would find these clearer if you used `syllabize_text_to_string(some_text, pre_syllabized=some_boolean_variable)`. Similarly in calls to `syllabize_text_and_melody` etc.

Again, neither of these are _necessary_ changes, but I'd find them helpful!

django/cantusdb_project/main_app/admin.py
jacobdgm: I just want things to be consistent, unless there's a good reason for the inconsistency!

PR #650, opened 2023-05-22
Allow option to edit from "view all chants"

### LINKED ISSUE #442: no option to edit from "view all chants"
image
http://206.12.93.196/chants/?source=674961

c.f. 
image
https://cantus.uwaterloo.ca/chants?source=123654

Note that on OldCantus an edit done from this link takes you back to "view all chants" when you are done with it, which I personally think is pretty handy.

lucasmarchd01:
This PR resolves #442 where there was no option to go to the edit page of a chant from the "view all chants" page of a source. This feature was available in OldCantus as an extra column on the table with an edit link. 

The implementation in NewCantus adds the same column to the table with the edit link
- the edit link is only shown if the use has editing permission
- after the user finishes editing the chant and clicks save, they will be taken back to the "view all chants" page

image

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1700,7 +1700,7 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
 
     def get_success_url(self):
         next_url = self.request.GET.get("ref")
-        if next_url:
+        if next_url == "chant-list":
             return self.request.POST.get("referrer")
         else:
             # stay on the same page after save
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1699,6 +1699,8 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
             return super().form_invalid(form)
 
     def get_success_url(self):
+        # Take user back to chant list page if that is the referrer page
+        # `ref` url parameter is used to indicate referring page
         next_url = self.request.GET.get("ref")
         if next_url == "chant-list":
             return self.request.POST.get("referrer")
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1699,8 +1699,12 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
             return super().form_invalid(form)
 
     def get_success_url(self):
-        # stay on the same page after save
-        return self.request.get_full_path()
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1696,8 +1699,12 @@ def form_valid(self, form):
             return super().form_invalid(form)
 
     def get_success_url(self):
-        # stay on the same page after save
-        return self.request.get_full_path()
+        next_url = self.request.GET.get("ref")
```
dchiller, line 1704: I don't think I follow the logic here...nothing wrong with the code, but can you explain?

dchiller (approved):

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: The `ref` parameter is included in the URL of the chant-list page to signify that the user has clicked the Edit link from that particular page. On the chant edit page, we utilize a hidden input called referrer to retrieve the value of `request.META.HTTP_REFERER`. This approach allows us to avoid hardcoding a URL as a parameter in the URL itself. However, upon reviewing the code, it does appear unclear and there seems to be a potential issue with the if statement if we decide to use the ref parameter elsewhere. I will make improvements to address these concerns.

jacobdgm: I second this! I think I understand what's going on - once the chant is saved, if the user landed on the chant edit page via another page (e.g. via a chant list page), we should send them back there; otherwise, we should stay on the same page - but I'm not certain. (did I get it right?)

Also, are we sure this is the intended behavior in all situations? If the user got to the Chant Edit page by clicking the "Edit" link on the Chant Detail page, do we want to drop them back on the Chant Detail page? (I think this is what happens with the code as it is; I think this behavior is fine, but it's different from what currently happens on OldCantus and NewCantus).

jacobdgm: ...actually, I just tested this out, and it redirects to a Chant List page but not to a Chant Detail page. And I don't understand why. So yes, comments/documentation would be very helpful here!

jacobdgm (approved):

jacobdgm (approved):

django/cantusdb_project/main_app/views/chant.py
annamorphism: Editing a chant from the Detail page does take you back to that Detail page after saving, I thought...

jacobdgm: huh! This makes sense (I think my understanding of how things work is shaped by my experience using the Edit-a-bunch-of-chants-in-a-single-source page exclusively during my time as an RA for CantusDB).

This should be an easy fix. Let me open an issue.

PR #661, opened 2023-05-25
Fix bug in CI Input Tool where table would be very narrow in Firefox

### LINKED ISSUE #639: Ci Input tool has narrow table
The table in the Input Tool pop-up is narrow, which looks silly and also means a lot of scrolling to find the relevant chant
image
cf Old Cantus: 
image

jacobdgm:
Fixes #639

The screenshots below are from Firefox; I've tested this in Safari and Chrome too. `test_views.py` runs as expected.

Before (screenshots from Firefox):
Screenshot 2023-05-25 at 9 18 32 AM

After:
Screenshot 2023-05-25 at 9 17 39 AM


lucasmarchd01:
As long as this doesn't change expected behaviour in other browsers, should be good to go.

lucasmarchd01 (approved):

PR #664, opened 2023-05-25
Fix CI input tool populate and popup bugs

### LINKED ISSUE #637: CI Input tool pop-up should go away
Currently after you choose a chant using the "Input tool" the window stays there, making it hard to tell if you selected the chant or not. In OldCantus it goes away after the information is in the edit form.

### LINKED ISSUE #638: CI input tool should also send text
After selecting a chant using the CI input tool, the edit form should get not only the selected Cantus ID, but also the full text from CantusIndex. Currently this is left blank, so you have to copy it from elsewhere.

lucasmarchd01:
This PR addresses a bug in the CI input tool on the create chant page. Previously, the window wouldn't close after using the tool, and the "Genre:" and "Text:" fields wouldn't populate correctly. The issue stemmed from a mismatch between genre names in the autoFill function. This PR introduces a solution by passing a "genres" list to the context data, containing the id and name of the Genre model. Now, the autoFill function can directly match the genre name with the corresponding id from the genres list.

Fixes #637, fixes #638

### DIFF
--- a/django/cantusdb_project/main_app/templates/ci_search.html
+++ b/django/cantusdb_project/main_app/templates/ci_search.html
@@ -42,8 +42,9 @@
             var cantus_id = document.getElementById(rowId).cells[1].innerText;
             var genre = document.getElementById(rowId).cells[2].innerText;
             var full_text = document.getElementById(rowId).cells[3].innerText;
-            var genres = {{ genres|safe }}; // genres contains "id" and "name" of Genre model
-            var genreId = (genres.find(item => item.name === genre) ?? {}).id;
+            var genres = {{ genres|safe }}; // genres contains "id" and "name" of all Genre objects in the database
+            var genreObj = genres.find(item => item.name === genre);
+            var genreId = genreId = genreObj ? genreObj.id : null;
 
             opener.document.getElementById('id_cantus_id').value = cantus_id;
             opener.document.getElementById('id_genre').value = genreId;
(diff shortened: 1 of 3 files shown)


lucasmarchd01:
Passed tests

django/cantusdb_project/main_app/templates/ci_search.html
```diff
@@ -42,12 +42,12 @@
             var cantus_id = document.getElementById(rowId).cells[1].innerText;
             var genre = document.getElementById(rowId).cells[2].innerText;
             var full_text = document.getElementById(rowId).cells[3].innerText;
-
-            // get the option value corresponding to the genre name
-            const genreId = Array.prototype.find.call(opener.document.getElementById('id_genre').children, ({ textContent }) => textContent === genre);
+            var genres = {{ genres|safe }}; // genres contains "id" and "name" of all Genre objects in the database
```
jacobdgm, line 47: the comment has disappeared!

django/cantusdb_project/main_app/templates/ci_search.html
```diff
@@ -42,12 +42,12 @@
...
+            var genreObj = genres.find(item => item.name === genre);
+            var genreId = genreId = genreObj ? genreObj.id : null;
```
jacobdgm, line 49: how many of these could be `const`s?

jacobdgm (approved):

django/cantusdb_project/main_app/templates/ci_search.html
lucasmarchd01: After testing this, I believe that they must remain as `var`s because if the user uses the chant input tool once it works fine but if they try and use it again and change the chant the variables will be locked to the previous chant's stored variables.

PR #665, opened 2023-05-25
Take user back to chant detail page after chant edit

### LINKED ISSUE #653: Chant Edit page should redirect to Chant Detail page upon submission of form under certain conditions
If the user got to the Chant Edit page by clicking on a link on the Chant Detail page, they should be redirected to the Chant Detail page when they press "Save".

lucasmarchd01:
This PR resolves #653, where the user was not redirected to the referring page after saving changes on the chant edit page. Building upon the solution implemented in PR #650, we use a ref parameter in the URL to indicate the referring page. This allows us to retrieve the referring page information from request.META.HTTP_REFERER and properly redirect the user after saving.

Instead of using a boolean flag, we utilize the ref parameter to differentiate between referring pages such as "chant-list" or "chant-detail". This approach offers flexibility for potential future developments that may require specific operations based on the referring page.

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -18,7 +18,7 @@
             
             {% if user_can_edit_chant %}
                 <p>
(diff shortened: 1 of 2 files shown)


lucasmarchd01:
Passed tests

jacobdgm (approved):

PR #667, opened 2023-05-26
Update ports in docker-compose.yml

jacobdgm:
This PR changes the ports in our `docker-compose.yml` file. This change is already present, as an uncommitted change, on staging and production (or at least, it is present when the apps are running and reachable). I don't know why the ports were set to `8000:80` in the first place, but testing this locally, upon running `docker-compose up -d`, I'm able to reach the app directly at `localhost`, rather than having to run `python manage.py runserver <a port that isn't 8000>`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
   nginx:
     image: nginx:alpine
     ports:
-      - 8000:80
+      - 80:80
     volumes:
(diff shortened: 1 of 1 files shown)


kqct (approved):
Looks good! Equally confused why we were exposing the nginx container on a nonstandard port.

dchiller:
Yay! excellent

Issue #467, opened 2022-12-08
Turn off debug on production server


jacobdgm:
Currently, if you navigate to an [erroneous page on production](http://206.12.88.113/source/12361012), you get a message saying "404" as well as a nice django debug page. We should ensure that this is turned off when we go to full production. This can probably happen as soon as we get logging set up (cf. #465).

Currently, we have the following code in `django/cantusdb_project/cantusdb/settings.py`:

```python
# SECURITY WARNING: don't run with debug turned on in production!
DEBUG = bool(strtobool(os.getenv("CANTUSDB_DEBUG", "False")))
```

I don't understand exactly what this is doing, but I gather there's an environment variable, `CANTUSDB_DEBUG`, whose value can be set in order to prevent the django debug page from being displayed.

lucasmarchd01:
I suppose we could just manually set this to False by default on the production server and when we develop locally or on staging we can set this to True?

jacobdgm:
While I don't know how to do this off the top of my head, I'm pretty sure it will not be difficult to do by changing the environment variable rather than by changing our code.

It's been helpful having error messages on "production" up until now, but now that we have the same version of Postgres running on both servers, now(-ish - it's not urgent) could be a good time to make this change.

jacobdgm:
This is complete! I've also updated the relevant wiki pages with instructions for turning DEBUG on on staging and on a local setup.

PR #671, opened 2023-05-30
Improve query optimization

lucasmarchd01:
This PR is a step in the process of improving our query optimization for slow-loading pages. Previously, certain pages such as source-detail and chant-list would take a significant amount of time to load due to a large number of similar or duplicate SQL queries. The solution mainly involves the use of django's `select_related` and `prefetch_related` methods that allow for improvements in the number of database hits. More specifically:
- add debug toolbar settings for development
- refactor `feast_selector_options` method to remove unnecessary nested loop
- use `select_related` and `prefetch_related` mothods to retrieve relevant data and optimize database queries by reduce the number of database hits when accessing related objects.

Here is an example of an improved query optimization on the source-detail page.

Before:
Screenshot 2023-05-30 at 11 05 06 AM copy

After:
Screenshot 2023-05-30 at 11 05 53 AM copy

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -61,11 +61,9 @@ INSTALLED_APPS = [
     "articles",
     "django_quill",  # to provide rich-text field for articles
     "users",
-    "debug_toolbar",
 ]
 
 MIDDLEWARE = [
-    "debug_toolbar.middleware.DebugToolbarMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
@@ -171,3 +169,7 @@ EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 INTERNAL_IPS = [
     "127.0.0.1",
 ]
+
+if DEBUG:
+    INSTALLED_APPS.append("debug_toolbar")
+    MIDDLEWARE.append("debug_toolbar.middleware.DebugToolbarMiddleware")
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -62,12 +62,12 @@ INSTALLED_APPS = [
     "django_quill",  # to provide rich-text field for articles
     "users",
     "debug_toolbar",
-    "silk",
+    # "silk",
 ]
 
 MIDDLEWARE = [
     "debug_toolbar.middleware.DebugToolbarMiddleware",
-    "silk.middleware.SilkyMiddleware",
+    # "silk.middleware.SilkyMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
@@ -174,3 +174,7 @@ if DEBUG:
 
     hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
     INTERNAL_IPS = [ip[: ip.rfind(".")] + ".1" for ip in ips] + ["127.0.0.1"]
+
+SILKY_MAX_REQUEST_BODY_SIZE = -1  # Silk takes anything <0 as no limit
+SILKY_MAX_RESPONSE_BODY_SIZE = 1024  # If response body>1024 bytes, ignore
+SILKY_META = True
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -62,12 +62,10 @@ INSTALLED_APPS = [
     "django_quill",  # to provide rich-text field for articles
     "users",
     "debug_toolbar",
-    # "silk",
 ]
 
 MIDDLEWARE = [
     "debug_toolbar.middleware.DebugToolbarMiddleware",
-    # "silk.middleware.SilkyMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
@@ -170,11 +168,7 @@ DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 
 if DEBUG:
-    import socket  # only if you haven't already imported this
+    import socket
 
     hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
     INTERNAL_IPS = [ip[: ip.rfind(".")] + ".1" for ip in ips] + ["127.0.0.1"]
-
-SILKY_MAX_REQUEST_BODY_SIZE = -1  # Silk takes anything <0 as no limit
-SILKY_MAX_RESPONSE_BODY_SIZE = 1024  # If response body>1024 bytes, ignore
-SILKY_META = True
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -167,8 +167,7 @@ DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 
-if DEBUG:
-    import socket
 
-    hostname, _, ips = socket.gethostbyname_ex(socket.gethostname())
-    INTERNAL_IPS = [ip[: ip.rfind(".")] + ".1" for ip in ips] + ["127.0.0.1"]
+INTERNAL_IPS = [
+    "127.0.0.1",
+]
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -62,10 +62,12 @@ INSTALLED_APPS = [
     "django_quill",  # to provide rich-text field for articles
     "users",
     "debug_toolbar",
+    "silk",
 ]
 
 MIDDLEWARE = [
     "debug_toolbar.middleware.DebugToolbarMiddleware",
+    "silk.middleware.SilkyMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -172,4 +172,5 @@ INTERNAL_IPS = [
 
 if DEBUG:
     INSTALLED_APPS.append("debug_toolbar")
+    # debug toolbar must be inserted as early in the middleware as possible
     MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -61,6 +61,7 @@ INSTALLED_APPS = [
     "articles",
     "django_quill",  # to provide rich-text field for articles
     "users",
+    "debug_toolbar",
 ]
 
 MIDDLEWARE = [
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/cantusdb/settings.py
```diff
@@ -61,9 +61,11 @@
     "articles",
     "django_quill",  # to provide rich-text field for articles
     "users",
+    "debug_toolbar",
```
dchiller: Was there someway this was secure if running live? I feel like there was but I can't remember....

django/cantusdb_project/cantusdb/settings.py
```diff
@@ -164,3 +166,9 @@
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+
+if DEBUG:
```
dchiller: What is this doing?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -5,7 +5,7 @@
 import threading
 from django.contrib import messages
 from django.contrib.postgres.search import SearchQuery, SearchRank
-from django.db.models import F, Q, QuerySet, Value
+from django.db.models import F, Q, QuerySet, Value, Prefetch
```
dchiller: Do you use `Prefetch` in this file?

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1463,7 +1462,7 @@ def get_feast_selector_options(source, folios):
                 folios (list of strs): A list of folios in the source.
 
             Returns:
-                zip object: A zip object combining a list of folios and Feast objects, to be unpacked in template.
+                list of tuples: A list of folios and Feast objects, to be unpacked in template.
```
dchiller: I know this is not you, but I do not like these repeated functions!

django/cantusdb_project/main_app/views/source.py
```diff
@@ -1,6 +1,6 @@
 from django.views.generic import DetailView, ListView, CreateView, UpdateView
-from django.db.models import Q
-from main_app.models import Source, Provenance, Century
+from django.db.models import Q, Prefetch
+from main_app.models import Source, Provenance, Century, Feast
```
dchiller: Do you use `Feast` in this file?

django/cantusdb_project/main_app/views/source.py
```diff
@@ -42,7 +42,7 @@ def get_feast_selector_options(source, folios):
             # initialize the feast selector options with the first chant in the source that has a feast
             first_feast_chant = chants_in_source.first()
             if not first_feast_chant:
-                # if none of the chants in this source has a feast, return an empty list
+                # if none of the chants in this source has a feast, return an empty zip
```
dchiller: I think it is a list....

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -516,7 +515,7 @@ def get_queryset(self):
         search_text = self.request.GET.get("search_text")
 
         # get all chants in the specified source
-        chants = source.chant_set
+        chants = source.chant_set.select_related("feast", "office", "genre")
```
dchiller, line 569: I'm confused about what this is doing...where is this additional query bit used?

dchiller (commented):
I think what you are doing makes sense for optimizing the current approach. 

I think there's also a different approach -- I'm not sure that it offers much improvement in terms of load time, so maybe it is out of scope, but I think it would greatly simplify the code and the multiple nested functions. 

As I understand it, you are basically trying to come up with a list of (`folio`, `feast`) pairs....For a given source, what are all the combinations of folios and feasts (represented by chants for those feasts) that exist in the source. 

That sounds to me like a use case for a single SQL query (pseudo-coding it here):
`SELECT DISTINCT chant_tbl.folio, feast_tbl.feast_name FROM chant_tbl JOIN feast_tbl ON chant_tbl.feast = feast_tbl.id WHERE source = source_id`
I don't quite know how one would do this in django, but I wouldn't be surprised if the queryset API could do it...

django/cantusdb_project/cantusdb/settings.py
lucasmarchd01: I followed the instructions from https://django-debug-toolbar.readthedocs.io/en/latest/installation.html where it says this code is needed to configure INTERNAL_IPS correctly with Docker. Not sure exactly what's going on though. I just tested it without this and it still works so I'll probably end up removing it.

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: I was using it in a previous commit but I since changed it to select_related. Somehow those changes are not being reflected in the "files changed" tab though.

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: Yes I don't like them either... There's definitely a way this function can be refactored so that it's not nested within another function and also not repeated throughout the code.

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: These were necessary to include because many of the queries were caused in the chant_list.html template, specifically in:

```
</td>
<td class="text-wrap" style="text-align:center">
    {% if chant.feast %}
        <a href="{% url 'feast-detail' chant.feast.id %}" title="{{ chant.feast.description }}">{{ chant.feast.name|default:"" }}</a>
    {% endif %}
</td>
<td class="text-wrap" style="text-align:center">
    {% if chant.office %}
        <a href="{% url 'office-detail' chant.office.id %}" title="{{ chant.office.description }}">{{ chant.office.name|default:"" }}</a>
    {% endif %}
</td>
<td class="text-wrap" style="text-align:center">
    {% if chant.genre %}
        <a href="{% url 'genre-detail' chant.genre.id %}" title="{{ chant.genre.description }}">{{ chant.genre.name|default:"" }}</a>
    {% endif %}
</td>
```

django/cantusdb_project/cantusdb/settings.py
lucasmarchd01: In settings.py we have `DEBUG = bool(strtobool(os.getenv("CANTUSDB_DEBUG", "False")))` which I assume would assure it doesn't run live, but I'm not entirely sure where these environment variables are stored

ahankinson:
```py
if DEBUG:
    INSTALLED_APPS.append("debug_toolbar")
```

django/cantusdb_project/cantusdb/settings.py
```diff
@@ -61,9 +61,11 @@
     "articles",
     "django_quill",  # to provide rich-text field for articles
     "users",
+    "debug_toolbar",
 ]
 
 MIDDLEWARE = [
+    "debug_toolbar.middleware.DebugToolbarMiddleware",
```
ahankinson: Same here

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -202,6 +201,58 @@ def make_suggested_chant_dict(
...
+    feast_selector_folios = []
+    # get all chants in the source, select those that have a feast
+    chants_in_source = (
+        source.chant_set.exclude(feast=None)
+        .order_by("folio", "c_sequence")
+        .select_related("feast")
+    )
+    # initialize the feast selector options with the first chant in the source that has a feast
+    first_feast_chant = chants_in_source.first()
+    if not first_feast_chant:
```
ahankinson, line 228:
```py
if not first_feast_chant:
    return []
```

Then you don't need the `else` clause, and can get rid of one level of nesting.

django/cantusdb_project/cantusdb/settings.py
lucasmarchd01: Mentioning issue #467 here as I do not think this has been resolved in the production server yet.

django/cantusdb_project/cantusdb/settings.py
dchiller: Makes sense!

django/cantusdb_project/main_app/views/chant.py
dchiller: yeah, I've had that happen sometimes too...cool!

django/cantusdb_project/main_app/views/chant.py
dchiller: Gotcha! Nice!

django/cantusdb_project/cantusdb/settings.py
jacobdgm: Fwiw my solution has been to have a separate branch - https://github.com/jacobdgm/CantusDB/tree/toolbar - and whenever I want to test something with DjDT, I run `git checkout toolbar cantusdb/settings.py` and `git checkout toolbar cantusdb/urls.py`.

I don't think we want this turned on all the time on the staging server, where `DEBUG` should always be set to `True` I think - while it's good for testing page load times, it makes things run really slowly. But I guess it's possible to turn off all the features we don't want in the browser.

All this is to say, I have this solution that you can use if you want, but there's quite possibly a better approach, so we should discuss this tomorrow.

lucasmarchd01: Ah okay, sounds good!

dchiller: I'm not sure it's best to run your staging server with DEBUG = True?

jacobdgm: I just closed #467; DEBUG is currently set to True on staging and False on Production.

I'm not sure whether to worry about having DEBUG set to True on staging. For the time being, at least, it could be quite useful to be able to check in on our query times on the ComputeCanada server (in some situations, load times are slower on CC than locally, so having the ability to test this on CC will be helpful).

A possible solution: we could start up another (limited-access) "development" server for debugging. I guess this would use up one more of our allotted IP addresses...

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -100,8 +100,7 @@ def parse_json_from_api(url: str) -> Union[list, None]:
             print(  # eventually, we should log this rather than printing it to the console
                 "Encountered an error in",
                 "ChantCreateView.get_suggested_chants.make_suggested_chant_dict",
-                "while parsing the response from",
-                f"https://cantusindex.org/json-cid/{cantus_id}:",
+                f"while parsing the response from {url}",
```
jacobdgm, line 103: nice catch

django/cantusdb_project/main_app/views/source.py
```diff
@@ -220,7 +168,13 @@ def get_queryset(self):
             )
             q_obj_filter &= indexing_search_q
 
-        return queryset.filter(q_obj_filter).distinct()
+        return (
+            queryset.filter(q_obj_filter)
+            .distinct()
```
jacobdgm: do we need `.distinct()` here? Why was this here in the first place - we don't/can't have any duplicate sources, do/can we? `.distinct`/`DISTINCT` can be a fairly heavy operation iirc, but maybe with <1000 sources it's not that bad.

"this would have been a great place for a comment," Jacob yells into the past...

jacobdgm (approved):
This looks good! Let's look into the one place with `.distinct` before merging, though, and either remove it or add a comment explaining why it's there...

django/cantusdb_project/cantusdb/settings.py
dchiller: Yeah, I don't think that it's an emergency in any case...just that probably the same concerns about running it in production apply to staging if those are substantially similar, which I imagine they are. 

For the longer term, something you might want to do is have the staging server no accessible to the public (eg. Cantus Ultimus staging is only accessible through the a VPN) ~~ again, I know there are bigger fish to fry right now, but maybe Katie can bring NewCantus staging into agreement with the rest of DDMAL's networking.

jacobdgm: @kqct?

django/cantusdb_project/main_app/views/source.py
lucasmarchd01: Hmm.. I tried removing distinct() and it didn't change the number of results so I'm not sure it's doing anything. It also reduces the runtime for loading the page so it might be best to remove this altogether.

PR #672, opened 2023-05-31
Reorder chant edit fields

### LINKED ISSUE #517: edit chant--fields out of order
On OldCantus all the fields on a chant's edit page are in the same order as they are when viewing the chant. 
image

Not so in NewCantus, where the text is above the fields with the folio, sequence, etc. 

image

lucasmarchd01:
This PR improves organization of fields in the chant edit page. The current order of fields is adjusted to match the order displayed in the chant detail page and aligns with the behaviour of oldCantus.

image

Resolves #517

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_create.html
+++ b/django/cantusdb_project/main_app/templates/chant_create.html
@@ -152,7 +152,7 @@
                         {{ form.melody_id }}
                     </div>
 
-                    <div class="form-row align-items-top"> <!-- the Addendum field is no longer there in OldCantus... -->
+                    <div class="form-row align-items-top">
                         <div class="form-group m-1 col-lg">
                             <small>{{ form.addendum.label_tag }}</small>
                             {{ form.addendum }}
--- a/django/cantusdb_project/main_app/templates/chant_create.html
+++ b/django/cantusdb_project/main_app/templates/chant_create.html
@@ -152,6 +152,13 @@
                         {{ form.melody_id }}
                     </div>
 
+                    <div class="form-row align-items-top"> <!-- the Addendum field is no longer there in OldCantus... -->
+                        <div class="form-group m-1 col-lg">
+                            <small>{{ form.addendum.label_tag }}</small>
+                            {{ form.addendum }}
+                        </div>
+                    </div>
+
                 </div>
             
                 <div class="form-row align-items-end">
@@ -208,13 +215,6 @@
                     <div class="form-group m-1 col-lg">
                         <small>{{ form.indexing_notes.label_tag }}</small>
                         {{ form.indexing_notes }}
-                    </div>
-                </div>
-
-                <div class="form-row align-items-end"> <!-- the Addendum field is no longer there in OldCantus... -->
-                    <div class="form-group m-1 col-lg">
-                        <small>{{ form.addendum.label_tag }}</small>
-                        {{ form.addendum }}
                     </div>
                 </div>
 
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -277,6 +277,7 @@ class ChantEditForm(forms.ModelForm):
             "extra",
             "image_link",
             "indexing_notes",
+            "addendum",
         ]
         widgets = {
             # manuscript_full_text_std_spelling: defined below (required)
@@ -298,6 +299,7 @@ class ChantEditForm(forms.ModelForm):
             "extra": TextInputWidget(),
             "image_link": TextInputWidget(),
             "indexing_notes": TextAreaWidget(),
+            "addendum": TextInputWidget(),
         }
 
     manuscript_full_text_std_spelling = forms.CharField(
--- a/django/cantusdb_project/main_app/templates/chant_edit.html
+++ b/django/cantusdb_project/main_app/templates/chant_edit.html
@@ -112,6 +112,14 @@
                             {{ form.extra }}
                         </div>
                     </div>
+                    <div class="form-row">
+                        <div class="form-row align-items-top"> 
+                            <div class="form-group m-1 col-lg-8">
+                                <small>{{ form.addendum.label_tag }}</small>
+                                {{ form.addendum }}
+                            </div>
+                        </div>
+                    </div>
                                         
                     <div class="form-row">
                         <div class="form-group m-1 col-lg-12">
--- a/django/cantusdb_project/main_app/templates/chant_edit.html
+++ b/django/cantusdb_project/main_app/templates/chant_edit.html
@@ -39,77 +39,7 @@
                 {% endif %}
                 <form method="post" style="line-height: normal">{% csrf_token %}
                     <input type="hidden" name="referrer" value="{{ request.META.HTTP_REFERER }}">
-                    <div class="form-row">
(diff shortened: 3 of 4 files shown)


jacobdgm:
I feel like we should make the same change on the proofread page, as well as the chant-create page. @annamorphism, does this sound right?

annamorphism:
Agree--all these fields should be consistently displayed throughout!

lucasmarchd01:
@jacobdgm @annamorphism Just noting some other differences between proofread, edit, and create. Can you let me know which fields should stay the same or added to others?

- proofread and create has a field for addendum, should this be available in chant edit as well? Also, the addendum field is located near the top of the page for proofread and at the bottom for create, so what is the preferred location?
- proofread and create have a chant range field that is not available in chant edit

Proofread page with addendum and chant range field:
image

- proofread has a siglum field not available in chant edit or chant create
- proofread has a macuscript syllabized full text field, should this also be available in chant edit? This also makes me wonder what the purpose of the "edit syllabification" link is for if we can just display and edit this field directly in the chant edit or proofread page without the need to open a new window.

Proofread page with siglum and syllabification field:
image

Otherwise, after this PR change the general ordering of the fields is the same except for these discrepancies between fields that are/aren't available in others.

annamorphism:
> @jacobdgm @annamorphism Just noting some other differences between proofread, edit, and create. Can you let me know which fields should stay the same or added to others?
> 
>     * proofread and create has a field for addendum, should this be available in chant edit as well? Also, the addendum field is located near the top of the page for proofread and at the bottom for create, so what is the preferred location?
>
In OldCantus this is near the top, so let's keep it there. As it's largely used for (I think) older chants in the Database, it's probably most useful in chant-edit, actually.

>     * proofread and create have a chant range field that is not available in chant edit
>
This is automatically generated after filling in the Volpiano, I thought (though I'm now struggling to get it to work...). If so maybe it's best as a non-editable field? 

>     * proofread has a siglum field not available in chant edit or chant create
I don't think this exists at all in oldCantus--will look around to see what you mean, but I think it's probably not necessary. 
> 
>     * proofread has a macuscript syllabized full text field, should this also be available in chant edit? This also makes me wonder what the purpose of the "edit syllabification" link is for if we can just display and edit this field directly in the chant edit or proofread page without the need to open a new window.
> 
Ideally you don't need to edit the syllabification very often, so in most cases this field is clutter when you already have two text fields, a volpiano field, a matched-volpiano-to-text display...That said if there is a way to streamline some of these actions that might be nice. 
> 
> Otherwise, after this PR change the general ordering of the fields is the same except for these discrepancies between fields that are/aren't available in others.

jacobdgm:
> > proofread and create have a chant range field that is not available in chant edit
> 
> This is automatically generated after filling in the Volpiano, I thought (though I'm now struggling to get it to work...). If so maybe it's best as a non-editable field?

If the `chant_range` field is meant to be auto-generated, this is a bit of functionality that we haven't yet implemented. But if it doesn't work in OldCantus, then the best approach is probably to shelve this until someone asks for it (i.e. keep the field where it is, and don't try to automate anything).

> >  proofread has a siglum field not available in chant edit or chant create
> 
> I don't think this exists at all in oldCantus--will look around to see what you mean, but I think it's probably not necessary.

I'm confused about this one. I know that all the chants in the database have a `siglum` field, and that it almost always duplicates its source's `siglum`. Maybe OldCantus added this field to their chant model in order to optimize something. In which case, we should hold onto this field just in case, but not ever display it. More directly, for the purposes of this PR, we should duplicate whatever OldCantus does on its proofread page.

> > proofread has a macuscript syllabized full text field, should this also be available in chant edit? This also makes me wonder what the purpose of the "edit syllabification" link is for if we can just display and edit this field directly in the chant edit or proofread page without the need to open a new window.
> 
> Ideally you don't need to edit the syllabification very often, so in most cases this field is clutter when you already have two text fields, a volpiano field, a matched-volpiano-to-text display...That said if there is a way to streamline some of these actions that might be nice.

Let's duplicate whatever's in OldCantus's proofread page. In all situations, let's do this.

annamorphism:
by "OldCantus's proofread page" do you mean the edit-volpiano page? I don't think it has "proofread" as such.

lucasmarchd01:
Updated proofread, edit, and create chant pages.
- Proofread page has syllabized text and siglum fields removed
- edit page has addendum added
- create page has addendum moved towards the top of the page

Proofread chant
image

Edit chant
image

Create chant
image

jacobdgm:
> by "OldCantus's proofread page" do you mean the edit-volpiano page? I don't think it has "proofread" as such.

https://cantus.uwaterloo.ca/proofread/699578

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -152,6 +152,13 @@ <h3>Create Chant</h3>
                         {{ form.melody_id }}
                     </div>
 
+                    <div class="form-row align-items-top"> <!-- the Addendum field is no longer there in OldCantus... -->
```
jacobdgm: if it's not in OldCantus any more, let's just get rid of it.

jacobdgm (approved):
Approving, but we should remove the Addendum field from the ChantCreate page before merging.

annamorphism:
> > by "OldCantus's proofread page" do you mean the edit-volpiano page? I don't think it has "proofread" as such.
> 
> https://cantus.uwaterloo.ca/proofread/699578

o.O I have no idea how to actually navigate to this in OldCantus! From my-sources one can get to "proofread2", but it promptly denies me access; if I manually delete the 2, I can proofread, but submitting gets me to an ajax gobbledygook page. So I don't think the "proofread" page is necessarily the authority to replicate, since it seems to have been deprecated...

django/cantusdb_project/main_app/templates/chant_create.html
annamorphism: It is there in OldCantus when you create chants....image
It's also in edit 
image
but not in edit-volpiano 
image

jacobdgm:
> > https://cantus.uwaterloo.ca/proofread/699578
> 
> o.O I have no idea how to actually navigate to this in OldCantus! From my-sources one can get to "proofread2", but it promptly denies me access; if I manually delete the 2, I can proofread, but submitting gets me to an ajax gobbledygook page. So I don't think the "proofread" page is necessarily the authority to replicate, since it seems to have been deprecated...

Is this one to ask Debra about? She's requested features for the proofread page before, so I think she (at least - and maybe some others) makes use of the proofread page. If this is so, I'm confused as to why it's evidently broken on OldCantus.

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: we should remove the comment, then!

jacobdgm: I realize that we have another discrepancy between OldCantus and NewCantus - OC's "edit-volpiano" is NC's "edit-chants": http://206.12.88.113/edit-chants/699578

I think I may have suggested/made this change last summer. I think "edit-volpiano" doesn't describe what happens on the page nearly as well as "edit-chants", but if we want perfect backwards compatibility, we should change this back.

PR #673, opened 2023-05-31
Reorder chants by folio in "Browse Chants" page

### LINKED ISSUE #524: chants on chant search page out of order
yes, it's a source to which chants have been added haphazardly, however it's not clear as to why 004r should be before 001r on http://206.12.88.113/chants/?source=699578 ...

Screen Shot 2023-01-13 at 10 51 43 AM

lucasmarchd01:
This PR addresses the issue of incorrect ordering of chants on the "Browse Chants" page for a source. Currently, the chants are ordered based on the time they were added, instead of their folio. As a result, if chants were added out of order, they would be displayed in the wrong sequence. The solution is simply reordering the queryset returned in the ChantListView to be ordered by "folio" rather than "id".

As for the other comments mentioned in #524 regarding the missing siglum, I could not replicate this behaviour. When I add a chant the siglum corresponds to the source that it has been added in. Perhaps this has already been resolved.

Fixes #524

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -530,7 +530,7 @@ class ChantListView(ListView):
                 | Q(incipit__icontains=search_text)
                 | Q(manuscript_full_text__icontains=search_text)
             )
-        return chants.order_by("folio")
+        return chants.order_by("folio", "c_sequence")
 
     def get_context_data(self, **kwargs):
         def get_feast_selector_options(source, folios):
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -530,7 +530,7 @@ def get_queryset(self):
                 | Q(incipit__icontains=search_text)
                 | Q(manuscript_full_text__icontains=search_text)
             )
-        return chants.order_by("id")
+        return chants.order_by("folio")
```
dchiller: Do you want to also order by sequence?

jacobdgm: This would be a good idea. I think it's just a second argument. It'll be `c_sequence` (because we're not dealing with any Sequence objects, correct?)

jacobdgm (approved):
Looks good. I like Dylan's idea of ordering by chants' `c_sequence` as well.

lucasmarchd01:
Great, makes sense. Now it's ordered by both folio and c_sequence

PR #691, opened 2023-06-05
rework Dockerfile and add additional gunicorn workers

jacobdgm:
This PR reorders the commands in our Dockerfile to make building faster, and increases the number of gunicorn workers.


kqct (approved):
Looks good!

Issue #699, opened 2023-06-06
Genre dropdown list formatting is strange


lucasmarchd01:
In the chant create page, the genre dropdown list has strange behaviour for some of the genres. We should reformat this dropdown menu to display the shorthand notation only without the entire string representation.

![image](https://github.com/DDMAL/CantusDB/assets/71031342/e6b70180-f8f2-4493-b140-9e0b2ef5647f)

annamorphism:
Just noting that OldCantus has no brackets here except for those items given in double brackets in NewCantus (for categories that are discouraged for new entries.). Hopefully this is taken into account in the shorthand notation on NewCantus.

jacobdgm:
@lucasmarchd01 was this fixed by #703?

edit: it seems like it should have been; not sure why GitHub didn't handle this on its own. Closing.

PR #703, opened 2023-06-06
Improved Dropdown Menu Display for Genre and Office

### LINKED ISSUE #512: Internal string representations of objects shown in Office, Genre dropdowns on chant-edit, source-edit-chants pages
found on http://206.12.88.113/edit-chants/123610?pk=243158&folio=001v

Screen Shot 2023-01-11 at 13 18 15

lucasmarchd01:
This PR resolves the problem with the dropdown menu in the ChantCreate and ChantEdit views. Previously, the menu displayed the complete string description of the genre and office objects, which was not desired. To address this, the template forms were modified to use the "name" attribute of the objects in the dropdown menu. This was achieved by introducing a custom `NameModelChoiceField` that overrides the `label_from_instance` method to specifically display the `name` attribute of the genre and office objects. The modification applies to the following pages:
- Chant create
- Chant edit
- Chant proofread
- Sequence edit

Fixes #512, #699

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -45,6 +45,17 @@ class CommentForm(forms.ModelForm):
 """
 
 
+class NameModelChoiceField(forms.ModelChoiceField):
+    """
+    A custom ModelChoiceField that overrides the label_from_instance method
+    to display the object's name attribute instead of str(object).
+    This field is specifically designed for handling genre and office objects.
+    """
+
+    def label_from_instance(self, obj):
+        return obj.name
+
+
 # 3 best
 class ChantCreateForm(forms.ModelForm):
     class Meta:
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/forms.py
```diff
@@ -45,6 +45,17 @@ def __init__(self, *args, **kwargs):
 """
 
 
+class NameModelChoiceField(forms.ModelChoiceField):
+    """
+    A custom ModelChoiceField that overrides the label_from_instance method
+    to display the object's name attribute instead of str(object).
+    This field is specifically designed for handling genre and office objects.
```
jacobdgm, line 52: Open to other opinions on this, but I'd really like a concrete, specific explanation of why this was implemented, either in this docstring or in a comment nearby. I.e. "we chose to create a custom class so that when we have dropdown `select` elements on our forms where users can select Office and Genre objects, we want to display the 1-2 letter initialisms for these objects instead of their string representations, e.g. 'M' rather than '[M] Matins'" (but edited for better clarity and conciseness if possible)

jacobdgm: this is true in other situations where we're adding "why" sorts of comments - I'm of the opinion that these should be very concrete rather than abstract.

In this specific case, future developers can see that we overwrote `label_from_instance` by looking down 5 lines - the questions I'm anticipating are "Why would we want to overwrite label_from_instance?" and "Why would we want to display an object's name, and why would we want to do this for genre and office objects in specific?"

jacobdgm (approved):
I'm suggesting tweaks to the docstring/comments, but on the whole, this looks good!

PR #716, opened 2023-06-08
Redirect /node/ urls to their corresponding detail page in NewCantus

### LINKED ISSUE #563: Set up redirects for /node/ URLs
UPDATE: We plan to solve this problem thus:

When we sync data with OldCantus for the final time, we make note of the largest ID in the database that works with a `/node/` URL. When someone tries to visit a `/node/` URL with an ID less than or equal to the noted ID, we search for the right detail page and redirect them to it. When someone tries to visit a `/node/` URL with an ID greater than the noted ID, we give them a 404 error.

In order to prevent Postgres from adding new chants/sequences/sources/etc. with IDs smaller than the noted ID, which would conflict with valid `/node/` URLs already in the database, we create a dummy chant, a dummy sequence, a dummy source, etc., each with an ID one greater than the noted ID. This will ensure that all new objects created in those tables are assigned IDs greater than the noted ID.

Between now and the final data sync, we should implement redirection for `/node/` URLs, choosing a large placeholder "noted ID". We can then update this noted ID once we've created our placeholder objects.

## Summary

Many pages on OldCantus can be accessed via a url with `/node/` in the path. NewCantus doesn't currently support these URLs. Since people may have bookmarked/cited these `/node/<id>` URLs, we need to set up a way to handle requests to these URLs.

OldCantus guarantees that all objects, no matter their type, have unique IDs, so any `/node/` URL will point to a single object in the database. Django doesn't provide this guarantee though, so multiple objects of different kinds may have conflicting IDs.

In order to ensure backwards compatibility, we need to make sure the expected information is displayed when a person visits a `/node/` URL. There are a few ways we could accomplish this:
- Prevent Django from saving objects with conflicting IDs
- Display a disambiguation page for `/node/` URLs, providing links to the detail pages of all objects the ID could represent.
- For each valid `/node/` URL on OldCantus, keep a record of the object it points to and automatically redirect to the proper page when a user enters that URL.
- Keep a list of IDs that can create valid `/node/` URLs, and ensure all  new objects are assigned IDs that aren't on that list (or even keep track of the largest such ID, and make sure new IDs are larger than that). For IDs that currently have valid `/node/` IDs, we can redirect to the proper detail page.

## Further detail

On OldCantus, many pages are accessible via two different URLs, one with `/node/` in the path and one with the object name in the path. For example, both https://cantus.uwaterloo.ca/source/123931 and https://cantus.uwaterloo.ca/node/123931 link to identical pages about the source with siglum `A-Gu 1584`. We don't currently have anything set up to handle urls with `/node/` in their path. People may have cited `/node/` paths, so it's important that we have a system in place to handle these URLs. These paths can point to [indexers](https://cantus.uwaterloo.ca/node/706860), [chants](https://cantus.uwaterloo.ca/node/243652), [sequences](https://cantus.uwaterloo.ca/node/628334), [sources](https://cantus.uwaterloo.ca/node/123658) and [articles](https://cantus.uwaterloo.ca/node/712597) but not to users, provenances, centuries, notations, feasts, genres, or offices.

My understanding is that OldCantus's Drupal setup stores all objects in a single table, so each is guaranteed to have a unique primary key. Each row in the table has a column indicating whether it represents a source, sequence, chant, indexer, etc., which allows Drupal to display the right kind of detail page for a given primary key.

A naive approach to handling `/node/<pk>` urls would be to check each of our models in turn to see whether there exists (say) a Chant with the given pk, a Source with the given pk, a Sequence, an Indexer, and so on - as soon as we find one, we redirect the user to the correct detail page. (This is how our [json-node API](https://github.com/DDMAL/CantusDB/wiki/APIs#json_node_export) currently works; see related issue #564)

On NewCantus's Django, however, each of our different models has its own table: all Chants are stored in a `chants` table, all Sequences in a `sequences` table, and so on. Django, thus, does not provide any guarantee that all objects will have unique primary keys. (see [this comment](#issuecomment-1438998934) (or just scroll down) to see how I managed to create a situation where objects of different types have conflicting IDs). The difficulty, then, arises when people begin adding new chants/sources directly to NewCantus.

Consider this situation: Let `largest_id_source` be Source in OldCantus with the largest ID. Any objects added to OldCantus will have IDs that follow it sequentially - let's say that the indexer that created `largest_id_source` immediately began adding chants to it, such that one of these chants has an ID equal to `largest_id_source.id + 1`. When we import that data into NewCantus, those objects' IDs are conserved. But when someone adds a new Source to NewCantus, Django will assign it the next ID that doesn't already belong to a source, i.e. `largest_id_source.id + 1`. Now, when a user navigates to `/node/<largest_id_source.id + 1>`, we won't know whether to redirect them to `/chant/<largest_id_source.id + 1>` or to `/source/<largest_id_source.id + 1>`, both of which are valid pages in NewCantus.

#### Prevent ID conflicts
One possible solution is to set up a validation step when saving new objects to prevent ID collisions from occurring. This way, we can guarantee that every ID corresponds to a single object, so there's no ambiguity when trying to redirect to the proper detail page.

Benefits:
- From user's point of view, works very similarly to OldCantus. Anyone visiting a `/node/` address can be sent directly to the proper detail page.

Drawbacks:
- This is not how Django is set up to work, and my understanding is that very few people use Django in this way - looking on StackOverflow, answers for "how do I ensure unique IDs among tables?" range from "why?" to "don't" to "have you thought of using UUIDs?". There doesn't appear to be an established best practice, so if something goes wrong with our setup, there will be fewer resources (documentation, answers on StackOverflow, etc.) to help us fix the problem.

#### Disambiguation Page
Another possible solution is to allow Django/Postgres to assign new IDs using their default behavior. When a user tries to navigate to a `/node/` page, we search for all objects that match the ID, and display a [wikipedia-style disambiguation page](https://en.wikipedia.org/wiki/Joker) explaining that CantusDB no longer uses `/node/` URLs, and providing a list of links to the detail pages of objects with matching IDs.

Benefits:
- We're not fighting against Django while developing.
- Behavior is transparent and consistent.

Drawbacks:
- Users need to click through to visit the page they were looking for, and may need to choose from among a list of options

#### Dictionary of valid IDs
Another possible solution is to keep the default behavior for assigning new IDs, but keep a separate dictionary of objects that have IDs that work with `/node/` URLs, i.e. objects added before the switch from OldCantus to New (we'd create this dictionary around the moment when we do the switch). Any time a user visits a `/node/` URL, we look through the keys in our dictionary and redirect the user to the URL stored there.

Benefits:
- We're not fighting against Django
- Behavior very similar to OldCantus - when user visits a `/node/` URL, they are silently redirected to the correct detail page

Drawbacks:
- Behavior is inconsistent - if you take the ID of a given chant (say) and append it to `/node/` in a URL, it will sometimes work (if the chant was added to OldCantus) but sometimes generate an error (if the chant was added to NewCantus).

#### Reserve old IDs
A suggestion from Ich at the most recent lab meeting: We keep a list of IDs that have a valid `/node/` page on OldCantus. Every time we make a new object, we ensure it is assigned an ID that is not on the list. And every time someone tries to visit a `/node/` page, we check the list of IDs to see if the specified ID is on the list. If it's on the list, we redirect to the proper page - there will be no ambiguity - and if it's not, we return a 404 error.

Considering there are no significant gaps in the IDs OldCantus assigns to objects, and since Postgres (NewCantus) picks the largest ID in a given table and assigns any newly created record an ID that is one larger, **we can actually simplify this procedure in NewCantus. Instead of keeping a list of reserved IDs, we take note of the largest ID, `n`, that leads to a valid `/node/` page.** Once the data is synced for the last time, we create a placeholder source, chant, sequence and user (i.e. indexer), each with an ID of `n+1` (these placeholders will be unpublished). Whenever a new object is created in one of those tables, it will be assigned an ID that is greater than `n+1`, guaranteeing that no object will be created that could conflict with an existing valid `/node/` path. Now, whenever someone tries to visit a `/node/` page, we check whether the specified ID <=`n`. If it is, we go in search of the single object it could match and redirect the user to the proper detail page. And if the specified ID > `n`, we return a 404 error.

Benefits:
- We're not fighting against Django - once we set up our placeholders with specified IDs, IDs are assigned as usual in Django
- the logic of deciding whether/where a redirect should happen is simple
- no ambiguity, so users can be sent to the correct page without much ado

Drawbacks:
- not many?
- new objects have IDs that don't lead to valid `/node/` pages (maybe a tiny bit inconsistent, but not a problem)

**I think that the last of these options, Reserve Old IDs (the simplified version with placeholder objects) is the solution I'd advocate most strongly for**

PythonSemicolon:
This fix allows users with previously bookmarked URLs from OldCantus using the /node/ path to view the appropriate Chant, Source, Sequence, User/Indexer or Article page in NewCantus. Note that the Indexer pages are now redirected to User pages (since indexers are users in NewCantus).

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -11,6 +11,8 @@ from django.db.models import Q
 import csv
 
 from faker import Faker
+
+from users.models import User
 from .make_fakes import (
     make_fake_century,
     make_fake_chant,
@@ -3782,3 +3784,49 @@ class ChangePasswordViewTest(TestCase):
         self.assertEqual(
             response_2.status_code, 200
         )  # if login failed, status code will be 302
+    
+class NodeURLRedirectTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_chant_redirect(self):
+        example_chant_id = 243652  # from issue #563
+        Chant.objects.create(id=example_chant_id)
+        response_1 = self.client.get(reverse("redirect-node-url", args=[example_chant_id]))
+        self.assertEqual(response_1.status_code, 302)
+    
+    def test_source_redirect(self):
+        fake = make_fake_source()
+        example_source_id = fake.id
+        response_1 = self.client.get(reverse("redirect-node-url", args=[example_source_id]))
+        self.assertEqual(response_1.status_code, 302)
+    
+    def test_sequence_redirect(self):
+        example_sequence_id = 628334 # from issue #563
+        Sequence.objects.create(id=example_sequence_id)
+        response_1 = self.client.get(reverse("redirect-node-url", args=[example_sequence_id]))
+        self.assertEqual(response_1.status_code, 302)
+    
+    def test_article_redirect(self):
+        # there is no way to generate a fake article, nor get an existing one here
+        # instead, this has been tested locally by updating the id of a NewCantus source manually to match OldCantus, and the redirect works
+        self.assertTrue(True)
+    
+    def test_indexer_redirect(self):
+        example_indexer_id = 706860 # from issue #563
+        example_matching_user_id = 251425 # from issue #563
+        User.objects.create(id=example_matching_user_id, old_indexer_id=example_indexer_id)
+        response_1 = self.client.get(reverse("redirect-node-url", args=[example_indexer_id]))
+        self.assertEqual(response_1.status_code, 302)
+
+    def test_bad_redirect(self):
+        invalid_node_id = 123456
+        response_1 = self.client.get(reverse("redirect-node-url", args=[invalid_node_id]))
+        self.assertEqual(response_1.status_code, 404)
+
+    def test_redirect_above_limit(self):
+        over_limit_node_id = 1000001
+        url = reverse("redirect-node-url", args=[over_limit_node_id])
+        print(f'url: {url}')
+        response_1 = self.client.get(reverse("redirect-node-url", args=[over_limit_node_id]))
+        self.assertEqual(response_1.status_code, 404)
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -3864,7 +3864,9 @@ class NodeURLRedirectTest(TestCase):
         self.assertEqual(response_1.status_code, 404)
 
     def test_redirect_above_limit(self):
+        # generate dummy object with ID outside of valid range
         over_limit_node_id = 1000001
+        Chant.objects.create(id=over_limit_node_id)
 
         # ID above limit
         response_1 = self.client.get(reverse("redirect-node-url", args=[over_limit_node_id]))
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -1,6 +1,7 @@
 import random
 from django.urls import reverse
 from django.test import TestCase
+from articles.tests.test_articles import make_fake_article
 from main_app.views.feast import FeastListView
 from django.http.response import JsonResponse
 import json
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +625,40 @@ def content_overview(request):
     return render(
         request, "content_overview.html", {"objects": recently_updated_50_objects}
     )
+
+def redirect_node_url(request, pk):
+    not_found = handle404(request, HttpResponseNotFound)
+    if pk >= 1_000_000:
```
jacobdgm, line 643: Can we add a comment around here explaining that we're manually creating unpublished dummy objects in the database to ensure that all subsequent objects will have IDs >= 1_000_000?

django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +625,40 @@ def content_overview(request):
     return render(
         request, "content_overview.html", {"objects": recently_updated_50_objects}
     )
+
+def redirect_node_url(request, pk):
```
jacobdgm: Can we have a docstring here, briefly outlining why we need this view: backwards compatibility, different numbering systems, and so on?

django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +625,40 @@ def content_overview(request):
...
+        (Source, 'source-detail'),
+        (Sequence, 'sequence-detail'),
+        (Article, 'article-detail')
+    ]
+
+    user_id = get_user_id(pk)
+    if get_user_id(pk) is not None:
+        return redirect('user-detail', user_id)
+    
+    response = next((redirect(view, pk) for (rec_type, view) in possible_types if record_exists(rec_type, pk)), not_found)
```
jacobdgm: I can't tell what's going on here. Can this be split into multiple lines or something?

django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +625,40 @@ def content_overview(request):
...
+# used to determine whether record of specific type (chant, source, sequence, article) exists for a given pk
+def record_exists(rec_type, pk):
+    try:
+        result = rec_type.objects.get(id=pk)
+        print(f'result: {result}')
+        return True
+    except rec_type.DoesNotExist:
+        return False
+
+# special case for User (using indexer ID)
```
jacobdgm: can this comment be fleshed out a bit, and maybe turned from a comment into a docstring? Something along the lines of "indexers were originally stored in the general Node table in OldCantus, but now they're represented as users, so we need to check `old_indexer_id` rather than `id`"?

jacobdgm (changes requested):
On the whole, this looks good! In order to merge this, it would be nice to have:
- Tests
- Type annotations, and
- Docstrings, especially if they can explain how we ended up needing to handle `/node/` urls in the first place

django/cantusdb_project/main_app/views/views.py
```diff
@@ -643,20 +655,30 @@ def redirect_node_url(request, pk):
...
     try:
         result = rec_type.objects.get(id=pk)
         print(f'result: {result}')
         return True
     except rec_type.DoesNotExist:
         return False
 
-# special case for User (using indexer ID)
-def get_user_id(pk):
+def get_user_id(pk: int) -> int:
```
jacobdgm: should be `-> Optional[int]` because it sometimes returns `None`

django/cantusdb_project/main_app/views/views.py
jacobdgm: The comment helps, I guess, but ideally I'm able to tell what's going on just by looking at the code. This is not a place where we need this function to run really efficiently - much better, in my opinion, that the code reads clearly!

Stated another way, a clever/subtle solution like this works fine until it doesn't, at which point we get to deal with a clever/subtle bug, rather than an easily identifiable bug.

If you feel really strongly about this, we can leave it as is. But I think it would be better to use a `for` loop here or something.

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -3782,3 +3784,49 @@ def test_change_password(self):
         self.assertEqual(
             response_2.status_code, 200
         )  # if login failed, status code will be 302
+    
+class NodeURLRedirectTest(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_chant_redirect(self):
+        example_chant_id = 243652  # from issue #563
```
jacobdgm: in this situation, and in similar places below, can we just choose a random number below 1000000? It doesn't need to be the id of an actual chant, because the test database doesn't have access to any actual objects. The fact that 243652 is the ID of an actual chant is an irrelevant detail.

jacobdgm (changes requested):

django/cantusdb_project/main_app/views/views.py
PythonSemicolon: updated in latest commit

django/cantusdb_project/main_app/tests/test_views.py
PythonSemicolon: redid testing to be more complete and self-documenting

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -3782,3 +3785,87 @@ def test_change_password(self):
...
+        expected_url = reverse("article-detail", args=[example_article_id])
+        
+        self.assertEqual(response_1.status_code, 302)
+        self.assertEqual(response_1.url, expected_url)
+    
+    def test_indexer_redirect(self):
+        # generate dummy object with ID in valid range
+        example_indexer_id = random.randrange(1, 1000000)
+        example_matching_user_id = random.randrange(1, 1000000)
+        User.objects.create(id=example_matching_user_id, old_indexer_id=example_indexer_id)
```
jacobdgm, line 3850: Instead of `User`, we should use `get_user_model()` - https://docs.djangoproject.com/en/4.2/topics/auth/customizing/#django.contrib.auth.get_user_model.

This means removing an import statement at the top of the file, too.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +626,64 @@ def content_overview(request):
...
+# used to determine whether record of specific type (chant, source, sequence, article) exists for a given pk
+def record_exists(rec_type: BaseModel, pk: int) -> bool:
+    try:
+        result = rec_type.objects.get(id=pk)
+        print(f'result: {result}')
+        return True
+    except rec_type.DoesNotExist:
+        return False
+
+def get_user_id(pk: int) -> Optional[int]:
```
jacobdgm: `def get_user_id_from_old_indexer_id(indexer_id: int)` or even `def get_user_id_from_indexer_id(...)` would make it clearer what this function does.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -620,3 +626,64 @@ def content_overview(request):
...
+            return redirect(view, pk)
+        
+    # if it reaches the end of the types with finding an existing object, a 404 will be returned
+    return not_found
+
+# used to determine whether record of specific type (chant, source, sequence, article) exists for a given pk
+def record_exists(rec_type: BaseModel, pk: int) -> bool:
+    try:
+        result = rec_type.objects.get(id=pk)
+        print(f'result: {result}')
```
jacobdgm: remove this print statement?

jacobdgm (changes requested):

django/cantusdb_project/main_app/tests/test_views.py
PythonSemicolon: done

django/cantusdb_project/main_app/views/views.py
PythonSemicolon: changed

django/cantusdb_project/main_app/views/views.py
PythonSemicolon: removed! oversight on my part

jacobdgm (approved):
Excellent! Thanks!

PR #721, opened 2023-06-08
Set up password reset feature

### LINKED ISSUE #413: login page--request password
On OldCantus you can initiate a request for a new password--on NewCantus it's just an eternal log-in loop.

Might also be nice to have a way of initiating the request to become a new contributor,  but maybe we want to keep new accounts separate from fixing old ones (so that new users can be vetted first, but old users can get automatic password resets).

jacobdgm:
This PR sets up a system for users to request a password reset link, and subsequently to reset their password, in the case they forget their password. It fixes #413.

Before we put these changes up on staging/production, **we should consider whether to implement some sort of captcha system** to prevent people/bots from using this page to spam email addresses with lots of emails. We should discuss this at lab meeting this week. (edit: I just opened an issue for this: #728)

In addition, this PR includes tweaks to a couple of existing authentication-related pages, adding a link to initiate the password-reset process, clarifying the purpose of each page and adding a title tag in one case.

The login page:
Screenshot 2023-06-08 at 4 42 41 PM

The request-a-new-password page:
Screenshot 2023-06-08 at 4 43 49 PM

The we-successfully-sent-you-an-email-with-a-reset-password-link page:
Screenshot 2023-06-08 at 4 44 34 PM

The specify-a-new-password page:
Screenshot 2023-06-08 at 4 47 33 PM

The you-successfully-reset-your-password page:
Screenshot 2023-06-08 at 4 48 23 PM


django/cantusdb_project/main_app/urls.py
```diff
@@ -95,6 +101,45 @@
...
+    ),
+    path(
+        # here, the user can specify their new password
+        "reset/<uidb64>/<token>",
+        # Template filename below includes an intentional trailing underscore.
+        # Without this extra underscore, django will find a django built-in template when
+        # it looks for registration/password_reset_confirm.html, rather than finding our
+        # own custom template
+        PasswordResetConfirmView.as_view(
+            template_name="registration/password_reset_confirm_.html"
```
lucasmarchd01: Good catch! I don't really like the trailing underscore though. Can this be renamed to something like reset_password_confirm.html to avoid the built in django template but have a nicer looking filename? You could similarly change the other templates to start with reset_password to maintain continuity.

lucasmarchd01 (changes requested):

lucasmarchd01:
Should there also be test functions for this new feature?

fujinaga:
Try Google's reCAPTCHA?

jacobdgm:
It looks as though there's a 3rd-party library - https://pypi.org/project/django-recaptcha/ - that will make integrating a Google reCAPTCHA into this form fairly straightforward.

It appears that we can perform 1 million verifications per month for free on a single google account, which seems like the an appropriate order of magnitude (we'll use this for the contact page as well - my guess is that the email reset feature will be used maybe 10-100 times per month, and the contact feature say 100, but I have no idea how many bots/crawlers the site will have to deal with each month). @fujinaga and @kqct, does DDMAL have a google account in which we can set this up? (if so, let's continue this conversation over email, of course)

fujinaga:
Try to minimize the use of third-party libraries, in general. 

Separate this captcha issue and mark it medium priority.

The password reset should be of super high priority.

jacobdgm:
Okay - all that's left to do with this PR is to write tests, so I'll work on figuring that out, and then we can merge it.

I'll open a new issue re: the captcha.

jacobdgm:
I was having trouble properly writing tests for this. I just opened #732 in order to move forward with this feature. I've tested the password reset feature many times and it works, and once we put this on staging, we can ask everyone in the lab to test it as well.

dchiller (approved):

Issue #741, opened 2023-06-14
We need to sync users from OldCantus


jacobdgm:
We discovered yesterday that the [user synchronization process](https://github.com/DDMAL/CantusDB/wiki/Migrating-users-from-old-to-new-CantusDB) is not included in the main [data migration process](https://github.com/DDMAL/CantusDB/wiki/Data-migration-from-the-old-cantus-to-the-new-cantus) instructions. Further, the necessary script depends on several CSV files that are to be "obtained from Junhao".

There was some talk of manually making changes to the user list following the final data migration, but this is impractical because users will have been given access to sources - we don't want anyone (developer or Debra) to have to re-grant access to a bunch of necessary sources for a bunch of users.

@jwang44 has graciously agreed to try to do his best to recreate these files around when we do the final data synchronization. Many thanks, Junhao!

(UPDATE: the following is now complete) **We also need to update these scripts to assign users a random password** (which they will then be able to change via the password reset links), rather than a set placeholder as is done currently.

lucasmarchd01:
We've assigned users random passwords and this has been completed.

PR #742, opened 2023-06-14
In add_all_users.py, give new users a random password

jacobdgm:
This PR updates the `add_all_users` management command, changing it so newly created users are assigned a random password rather than the placeholder that has been used up until now.

This is a necessary step towards resolvin\g #741 

Along the way, I noticed that newly-created users' `full_name` fields were not being given a default value. The script has been updated to also populate this field.

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/add_all_users.py
+++ b/django/cantusdb_project/main_app/management/commands/add_all_users.py
@@ -32,7 +32,7 @@ class Command(BaseCommand):
                 if created:
                     user.first_name = name
                     user.last_name = surname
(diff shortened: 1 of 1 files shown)


lucasmarchd01 (approved):
Just to clarify, is the idea here that new users will be given a random password and use the reset password feature to get a new one?

jacobdgm:
> Just to clarify, is the idea here that new users will be given a random password and use the reset password feature to get a new one?

"new users", in the sense of any user accounts that were present on OldCantus and have been newly created on NewCantus. So in this sense, yes, you're exactly right.

PR #770, opened 2023-06-20
Remove instances of "changed_initial_password"

### LINKED ISSUE #745: we need to get rid of `User.changed_initial_password`
Currently, newly created user accounts have a `changed_initial_password` field that defaults to `False`.

If a user logs in and their `changed_initial_password==False`, we show them a message telling them they need to reset their password. On our main password reset page (i.e. the page where they enter their current password and set a new password), we flip this flag to `True` when a user successfully resets their password.

But, we don't currently have any functionality to flip the flag to `True` when a user goes through the request-a-password-reset-link process.

Contingent on the request-a-password-reset-link process working, it would be better (i.e. more secure, and less complicated) to give users a random password rather than a placeholder. If/when we do this, we can remove the `changed_initial_password` field from the User model, and remove all references to it in the code (including the please-reset-your-password message).

lucasmarchd01:
This PR focuses on enhancing our password management system by implementing a more secure and user-friendly password reset mechanism. Previously, we utilized a "changed_initial_password" flag to prompt users to reset their passwords if they hadn't done so already. However, with the introduction of password reset via email, it is more appropriate to remove this field and all associated references from the codebase. Resetting the password through the password reset email provides a safer and more reliable approach.

Note that we need to make migrations again after removing this field from the Users model.

Resolves #745

### DIFF
--- a/django/cantusdb_project/main_app/urls.py
+++ b/django/cantusdb_project/main_app/urls.py
@@ -55,7 +55,7 @@ from main_app.views.source import (
     SourceListView,
 )
 from main_app.views.user import (
-    CustomLoginView,
+    LoginView,
     CustomLogoutView,
     IndexerListView,
     UserDetailView,
@@ -73,7 +73,7 @@ urlpatterns = [
     # login/logout/user
     path(
         "login/",
-        CustomLoginView.as_view(redirect_authenticated_user=True),
+        LoginView.as_view(redirect_authenticated_user=True),
         name="login",
     ),
     path(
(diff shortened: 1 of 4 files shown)


jacobdgm (approved):

jacobdgm:
going to hold off on merging for the time being. Once we've finished renumbering articles and creating dummy objects for /node/ redirects, we can merge this and run `makemigrations` and `migrate` and not need to worry about this creating complications.

PR #771, opened 2023-06-20
Add permissions for editors to proofread their created sources

### LINKED ISSUE #491: should editors get automatic proofreader access to their own sources?
It seems weird to me that on sources you (an editor) create, you can always access the Volpiano editor (assuming the source is non-empty!) but you have to be added as an editor to access the proofreading form. 
image
image

If there were something preventing you from proofreading your own source as an intentional part of the workflow, this might be sort of useful (?) but as it is it just adds an extra step to get permissions you should in theory already have.

### LINKED ISSUE #736: New source created doesn't show up in home page
I just created a new source from my account. It doesn't show up immediately on the home page in the My Sources sidebar (even after refreshing the page and re-logging in). It does however appear in the database as well as in the my-sources page.

lucasmarchd01:
This PR is related to changing the permissions system to allow editors to proofread the sources they create. Previously, editors did not automatically have proofreading permission for the sources they created unless explicitly added as editors. This solution addresses this issue by automatically adding the source to the `sources_user_can_edit` field of the user when an editor creates a new source. Additionally, contributors should not have access to the proofreading form. In the "My Sources" page, the "Proofreading form" link is hidden for contributors, as it should only be available to the editor and project manager groups. 

I believe all of the permissions are correct here, but do check the permissions spreadsheet to make sure before approving.

resolves #491, fixes #736

### DIFF
--- a/django/cantusdb_project/main_app/templates/user_source_list.html
+++ b/django/cantusdb_project/main_app/templates/user_source_list.html
@@ -22,11 +22,13 @@
                                         Full text &amp; volpiano editor
                                     </a>
                                 </td>
-                                <td class="h-25" style="width: 30%; text-align:center">
-                                    <a href="{% url "chant-proofread" source.pk %}">
-                                        Proofreading form
-                                    </a>
-                                </td>
+                                {% if request.user|has_group:"project manager" or request.user|has_group:"editor" %}
+                                    <td class="h-25" style="width: 30%; text-align:center">
+                                        <a href="{% url "chant-proofread" source.pk %}">
+                                            Proofreading form
+                                        </a>
+                                    </td>
+                                {% endif %}
                             {% else %}
                             <td class="h-25" style="width: 30%; text-align:center" colspan="2">
                                 Source contains no chants –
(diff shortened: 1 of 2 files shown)


jacobdgm:
As far as I can tell from the user permissions doc, people in the "Editor" group should only be able to access the Proofread page for specified sources (i.e. if they're a "current editor", I think), rather than on all sources.

lucasmarchd01:
@jacobdgm Yes, and that is still the case with this PR. Editors will retain their permission to access the proofread page for sources where they are listed as "current editors." The only modification we're making is granting "current editor" status to sources they create, enabling them to access the proofread page for sources that they have created as well.

jacobdgm:
Just to put in writing what we just discussed: In `user_source_list.html`, we're adding proofread-page links if the user is in certain groups. And we know that the user will have permission to visit all the proofread pages, since _we're already filtering the sources - we're only displaying sources that the user already has permission to edit._ Got it. Approving!

jacobdgm (approved):

PR #776, opened 2023-06-21
change url path from /chant-search-ms/ to /searchms/

### LINKED ISSUE #565: Chant Search Manuscript view - change URL path to match OldCantus
I understand we're trying to keep URLs the same between OldCantus and NewCantus, but there's a difference in the Chant Search Manuscript view. OldCantus uses `/searchms/` (e.g. https://cantus.uwaterloo.ca/searchms/123610?t=est), whereas NewCantus uses `/chant-search-ms/` (e.g. http://206.12.93.196/chant-search-ms/123610?t=est).

This doesn't strike me as a particularly vital difference - I doubt many people will have bookmarked and or cited a Search Manuscript page. But this would be a fairly simple fix, so we may as well make NewCantus work the same as OldCantus in this case.

Bigger picture question: how important is it that all URL paths match between OldCantus and New? @annamorphism, do you have a sense of this?

jacobdgm:
fixes #565

### DIFF
--- a/django/cantusdb_project/main_app/urls.py
+++ b/django/cantusdb_project/main_app/urls.py
@@ -319,7 +319,7 @@ urlpatterns = [
     ),
     # misc search
     path(
-        "chant-search-ms/<int:source_pk>",
+        "searchms/<int:source_pk>",
         ChantSearchMSView.as_view(),
         name="chant-search-ms",
     ),


dchiller (commented):
What about places where `chant-search-ms/[source_id]` is referenced?

jacobdgm:
> What about places where `chant-search-ms/[source_id]` is referenced?

I just looked through the code base, and can confirm we never build urls like this directly - it's all `reverse("chant-search-ms", args=[source.id])` and `{% url "chant-search-ms" source.id %}` and the like.

All tests are passing, so I'm pretty sure this change doesn't break anything.

dchiller:
Gotcha! I didn't know how reverse worked...so now I do.

dchiller (approved):

dchiller:
Also, just want to note the suggestion from Andrew in the issue that this will close... not sure there's anything to do there but I noticed it's not addressed here so just don't want it to be an unwitting loss

jacobdgm:
> Also, just want to note the suggestion from Andrew in the issue that this will close... not sure there's anything to do there but I noticed it's not addressed here so just don't want it to be an unwitting loss

This is a good point, and thanks for noticing this! I added a comment on #565 explaining why I think innovation in the URL path for this page is not necessary.

PR #785, opened 2023-06-23
Prevent Chant List page from being accessed for sources in the Bower segment

### LINKED ISSUE #458: Only sources from the Cantus Database segment should appear in the Browse Chants "Source" select element
Jacob: EDITING this to put some conclusions, which otherwise would be buried in the discussion, near the top:

- Since sources from the Bower segment don't contain any chants, they _appear_ to be buggy when a user attempts to access them via the Browse Chants page (i.e. 0 chants are displayed)
- Thus, for the time being **we should not list sources from the Bower segment in the "Source" select element on the Browse Chants page.**
  - **Also, in the source navigation sidebar, for sources in the Bower segment, we should not display the "view all chants" link that currently links to the Browse Chants page** 
- At some point in the future, we may want to better harmonize how CantusDB handles/displays `Sequence`s (especially as compared to `Chant`s with a genre of "Sq" (i.e. "sequence"))

/jacob

When looking at chants in one source, you can also go to another via a drop down.  However, the dropdown displays all possible sources--including those in the Bower Sequence segment. If you choose one of those sequence sources, though, you end up seeing it has "0 chants", presumably because by default you are looking at "chants in this source in the Cantus DB segment". 
image
You can actually see this source just fine via  the Clavis Sequentiarum source list (http://206.12.88.113/source/124052) ; it's getting there from this dropdown that's a problem. http://206.12.88.113/chants/?source=124052&folio=&feast=&search_text=&genre=

This bug seems to also exist on OldCantus--it's just sufficiently weird that I never noticed it before. Still, either the Bower sources shouldn't be in the dropdown, or they need to correctly display their chants.

jacobdgm:
fixes #458

This PR:
- ensures that, on the Chant List (i.e. "Browse Chants") page, only sources from the Cantus segment appear in the dropdown
- ensures that the Chant List page can only be accessed for sources in the Cantus segment (it now raises a 404 for sources not in the segment)
- hides, on the Source Detail page for sources not in the Cantus segment, the link to the Chant List page.
- updates tests that were broken by this change, ensuring all sources created during the tests belong to the right segment
- adds a test for the Chant List page to ensure it is accessible (200 http status code) for sources in the Cantus Segment and inaccessible (404) for sources in the Bower segment.
- adds a test for the Source Detail page to ensure that the link to the Chant List page is present when it should be and hidden when it shouldn't

### DIFF
--- a/django/cantusdb_project/main_app/templates/source_detail.html
+++ b/django/cantusdb_project/main_app/templates/source_detail.html
@@ -175,7 +175,10 @@
                         </select>
 
                         <br>
-                        <a href="{% url "chant-list" %}?source={{ source.id }}" class="guillemet" target="_blank">View all chants</a>
+                        {% if source.segment.id == 4063 %}
+                            {# only display this link for sources in the CANTUS segment #}
+                            <a href="{% url "chant-list" %}?source={{ source.id }}" class="guillemet" target="_blank">View all chants</a>
+                        {% endif %}
                         <a href="{% url "chant-index" %}?source={{ source.id }}" class="guillemet" target="_blank">View full inventory</a>
                         <a href="{% url "csv-export" source.id%}" class="guillemet" target="_blank" download="{{ source.id }}.csv">CSV export</a>
                         <a href="{% url "chant-search-ms" source.id %}" class="guillemet" target="_blank">Search chants in this manuscript</a>
--- a/django/cantusdb_project/main_app/tests/make_fakes.py
+++ b/django/cantusdb_project/main_app/tests/make_fakes.py
@@ -263,11 +263,14 @@ def make_fake_rism_siglum() -> RismSiglum:
     return rism_siglum
 
 
-def make_fake_segment(name=None) -> Segment:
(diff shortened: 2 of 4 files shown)


django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -175,7 +175,10 @@ <h4>{{ source.siglum }}</h4>
                         </select>
 
                         <br>
-                        <a href="{% url "chant-list" %}?source={{ source.id }}" class="guillemet" target="_blank">View all chants</a>
+                        {% if source.segment.id == 4063 %}
```
dchiller, line 178: I assume that 4063 here is the auto-assigned ID of the CantusDB segment in the database?

dchiller (approved):

django/cantusdb_project/main_app/templates/source_detail.html
jacobdgm: It's the same ID used on OldCantus afaik, so we manually assigned it. https://github.com/DDMAL/CantusDB/wiki/Segment-Model

PR #786, opened 2023-06-23
Chant List page: ensure search_bar search behaves the same as keyword+starts_with

### LINKED ISSUE #516: keyword/search_bar search give different results
If I search for "urbs" in the keyword search of Chant search, I get 318 results:
http://206.12.88.113/chant-search/?op=contains&keyword=urbs&office=&genre=&cantus_id=&mode=&feast=&position=&melodies=
image

BUT if I search the same word in the little magnifying-glass box on the upper right, I now get only 256: http://206.12.88.113/chant-search/?search_bar=urbs 
image

If the two searches are going to work differently, it needs to be made obvious somewhere what is different, so people know which one to use.

jacobdgm:
fixes #516

This PR:
- fixes a bug while conducting a search_bar search, where chants from unpublished sources were not being displayed, whether or not the user is logged in (now they're displayed IFF the user is logged in)
- fixes a bug while conducting a search_bar search, where we would only search the standard_spelling fulltext field, rather than all the full text fields.

In sum, now search_bar searches work the same as keyword+starts_with searches.

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -750,8 +750,12 @@ class ChantSearchView(ListView):
         display_unpublished = self.request.user.is_authenticated
         # if the search is accessed by the global search bar
         if self.request.GET.get("search_bar"):
-            chant_set = Chant.objects.filter(source__published=True)
-            sequence_set = Sequence.objects.filter(source__published=True)
+            if display_unpublished:
+                chant_set = Chant.objects.all()
+                sequence_set = Sequence.objects.all()
+            else:
+                chant_set = Chant.objects.filter(source__published=True)
+                sequence_set = Sequence.objects.filter(source__published=True)
             if self.request.GET.get("search_bar").replace(" ", "").isalpha():
                 # if search bar is doing incipit search
                 incipit = self.request.GET.get("search_bar")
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -750,17 +750,29 @@ def get_queryset(self) -> QuerySet:
...
-            chant_set = Chant.objects.filter(source__published=True)
-            sequence_set = Sequence.objects.filter(source__published=True)
+            if display_unpublished:
+                chant_set = Chant.objects.all()
+                sequence_set = Sequence.objects.all()
+            else:
+                chant_set = Chant.objects.filter(source__published=True)
+                sequence_set = Sequence.objects.filter(source__published=True)
             if self.request.GET.get("search_bar").replace(" ", "").isalpha():
                 # if search bar is doing incipit search
```
dchiller, line 760: Is this comment correct?

dchiller (approved):

dchiller:
One more thing: In NewCantus, it looks like the search bar has text suggesting it is "Incipit/CantusID" search, but with this change, that would be out of date yes?

image

django/cantusdb_project/main_app/views/chant.py
jacobdgm: we just discussed this in person, but leaving a tiny summary in case people read this in the future: while it's true that it's not searching only the `incipit` field, it's searching for the search term at the very beginning of each of the text fields - from the user's point of view, they're searching for an incipit.

Issue #438, opened 2022-11-30
"My sources" is weird


annamorphism:
On both staging and production, the My Sources sidebar doesn't show very many sources at all. I only see 6, but there are actually dozns assigned to me; and I don't see the sources I've created at all (the sidebar for this only  shows up within http://206.12.93.196/my-sources/, not on the home page, and it's empty.)

The displayed name is also weird--it should be the whole manuscript siglum, not just the library siglum. see New and Old below.

![image](https://user-images.githubusercontent.com/67451875/204893178-b4175024-1d1d-465d-97ea-bf2c1faad2c5.png)
![image](https://user-images.githubusercontent.com/67451875/204893254-7709a59a-ab2b-4e3c-adb5-82d6a28473b8.png)

jacobdgm:
> The displayed name is also weird--it should be the whole manuscript siglum, not just the library siglum.

Can you check the sources' `rism_siglum`s? In each place where you're seeing `D-TRs`, the page is displaying the source's `rism_siglum`, so I think this could be a data issue.

That said, each source has a `siglum` field that is separate from its `rism_siglum` field, and often includes additional information (see [Source Model](https://github.com/DDMAL/CantusDB/wiki/Source-Model#fields) in the Wiki). Maybe we should be using the `siglum` field here rather than `rism_siglum`?

annamorphism:
> That said, each source has a `siglum` field that is separate from its `rism_siglum` field, and often includes additional information (see [Source Model](https://github.com/DDMAL/CantusDB/wiki/Source-Model#fields) in the Wiki). Maybe we should be using the `siglum` field here rather than `rism_siglum`?

Yes, I think this should be the `siglum` field.  (`rism_siglum` may be empty for some sources, but `siglum` should never be; plus it is more informative when there are many sources from the same library.)

jacobdgm:
Question: can part of this be explained by what's going on with [adding users to sources](https://github.com/DDMAL/CantusDB/issues/450#issuecomment-1357932581) - i.e. you're an indexer, and you have sources associated with you as an indexer, but those aren't necessarily the right relations for changing your permissions for sources as a user?

jacobdgm:
Also, I can find you in the indexer list on [production](http://206.12.93.196/indexers/) but not on [staging](http://206.12.88.113/indexers/)... I'm not sure how concerned to be about this.

annamorphism:
[...]

PR #787, opened 2023-06-23
On homepage, in My Sources card, use source.siglum instead of .rism_siglum

jacobdgm:
goes part of the way toward fi\xing #438


jackyyzhang03 (approved):
Looks good to me

PR #803, opened 2023-06-28
Create json-cid API

### LINKED ISSUE #378: `json-cid` API
OldCantus has a `json-cid` API: https://cantus.uwaterloo.ca/json-cid/006583a .

The format of the data it returns looks very similar to that of CantusIndex's `json-con` API. In OldCantus's source code, there are numerous calls to other databases' `json-cid` pages, so it's likely that CI uses these APIs for compiling the data in its `json-con` API. So, we'll probably have to implement it if we want CI to be able to display concordances from CD.

jacobdgm:
This PR fixes #378, implementing an API that was present in OldCantus.

This PR also includes several tests for the new view.

While I was working in `main_app/views/views.py`, I took the opportunity to refactor a few of our other API views, pulling out nested functions that I somehow thought were a good idea a year ago. I also added some type annotations and docstrings.


dchiller (approved):
Approving...noting that API documentation on the CantusDB wiki will need to be updated with these changes.

Issue #813, opened 2023-07-06
If NewCantus is open and inactive for 5+ minutes, 502 errors ensue


JoyfulGen:
Steps:

1. Open a NewCantus tab on the Home Page;
2. Do something else for at least five minutes;
3. Back to NewCantus, select any source from the "Jump to Source" drop-down menu;
4. This will likely lead to a 502 Bad Gateway error. If it doesn't the first time, click on another source.

I've tried this four times, and so far the 502 error happens every time.

When it does happen, it takes a while for NewCantus to warm up again. Clicking on different sources produces the same error. What has worked for me so far is clicking on a series of innocuous and easy-to-load pages, such as things under the "About" header; after a few clicks, I can generally come back to the sources menu and select one successfully. (I'm going to give this a priority: medium tag; I hope it's not too brash.)

annamorphism:
Thanks for looking into this, @JoyfulGen --I'd noticed a surprising number of 502 but hadn't managed to replicate them consistently!

jacobdgm:
I noticed 502s on certain Source Detail pages, but only on Production and never on Staging or locally - I don't know why this should happen.

While I was investigating this issue, I noticed a optimization we could be making on many of our pages that we weren't at the time. #834 implements this optimization, and testing locally, I found it cut load times for a Source Detail page roughly in half. I doubt it will fully solve this issue, but might alleviate things somewhat.

@kqct, are there any tests we can run on production to try to identify a bottleneck that might be causing this?

ahankinson:
502s generally happen when the front-end server (nginx) can't contact the upstream application server within a set amount of time. This typically has a number of causes, some of which may be found in combination:

1. Your upstream server cannot handle the amount of traffic coming to it. If someone's request takes a long time to process, your app server will stop accepting requests until it's returned.
2. Your front-end server has too small a timeout set. If requests take too long to return, the frontend server may close the connection. Sometimes this results in 504 Timeout error, but this depends on your configuration.
3. Your upstream app server has too small a timeout set. For example, gunicorn may have a timeout set, after which it closes the connection.

What can you do to fix this?

1. Make your responses as fast as possible. Your site will pause while slow queries are processed, so the faster a query is returned, the more traffic your site can handle.
2. Increase the number of workers. In the DIAMM gunicorn process handler we use 5 (https://github.com/diamm/diamm/blob/master/gunicorn_start.sh). I see you do as well.
[...]

PR #834, opened 2023-07-12
Optimize templates to use `.exists` instead of `.all` within `if` tags

jacobdgm:
This PR goes through all the templates, and substitutes `.exists` for `.all` wherever it was found in an `if` tag. It was prompted by #813, when I was looking into why the source detail page might be taking so long to load. Testing locally, the Source Detail page spends about half as long sending SQL requests with this change, compared to how it was previously.

Timing results on other pages are inconsistent: Request times are decreased significantly on certain pages, while they increase or decrease request by a handful of milliseconds on some pages that already load quickly. We know that Staging and Production have different limitations than local machines, so since there's a good theoretical reason to use `.exists` in place of `.all`, and it improves the clarity of the code in the templates, I think this change is a good one in spite of the variability observed while testing.

It also fixes a discovered bug on the Chant Detail page, where if a request to gregorien.info would fail, we'd hit a `NameError` because we forgot to set a variable.

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -499,14 +499,14 @@
                                     <br>
                                 {% endif %}
 
-                                {% if source.notation.all %}
+                                {% if source.notation.exists %}
                                     <b>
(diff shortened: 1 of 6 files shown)


lucasmarchd01 (approved):

dchiller (approved):

dchiller:
This seems like a good change!

PR #841, opened 2023-07-12
Redirect user to Source Detail page after they create a source

### LINKED ISSUE #814: Redirect user to source detail page after they create a source
As mentioned in Kathleen's email with feedback on our production and staging severs.

When a source or chant is created, we display a success message but with an empty form still on the source/chant create page. This is confusing to users because it looks like none of the data saved into their source. Instead, we should redirect them back to the source/chant detail page after they save. A similar issue has been addressed here (#665) from the chant edit page, so we need a similar functionality on the source and chant create pages

lucasmarchd01:
Previously, if the user creates a source, the success message will be displayed on an empty create form. This is confusing because it looks like none of their data was saved to the source. Now, if they create a source they will be redirected to the source detail page.

Also removing the inheritance to SourceDetailView in the SourceEditView to avoid any unwanted behaviour. Reworking the logic inside of get_context_data outside of both of these views seemed to me like it would be more convoluted rather than helpful, especially since much of the logic is dependent on the object instance. Instead, SourceEditView now has it's own get_context_data that behaves similarly to that of SourceDetailView.

Fixes #814, but we may still need to consider whether we should do the same for the chant create page.

### DIFF
--- a/django/cantusdb_project/main_app/views/source.py
+++ b/django/cantusdb_project/main_app/views/source.py
@@ -211,13 +211,35 @@ class SourceCreateView(LoginRequiredMixin, UserPassesTestMixin, CreateView):
         return HttpResponseRedirect(self.get_success_url())
 
 
-class SourceEditView(
-    LoginRequiredMixin, UserPassesTestMixin, UpdateView, SourceDetailView
-):
+class SourceEditView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
     template_name = "source_edit.html"
     model = Source
     form_class = SourceEditForm
     pk_url_kwarg = "source_id"
+
+    def get_context_data(self, **kwargs):
+        source = self.get_object()
+        context = super().get_context_data(**kwargs)
+
+        if source.segment and source.segment.id == 4064:
+            # if this is a sequence source
+            context["sequences"] = source.sequence_set.order_by("s_sequence")
+            context["folios"] = (
+                source.sequence_set.values_list("folio", flat=True)
+                .distinct()
+                .order_by("folio")
+            )
+        else:
+            # if this is a chant source
+            folios = (
+                source.chant_set.values_list("folio", flat=True)
+                .distinct()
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/views/source.py
```diff
@@ -191,35 +191,56 @@ def test_func(self):
             return False
 
     def get_success_url(self):
-        return reverse("source-create")
+        return reverse("source-detail", args=[self.object.id])
 
     def form_valid(self, form):
         form.instance.created_by = self.request.user
-        source = form.save()
+        self.object = form.save()
```
jacobdgm, line 198: this looks fine, but I'm just curious - why was it necessary to substitute `self.object` for `source` here?

jacobdgm:
> Fixes https://github.com/DDMAL/CantusDB/issues/814, but we may still need to consider whether we should do the same for the chant create page.

It appears this PR actually goes ahead and does the analogous redirect for Chant Create, no?

jacobdgm (approved):

django/cantusdb_project/main_app/views/source.py
lucasmarchd01: @jacobdgm See the definition of `form_valid` in django's ModelFormMixin
```    
def form_valid(self, form):
        """If the form is valid, save the associated model."""
        self.object = form.save()
        return super().form_valid(form)
```

since `form_valid` is overwritten in our code, self.object is never saved and so we cannot access self.object in `get_success_url`. To avoid passing `source` as an argument to `get_success_url` I just replicated the behaviour from the ModelFormMixin `form_valid`

lucasmarchd01:
> > Fixes #814, but we may still need to consider whether we should do the same for the chant create page.
> 
> It appears this PR actually goes ahead and does the analogous redirect for Chant Create, no?

This actually wasn't the intended behaviour, I guess I initially made the change and forgot to put it back.

Issue #843, opened 2023-07-13
Indexing notes in "search manuscript"


annamorphism:
From an email from Debra:

"What I had wanted to check was that you have the search within the "indexing notes" field per source, as here:
https://cantusdatabase.org/searchms/123612

On the left side underneath Office/Mass and Genre, you can search for words that are entered in the indexing notes. I think maybe you have this already (I think I have checked with you before). The option I am using today is searching for a length greater than zero, so that I can pull up any/all notes in one particular source."

on OldCantus this looks like this, for reference 
<img width="954" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/ac5d0aaf-c2bd-4175-8be0-9b130cee619c">
and searching for " " (space) gives 180 results on this particular manuscript.

annamorphism:
addendum:
1. this probably wasn't initially included because it is comparatively new to OldCantus
2. This should only be on "search manuscript". Per Debra "Also, Jan put it in only one place: the "search this manuscript" page that you get to in the RH menu. This is the spot for it, I think, since you would likely only want to search within a source for the notes (for now, anyways - and the notes aren't tidy, so we don’t want to advertise them too much on the main search page)."

jacobdgm:
I'm working on this.

In starting to do so, I realized that the main Chant Search view and the Chant Search MS view, which share a template, have a bunch of copy-pasted and slightly tweaked code, so I'm going to spend some time writing additional tests, and then refactoring these views to make things simpler. (I was starting to work on this for the Search MS view, and in trying to make the clicking-on-column-headers thing work, I realized that with so much duplicated code, I'm really likely to introduce bugs if we don't clean these views up).

jacobdgm:
the refactoring work I mentioned in July has been partially implemented, but work remains on this. I'll try to get back to this once we sort out #1098.

PR #847, opened 2023-07-13
Improve layout of Chant Search, Chant Search MS views

### LINKED ISSUE #846: Formatting of Chant Search Page should be improved
In an email from Debra, relating feedback from another person in the Cantus community testing NewCantus:

> > Maybe it's just me, but I keep trying to type in the search word in the drop-down box ("Starts with" or "Contains") instead of in the "Keyword" box. I think it might be because they are the same size. Is there something one can easily do about that?
> 
> Me too! Now that you say it, I see what you mean. I clicked in the "Contains" box as well, and then realized what I had done. I'm sure we could train ourselves not to do this, but the "Contains" / "Starts with" box doesn't likely have to be that large. OR, how about we switch the order and have the box you type in come first? With "Contains/Start with" beside it to the right? Or is that strange and out of order the way you would think about it?

We should make the Contains/starts-with select element smaller.

jacobdgm:
This PR improves the layout of the Chant Search and Chant Search Manuscript views. It is a first step towards setting up #843 (I had started to work on implementing a "search indexing notes" field, and the page began to look quite terrible - in an attempt to make PRs that aren't enormous, this PR is a first step towards implementing the latter feature, where we can just take everything above "Office/Mass" and duplicate/adapt it below "Melodies" for the new field.)

It fixes #846.

Previously, the "Contains"/"Starts with" box was just as large as the "Keyword search" field, which was tripping up testers:

Screenshot 2023-07-13 at 1 15 45 PM

This PR makes the "Contains" box smaller.

The fields beneath this row have also been rearranged (bootstrap deals with 12 columns, but there were 7 items equally spaced on the lower line - without this rearrangement, nothing would align visually):

Screenshot 2023-07-13 at 1 18 05 PM

Also, previously, when the window became somewhat narrow, every field would go onto its own line:
Screenshot 2023-07-13 at 1 19 04 PM

Now, when the window becomes somewhat narrow, the fields pair up, which looks a bit neater:
Screenshot 2023-07-13 at 1 20 47 PM

And now, the fields take up full lines only when the window becomes very narrow. When this happens, the "Contains" box stays smaller than the "Keyword" field:
Screenshot 2023-07-13 at 1 22 07 PM

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_search.html
+++ b/django/cantusdb_project/main_app/templates/chant_search.html
@@ -31,23 +31,27 @@
 
     <form method="get">
         <div class="form-row align-items-end">
-            <div class="form-group m-1 col-lg">
+            <div class="form-group m-1 col-12">
+                <label for="keywordSearch"><small>Keyword search</small></label>
+            </div>
+        </div>
+        <div class="form-row align-items-end">
+            <div class="form-group m-1 col-lg-3 col-sm-3 col-3">
                 <select class="form-control custom-select custom-select-sm" id="opFilter" name="op">
                     <option selected value="contains">Contains</option>
                     <option value="starts_with">Starts with</option>
                 </select>
             </div>
-            <div class="form-group m-1 col-lg">
-                <label for="keywordSearch"><small>Keyword search</small></label>
+            <div class="form-group m-1 col-lg col-sm col-">
                 <input type="text" class="form-control form-control-sm" name="keyword" id="keywordSearch" value="{{ request.GET.keyword }}">
             </div>
         </div>
         <div class="form-row align-items-end">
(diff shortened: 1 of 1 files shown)


lucasmarchd01:
This might be too nit-picky, but the text "keyword search" is ever so slightly higher above the widget than the other fields.

lucasmarchd01 (approved):

jacobdgm:
> This might be too nit-picky, but the text "keyword search" is ever so slightly higher above the widget than the other fields.

Nice catch, and yes, I noticed this too - I think it also starts a tiny bit further left than the other labels. Both of these are because it's on a different row (this is to make sure it stays on its own line when the window gets really narrow). Maybe a front-end wizard can clean this up at some point, but for now, it's probably good enough.

PR #851, opened 2023-07-14
Write Tests for Chant Search View

jacobdgm:
This PR adds a bunch of tests for our Chant Search view. It already had some tests, but none to test the click-on-the-column-headers-to-reorder-results feature. So, now we have a thorough suite of tests to ensure this feature doesn't break when we refactor this view and the Chant Search MS View.

A task for early next week: I need to do the same thing for the ChantSearchMSViewTest. I think this will mostly be copy-pasting, so shouldn't take nearly as long.

Also in this PR, I stumbled across a commented-out import statement in a management command, and found some print statement / debugging apparatus that had accidentally been left in one of the tests. I got rid of them. Finally, the Chant Create view redirect behavior is meant to be different now, so that test is updated too.

### DIFF
--- a/django/cantusdb_project/main_app/tests/make_fakes.py
+++ b/django/cantusdb_project/main_app/tests/make_fakes.py
@@ -113,6 +113,7 @@ def make_fake_chant(
     c_sequence=None,
     cantus_id=None,
     feast=None,
+    mode=None,
     manuscript_full_text_std_spelling=None,
     incipit=None,
     manuscript_full_text_std_proofread=None,
@@ -142,6 +143,8 @@ def make_fake_chant(
         cantus_id = make_random_string(6, "0123456789")
     if feast is None:
         feast = make_fake_feast()
+    if mode is None:
+        mode = make_random_string(1, "0123456789*?")
     if manuscript_full_text_std_spelling is None:
         manuscript_full_text_std_spelling = faker.sentence()
     if incipit is None:
@@ -167,7 +170,7 @@ def make_fake_chant(
         position=position,
         cantus_id=cantus_id,
         feast=feast,
-        mode=make_random_string(1, "0123456789*?"),
+        mode=mode,
         differentia=differentia,
         finalis=make_random_string(1, "abcdefg"),
         extra=make_random_string(3, "0123456789"),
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -871,6 +871,487 @@ def test_keyword_search_contains(self):
...
+                "op": "contains",
+                "order": "siglum",
+                "sort": "desc",
+            },
+        )
+        descending_results = response_descending.context["chants"]
+        first_result_incipit = descending_results[1]["incipit"]
+        self.assertEqual(first_result_incipit, chant_1.incipit)
+        last_result_incipit = descending_results[0]["incipit"]
+        self.assertEqual(last_result_incipit, chant_2.incipit)
```
lucasmarchd01: I think it would be clearer if here (and in the other ordering tests) you still set the key in `first_result_incipit = descending_results[1]["incipit"]` to 0 instead of 1, otherwise the "first result" variable is set to the second (last) result. Then in the following line say `self.assertEqual(first_result_incipit, chant_2.incipit)`

When I read through the first time it looked like there wasn't any difference from the first block.

lucasmarchd01 (approved):

django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -140,8 +142,12 @@ def make_fake_chant(
         c_sequence = random.randint(1, MAX_SEQUENCE_NUMBER)
     if cantus_id is None:
         cantus_id = make_random_string(6, "0123456789")
+    if image_link is None:
+        image_link = faker.image_url()
     if feast is None:
         feast = make_fake_feast()
+    if mode is None:
+        mode = make_random_string(1, "0123456789*?")
```
dchiller, line 150: Just want to note that the mode field could have different values that these here and could be 2 characters long...that may not be material here, but since you are doing ordering it may!

django/cantusdb_project/main_app/tests/test_views.py
dchiller: Agreed. `first_result_incipit` and `last_result_incipit` variable names are confusing here.

dchiller (approved):

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: this was an oversight on my part - apologies, and thanks for catching this.

PR #855, opened 2023-07-18
Add tests to ChantSearchMSViewTest

jacobdgm:
This PR adds tests to ChantSearchMSViewTest, ensuring that when a user clicks on one of the column header links, the ordering/sorting of results works as expected.

These tests are a lightly-adapted version of the tests in #851, with a couple of medium-ish changes: I removed the order-by-siglum test (the link is there on the page, but all the chants will have the same siglum since they come from the same source), and we don't test that `search_bar` is maintained (you shouldn't be able to use a `search_bar` query parameter for this page).

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -1782,6 +1782,454 @@ class ChantSearchMSViewTest(TestCase):
         )
         self.assertEqual(len(response_contains.context["chants"]), 3)
 
+    def test_order_by_incipit(self):
+        source = make_fake_source(published=True)
+        chant_1 = make_fake_chant(source=source, incipit="higgledy")
+        chant_2 = make_fake_chant(source=source, incipit="piggledy")
+
+        search_term = "iggl"
+
+        response_ascending = self.client.get(
+            reverse("chant-search-ms", args=[source.id]),
+            {
+                "keyword": search_term,
+                "op": "contains",
+                "order": "incipit",
+                "sort": "asc",
+            },
+        )
+        ascending_results = response_ascending.context["chants"]
+        first_result_incipit = ascending_results[0]["incipit"]
+        self.assertEqual(first_result_incipit, chant_1.incipit)
+        last_result_incipit = ascending_results[1]["incipit"]
+        self.assertEqual(last_result_incipit, chant_2.incipit)
+
+        response_descending = self.client.get(
+            reverse("chant-search-ms", args=[source.id]),
+            {
+                "keyword": search_term,
+                "op": "contains",
+                "order": "incipit",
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1771,6 +1782,454 @@ def test_keyword_search_searching_all_fields(self):
...
+            response_asc = self.client.get(
+                reverse("chant-search-ms", args=[source.id]),
+                {
+                    "search_bar": search_term,
+                    "order": ordering,
+                    "sort": "asc",
+                },
+            )
+            html_asc = str(response_asc.content)
+            expected_substring = f"&order={ordering}&sort=desc"
```
lucasmarchd01, line 2225: should `expected_substring = f"&order={ordering}&sort=asc"` for the ascending response?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1771,6 +1782,454 @@ def test_keyword_search_searching_all_fields(self):
...
+            response_desc = self.client.get(
+                reverse("chant-search-ms", args=[source.id]),
+                {
+                    "search_bar": search_term,
+                    "order": ordering,
+                    "sort": "desc",
+                },
+            )
+            html_desc = str(response_desc.content)
+            expected_substring = f"&order={ordering}&sort=asc"
```
lucasmarchd01, line 2252: Same here, desc?

lucasmarchd01 (approved):

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: nope - the expected substring should be found in the link in the column header currently being ordered by in ascending sort. By clicking on this link, the sorting should switch _from_ ascending (which it currently is) _to_ descending.

lucasmarchd01: Ah, gothca.

jacobdgm: also, in hindsight, I'm regretting using the same query parameters as OldCantus on the ChantSearchView and ChantSearchMSView when I implemented them a few months ago - it makes much more sense for a column to be `sort`ed by a field, in ascending or descending `order`. The current implementation is not only backwards-compatible, but also straight-up backwards. Sigh...

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: nope - the column in question is currently sorted descending, so by clicking on the link, the sorting should flip from descending to ascending.

Let me add some comments clarifying this in another commit.

PR #856, opened 2023-07-18
Merge develop into staging, 2023 July 18 edition

jacobdgm:


django/cantusdb_project/main_app/views/views.py
```diff
@@ -654,6 +623,141 @@ def build_json_cid_dictionary(chant, request) -> dict:
...
+NODE_ID_CUTOFF = 1_000_000
+
+
+def json_node_export(request, id: int) -> JsonResponse:
+    """
+    returns all fields of the chant/sequence/source/indexer with the specified `id`
+    """
+
+    # all IDs above this value are created in NewCantus and thus could have conflicts between types.
+    # when data is migrated from OldCantus to NewCantus, (unpublished) dummy objects are created
```
ahankinson, line 705: Are these unpublished dummy objects kept around or deleted?

You can also just run a single SQL command once to set the starting ID in the new database... `ALTER SEQUENCE "the_table_sequence_id" RESTART WITH 100000000;`.

jacobdgm: They can be kept around, or deleted - they're unpublished (so only visible to logged-in users), and basically serve to do what your command does. They have a message saying they can be deleted once a source/chant/sequence with an ID greater than 1000000 has been created, so they shouldn't cause too much confusion.

But, @lucasmarchd01, based on this, while doing the final data sync, we can probably just remove the bit about creating the dummy objects if it makes things simpler and then run this command instead (n.b. with the right number of zeroes).

PR #857, opened 2023-07-18
Create custom 502 and 504 pages

### LINKED ISSUE #853: We should create custom 502 and 504 pages
We should create custom, nice-looking 502 and 504 (and possibly other 5xx - 502 and 504 are the only ones I've seen so far, though) pages for our nginx container to return in case requests take too long to process.

jacobdgm:
This PR fixes #853, creating basic 502 and 504 pages for nginx to serve in case NewCantus has a bad day.

This is what the pages look like in Chrome:
Screenshot 2023-07-18 at 3 29 39 PM
Screenshot 2023-07-18 at 3 29 57 PM

Are we happy with the error messages?


lucasmarchd01:
Is the reason we cant include `{% extends "base.html" %}` or `{% block content %}` like in the other error pages because of the nature of the error (bad server response) which wouldn't be a problem in 403/404 errors?

jacobdgm:
This is meant to be a page returned by Nginx, not by django - it's a last-resort page that nginx returns when django doesn't respond for some reason.

lucasmarchd01 (approved):

jacobdgm:
I pushed a couple of commits configuring this to automatically be set up upon running `docker-compose build` and `docker-compose up -d`.

@kqct: can you please look this over as soon as you can to make sure we've not done anything silly? Awaiting your approval before we merge.

Issue #800, opened 2023-06-28
We can improve the Admin area


jacobdgm:
During yesterday's video call with Debra, she went through a couple of normal editing/administrative tasks she's used to doing in OldCantus. As she did them in NewCantus, she had several suggestions as to how the Admin Area could be improved to make it easier to interact with.

General:

- [x] - it would be helpful to have required fields marked with asterisks
- [x] - ensure that on the various admin list pages, there's a search box at the top (the models that Debra noted were sources, provenances and feasts)
- [x] - We should make sure as many relevant columns as possible are included on the list pages for easy sorting/finding.

Users:

- [ ] - we should hide the checkbox that would allow Debra to turn a user into a superuser
- [x] - currently, the "staff status" checkbox is above the fields for assigning people to groups. We should move the "staff status" checkbox below the groups fields.
- [ ] - We should make the Full Name field required, and eventually hide the First Name and Last Name fields once we finish implementing #168.
- [x] - on the user list page, we should include a "full name" column (the `full_name` field is where indexers' names are kept, and it's hard to find an indexer if their full name is not displayed).
- [x] - On the user detail page, we should ensure the lists of sources (i.e. where the user can be added as an editor/proofreader/entered-melodies-er, etc.) are sorted alphabetically. It could be helpful for these to be filterable lists as well (if they're not already filterable lists)
- [x] - on the source list page, we should ensure the sources are sorted by sigla, and that there's a box at the top to search for sources.

It may be worth thinking about making a second admin area where we keep options like Users's "superuser" checkboxes, the "sites" area, etc. We could give this its own distinct URL (`/dev-admin/`?), but not include any links to it; its purpose would be to allow developers to change things as necessary like the site configuration.

annamorphism:
Debra specially requests a search for sources, to make it easier to give people permissions to edit:

"
http://206.12.93.196/admin/main_app/source/

Could we please have a search box for this list? Or if it's sorted in alpha order that would also be fine.

(I was also trying to go to "user" and add a source to someone that way, but those are also not in any particular order so it was difficult to find the source since we have so many. That's when I thought maybe I was to go to "source" and add a user.)"

jacobdgm:
> Could we please have a search box for this list?

I've added this and another item or two to the bottom of the list. Thanks for these suggestions/reminders!

lucasmarchd01:
> we should hide the checkbox that would allow Debra to turn a user into a superuser

[...]

PR #861, opened 2023-07-19
add search fields to admin model pages

lucasmarchd01:
Towards #800, we are adding search boxes to the top of several models (Provenance, Feast, Sequence) to make searching much easier.

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -37,7 +37,7 @@ class ChantAdmin(BaseModelAdmin):
 
 
 class FeastAdmin(BaseModelAdmin):
-    pass
+    search_fields = ("name", "feast_code")
 
 
 class GenreAdmin(BaseModelAdmin):
@@ -53,7 +53,7 @@ class OfficeAdmin(BaseModelAdmin):
 
 
 class ProvenanceAdmin(BaseModelAdmin):
-    pass
+    search_fields = ("name",)
 
 
 class RismSiglumAdmin(BaseModelAdmin):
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/admin.py
```diff
@@ -65,7 +65,10 @@ class SegmentAdmin(BaseModelAdmin):
 
 
 class SequenceAdmin(BaseModelAdmin):
+    search_fields = ("title", "incipit", "cantus_id",)
     exclude = EXCLUDE + ("c_sequence", "next_chant", "is_last_chant_in_feast")
+    list_display = ("incipit", "siglum", "genre")
```
jacobdgm: > `"siglum"`

- `list_display` dictates which columns are displayed, yes? Do we want to use `"source__siglum"` instead of just `"siglum"`? (I'm still a little confused as to why chants/sequences have a "siglum" field in OldCantus, and I don't think we have anything set up to automatically set the siglum of new chants created in NewCantus, so I've been avoiding using BaseChant.siglum...)

lucasmarchd01: In our chant sync script we have
```    
try:
        siglum = json_response["field_siglum_chant"]["und"][0]["value"]
except (KeyError, TypeError):
        siglum = None
```

And the taxonomy manager does show each siglum that corresponds to the chant:
image

However, if we are not using this field anywhere else in the code, it does make more sense to use source__siglum instead of siglum. (However, "source__siglum" doesn't behave as expected in the model admin, so we have to access it in a different way. Alternatively we could just use "source" which displays the full source __str__ representation.)

jacobdgm (approved):

jacobdgm:
Just noting, as I'm currently updating the staging and production servers with a few changes: this PR is not included in this round of develop->staging or staging->production.

PR #862, opened 2023-07-19
Optimize postgres configuration

jacobdgm:
This PR changes the Postgres configuration file, `postgresql.config`, so that it hopefully runs better on our staging and production servers. It also changes the docker setup so that this file ends up in the right place in the postgres container.

Requesting @kqct's review, to ensure all the Docker-related changes are sensible, and requesting @ahankinson's review to ensure all the Postgres-related changes are reasonable.

Testing locally, these changes appear to have a slight, but hardly significant significant, positive effect on query evaluation times. We expect it to have a larger effect when running on the DRAC servers, since the configuration is meant to match the specs of those VMs and not those of my local computer.

### DIFF
--- a/postgres/postgresql.conf
+++ b/postgres/postgresql.conf
@@ -1,0 +1,822 @@
+# -----------------------------
+# PostgreSQL configuration file
+# -----------------------------
+#
+# This file consists of lines of the form:
+#
+#   name = value
+#
+# (The "=" is optional.)  Whitespace may be used.  Comments are introduced with
+# "#" anywhere on a line.  The complete list of parameter names and allowed
+# values can be found in the PostgreSQL documentation.
+#
+# The commented-out settings shown in this file represent the default values.
+# Re-commenting a setting is NOT sufficient to revert it to the default value;
+# you need to reload the server.
+#
+# This file is read on server startup and when the server receives a SIGHUP
+# signal.  If you edit the file on a running system, you have to SIGHUP the
+# server for the changes to take effect, run "pg_ctl reload", or execute
+# "SELECT pg_reload_conf()".  Some parameters, which are marked below,
+# require a server shutdown and restart to take effect.
+#
+# Any parameter can also be given as a command-line option to the server, e.g.,
+# "postgres -c log_connections=on".  Some parameters can be changed at run time
+# with the "SET" SQL command.
+#
+# Memory units:  B  = bytes            Time units:  us  = microseconds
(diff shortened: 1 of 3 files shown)


postgres/postgresql.conf
```diff
@@ -0,0 +1,822 @@
...
+#ssl_passphrase_command_supports_reload = off
+
+
+#------------------------------------------------------------------------------
+# RESOURCE USAGE (except WAL)
+#------------------------------------------------------------------------------
+
+# - Memory -
+
+shared_buffers = 128MB			# min 128kB
```
ahankinson, line 155: This is not correct. You need to check these values against the PG config generator. I'll post what I get from that in a comment so it doesn't get lost here, but basically this setting, and the cache size settings are the most important ones.

So this should be *at least* 1GB, if not 2.

ahankinson:
Here's what I get from pgtune:

```
# DB Version: 14
# OS Type: linux
# DB Type: web
# Total Memory (RAM): 8 GB
# CPUs num: 8
# Connections num: 100
# Data Storage: hdd

max_connections = 100
shared_buffers = 2GB
effective_cache_size = 6GB
maintenance_work_mem = 512MB
checkpoint_completion_target = 0.9
wal_buffers = 16MB
default_statistics_target = 100
random_page_cost = 4
effective_io_concurrency = 2
work_mem = 5242kB
min_wal_size = 1GB
max_wal_size = 4GB
max_worker_processes = 8
max_parallel_workers_per_gather = 4
max_parallel_workers = 8
max_parallel_maintenance_workers = 4
```

postgres/postgresql.conf
jacobdgm: ah - thanks for catching this! I thought all the lines below what I copied from pgtune were commented out, but I see this overwrites the 2GB specified near the top of the file.

ahankinson: You should clean this file up. Any line that is commented out is anyway the default setting, so you don't lose anything by just deleting those lines.

jacobdgm:
> ```
> # DB Version: 14
> # OS Type: linux
> # DB Type: web
> # Total Memory (RAM): 8 GB
> # CPUs num: 8
> # Connections num: 100
> # Data Storage: hdd
> ```

Thank you for this - evidently, I guessed wrong for Data Storage, and didn't know what to use for Connections num. I'll use the values from your comment.

jacobdgm:
Github is being ornery - closing this; going to reopen and see if my most recently pushed changes will appear.

PR #866, opened 2023-07-20
Merge staging into production, 2023-07-20 edition

jacobdgm:
Changes involved:
- general
  - we now have custom, Cantus-themed 502 and 504 pages
  - postgres settings have been tweaked, which should mean most (all?) pages load faster and we get fewer 502/504 errors
- Chant Search page and Chant Search MS page
  - When window is wide, table of search results now spans the full width of the main body


jacobdgm:
Going to go ahead and merge this - the Chant Search pages both look right, and page feel like they are loading snappily.

PR #871, opened 2023-07-21
Update data migration sync scripts

lucasmarchd01:
This PR includes all of the updates that were made to the synchronization scripts to migrate the data from oldcantus to newcantus. We may never have to use these scripts again since we have now completed the final migration, but this will be the most up-to-date code for this task regardless.
- Handle Index errors and ObjectDoesNotExist errors
- Remove calls to "make dummy object" functions since we have now directly altered the table to start ID's for these objects at 1000000

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/sync_sequences.py
+++ b/django/cantusdb_project/main_app/management/commands/sync_sequences.py
@@ -233,7 +233,7 @@ def make_dummy_sequence() -> None:
             "This unpublished dummy sequence exists in order that all newly created "
             "sequences have IDs greater than 1,000,000 (which ensures that requests "
             "made to /node/<id> URLs can be redirected to their proper "
-            "sequence/sequence/article detail page). Once a sequence with an ID greater than "
+            "chant/sequence/article detail page). Once a sequence with an ID greater than "
             "1,000,000 has been created, this dummy sequence may be safely deleted."
         ),
     )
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/management/commands/sync_sequences.py
```diff
@@ -226,16 +226,18 @@ def make_dummy_sequence() -> None:
...
-    dummy_sequence = Sequence.objects.create(
+    Sequence.objects.create(
+        title="DUMMY",
         source=dummy_source,
         manuscript_full_text_std_spelling=(
             "This unpublished dummy sequence exists in order that all newly created "
             "sequences have IDs greater than 1,000,000 (which ensures that requests "
             "made to /node/<id> URLs can be redirected to their proper "
-            "chant/sequence/article detail page). Once a sequence with an ID greater than "
+            "sequence/sequence/article detail page). Once a sequence with an ID greater than "
```
jacobdgm: this doesn't look right

jacobdgm (commented):
Even after these changes, it appears these dummy objects are created, are they not?

lucasmarchd01:
> Even after these changes, it appears these dummy objects are created, are they not?

I left the functions there, but they are not actually being called anywhere so the dummy objects are not being created.

jacobdgm (approved):

jacobdgm:
Come to think of it, it will be good to have up-to-date syncing scripts, for reference in the future. So thanks for cleaning this up!

PR #881, opened 2023-07-24
Merge Staging into Production, 2023-07-24 edition

jacobdgm:
Pages/views that have been modified:

- admin area
  - [x] feast list: a search box has been added to the top of the page, allowing filtering by `name` and `feast_code`
  - [x] provenance list: a search box has been added to the top of the page, allowing filtering by `name`
  - [x] sequence list: a search box has been added to the top of the page, allowing filtering by `incipit`, `title` and `cantus_id`
- [x] Full Index page
  - has been optimized. Other than loading faster, there should be no changes to how it looks/behaves
- [x] all pages
  - SSHRC logo is now clickable, and takes you to the SSHRC website (previously, it wasn't a link)

Other changes:
- a restart policy has been added to `docker-compose.yml`. This means that if one of our servers goes down for some reason (as appears to have happened on staging over the weekend), it will start back up right away.
- the sync scripts have been updated to the version we used while doing the final sync. We shouldn't need to run these again, but in the future we may want to know what they did, so now they're up-to-date
- The project's CSRF settings have been updated, meaning users will be able to log in on Production (this change had already been applied as a hotfix last Friday)

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -196,6 +196,8 @@ INTERNAL_IPS = [
     "127.0.0.1",
 ]
 
+CSRF_TRUSTED_ORIGINS = ["https://cantusdatabase.org", "https://www.cantusdatabase.org"]
+
 if DEBUG:
     INSTALLED_APPS.append("debug_toolbar")
     # debug toolbar must be inserted as early in the middleware as possible
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -37,7 +37,7 @@ class ChantAdmin(BaseModelAdmin):
 
 
 class FeastAdmin(BaseModelAdmin):
-    pass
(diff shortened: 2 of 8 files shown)


JoyfulGen:
@jacobdgm just a heads-up, you still have test manuscripts on https://cantusdatabase.org/, like this one:

My favourite manuscript

I usually type "hello" in search boxes to see what happens if I search for something that won't exist... imagine my surprise when I actually got a result! I know you're probably aware of tests like these being on there, but my surprise at finding a chant called "Hi I'm a chant" was too good not to share.

jacobdgm:
Ha! After consulting with @annamorphism, I actually just deleted all the test sources from Production earlier this morning (but evidently after you found this).

So long as you only saw this while you were logged-in and not while you were logged-out, @JoyfulGen, then everything is working as it should.

jacobdgm:
note: we're waiting to put these changes up on the Production machine until a couple of changes applied as hotfixes can be properly integrated into the repo.

PR #885, opened 2023-07-25
Ensure Article objects have date_created set when creating new Object

### LINKED ISSUE #794: It's possible to create an article with no `date_created`
Trying to figure out #793, I realized that it's currently possible to save an article in the Admin area without supplying a `date_created`. If this happens, all the article pages/flat pages will lead to errors, when `strftime` tries to format a `None`.

I've spent a while adding various combinations of `default=datetime.now`, `auto_now_add=True`, `blank=False` and `null=False`, and none have achieved the desired result of cleanly preventing an article from being saved without a `date_created`. When `blank` and `null` are both `False`, trying to save an article without specifying a `date_created` leads to a scary `IntegrityError` traceback, rather than the desired popup gently informing the user that they've left out a bit of required information (I'm under the impression that `blank=False` should accomplish this)

While this is not an _enormous_ issue - as long as all new articles have a date and time specified, nothing will break - the current state of affairs makes it possible for a small slip-up by a user to prevent the homepage from loading for everyone.

lucasmarchd01:
Fixes #794. To ensure that the `date_created` field is set when an Article is created, we can use the `auto_now_add` field which will automatically set the field to now only when the object is first created. When a user creates an Article from the admin area, this field is now displayed as "read only". Additionally, write tests for the date_created and date_updated fields to ensure they are always populated.
image

Finally, in this PR we are adding migration files to our code and removing them from .gitignore so that we can keep track of changes to our database.


ahankinson:
Did this not trigger the addition of a database migration file? I don't see the migration in the PR

jacobdgm:
> Did this not trigger the addition of a database migration file? I don't see the migration in the PR

For as long as I've been working on the project, we've been git-ignoring migrations. I don't know why this decision was made, as I understand migrations are kind of designed to be included in the git history. Maybe now is the time to begin including them?

(A further complication: our [documentation for setting up the project](https://github.com/DDMAL/CantusDB/wiki/Getting-Started-with-Development#loading-initial-data-two-methods) includes an instruction, "If you're not starting from a fresh clone, delete the migrations folder in every app: articles, main_app, users." Maybe this will be less of an issue now that we've finished syncing data from OldCantus and we won't need to be shuttling copies of the database from machine to machine. But we'll need to figure out what to do with these instructions)

ahankinson:
> For as long as I've been working on the project, we've been git-ignoring migrations. I don't know why this decision was made, as I understand migrations are kind of designed to be included in the git history. Maybe now is the time to begin including them?

Without migrations you cannot make changes to the database structure. Django keeps track of the migrations that it has applied, and the purpose of migrations is to change the database without requiring a complete database rework, and to keep track of the changes in the structure of the database in code. If you haven't had problems yet, you will most certainly in the future.

lucasmarchd01:
@ahankinson Thank you for the suggestion! We have now added the migration files for our apps so we can keep track of them.

django/cantusdb_project/articles/tests/test_articles.py
```diff
@@ -61,3 +60,32 @@ def test_context(self):
...
+        article.save()
+
+        updated_article = Article.objects.get(id=article.id)
+        self.assertNotEqual(initial_date_updated, updated_article.date_updated)
+
+        # Make sure the time difference is reasonable (e.g., less than a minute)
+        time_difference = updated_article.date_updated - initial_date_updated
+        self.assertLess(time_difference, timedelta(minutes=1))
+
+    def test_null_date_created(self):
```
jacobdgm, line 85: this would be a good place for a comment

ahankinson:
I'd be interested to know if your staging / production environments take the new migrations without complaining, so please let me know.

jacobdgm (approved):

PR #886, opened 2023-07-25
Configure site to use certbot to work with https

jacobdgm:
@kqct and I made a couple of changes to some configuration files to get https working properly on Production last week. I updated the documentation, but the instructions ended up being quite long and fiddly. Hoping that by adding these changes to the repo, we can simplify the instructions.

@kqct, can you look this over to ensure we're not doing anything foolish here?

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -1,6 +1,23 @@
 server {
-
     listen 80;
+
+    server_tokens off;
+
+    location ^~ /.well-known/acme-challenge/ {
+        root /var/www/certbot;
+    }
+
+    location / {
+        return 301 https://$host$request_uri;
+    }
+}
+
+server {
+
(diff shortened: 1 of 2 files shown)


jacobdgm:
N.B. I tried building this locally, and both the certbot and the nginx containers exited with error codes (due, it seems, to missing certificates). I'm guessing that in order to run this on staging or locally, we'll need to comment out lines related to the certbot container and the certificates. @kqct, does this make sense?

jacobdgm:
I've updated the documentation to comment out specific lines when setting up locally or on production. Going to merge this, put it up on staging, and test that my documentation actually works.

kqct:
@jacobdgm, the changes on lines 9-20 of `config/nginx/conf.d/cantusdb.conf` are what's probably breaking your local build -- they require Nginx to already be active. The changes on lines 6-8 are also involved in creating certificates, but you don't necessarily have to remove them in order for the container to build.

Issue #883, opened 2023-07-24
saving creation date


annamorphism:
We would like to preserve the creation date of sources in Cantus before the old site goes down. Currently they all show up in NewCantus as newly created from the date of import. (This information loss already happened at the last Cantus move in 2014, so it would be nice not to also lose the dates for everything created since then...)
I think this information my only show for admins, but Debra requested it. 

Would it be possible to import the old created date, as happened for "created by"?
If not, it would be useful to at least create a file containing the source ID, Rism Siglum, and creation date, so somebody can at least have it on file after the old site goes down in November.

jacobdgm:
looking into this

jacobdgm:
It looks like we have the information we need to do this!

lucasmarchd01:
@jacobdgm okay to close? If everything looks in order on production?

PR #889, opened 2023-07-26
Add script to sync source creation date

lucasmarchd01:
Haven't run this script yet, but I am making this PR so we have the code for future reference.

This will fix issue #883

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
+++ b/django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
@@ -1,0 +1,87 @@
+from django.core.management.base import BaseCommand
+from main_app.models import (
+    Source,
+)
+import requests, json
+import datetime
+from django.utils import timezone
+
+
+def get_source_list():
+    source_list = []
+    file = open("source_list.txt", "r")
+    for line in file:
+        line = line.strip("\n")
+        source_list.append(line)
+    file.close()
+    return source_list
+
+
+def convert_epoch_to_date_time(epoch):
+    try:
+        epoch_time = float(epoch)
+        datetime_str = datetime.datetime.fromtimestamp(epoch_time)
+        datetime_obj = timezone.make_aware(datetime_str, timezone.utc)
+    except (ValueError, TypeError) as e:
+        print(f"Error converting epoch time: {e}")
+        return None
+    return datetime_obj
+
+
+def update_date_created(source_id):
+    url = f"http://cantus.uwaterloo.ca/json-node/{source_id}"
+    response = requests.get(url)
+    json_response = json.loads(response.content)
+    try:
+        date_created_epoch = json_response["created"]
+    except (KeyError, TypeError) as e:
+        print(f"Exception for source {source_id}: {e}")
+        date_created_epoch = None
+    if date_created_epoch is not None:
+        datetime_str = convert_epoch_to_date_time(date_created_epoch)
+        if datetime_str is not None:
+            Source.objects.filter(id=source_id).update(date_created=datetime_str)
+            source_obj = Source.objects.get(id=source_id)
+        return source_obj
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "id",
+            type=str,
+            help="update one source (<source_id>)) or update all sources ('all')",
+        )
+
+    def handle(self, *args, **options):
+        # before running this, run sync_indexers first
+        id = options["id"]
+        if id == "all":
+            all_sources = get_source_list()
+            total = len(all_sources)
+            count = 0
+            for source_id in all_sources:
+                try:
+                    source = Source.objects.get(id=source_id)
+                except Source.DoesNotExist:
+                    print(f"Source with ID {source_id} does not exist")
+                    source = None
+                if source is not None:
+                    print(
+                        f"old date created for source {source_id}: {source.date_created}"
+                    )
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+        date_created_epoch = json_response["created"]
+    except (KeyError, TypeError) as e:
+        print(f"Exception for source {source_id}: {e}")
+        date_created_epoch = None
+    if date_created_epoch is not None:
+        datetime_str = convert_epoch_to_date_time(date_created_epoch)
+        if datetime_str is not None:
+            Source.objects.filter(id=source_id).update(date_created=datetime_str)
+            source_obj = Source.objects.get(id=source_id)
+        return source_obj
```
jacobdgm: I don't understand why we're returning `source_obj` here

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "id",
+            type=str,
+            help="update one source (<source_id>)) or update all sources ('all')",
```
jacobdgm: unmatched `)`, minor, feel free to ignore this comment

jacobdgm (approved):
Looks good in general. Feel free to make changes (or not) based on my two comments.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+from main_app.models import (
+    Source,
+)
+import requests, json
+import datetime
+from django.utils import timezone
+
+
+def get_source_list():
+    source_list = []
```
ahankinson: The body of this function can be simplified:

```py
with open("source_list.txt", "r") as myfile:
    source_list = myfile.read().splitlines()

return source_list
```

Using a context manager (`with ...`) on a file object will automatically close the file when it goes out of scope.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+        if id == "all":
+            all_sources = get_source_list()
+            total = len(all_sources)
+            count = 0
+            for source_id in all_sources:
+                try:
+                    source = Source.objects.get(id=source_id)
+                except Source.DoesNotExist:
+                    print(f"Source with ID {source_id} does not exist")
+                    source = None
```
ahankinson: There's no point in this loop going, so just do `continue` here. That will then loop to the next source.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+        print(f"Error converting epoch time: {e}")
+        return None
+    return datetime_obj
+
+
+def update_date_created(source_id):
+    url = f"http://cantus.uwaterloo.ca/json-node/{source_id}"
+    response = requests.get(url)
+    json_response = json.loads(response.content)
+    try:
```
ahankinson, line 26: Again, simplify and return early. 

```py
if "created" not in json_response:
    return None
```

Adding a type annotation to this function will help you to sort out what this function actually returns; Even though you don't ever explicitly return `None` it will implicitly return None if the `date_create_epoch` is None. So adding `-> Optional[Source]` is definitely helpful.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+        date_created_epoch = json_response["created"]
+    except (KeyError, TypeError) as e:
```
ahankinson: At this point, if you've checked that "created" is not in `json_response` then you don't need the `try/except ...`.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
```diff
@@ -0,0 +1,87 @@
...
+    def handle(self, *args, **options):
+        # before running this, run sync_indexers first
+        id = options["id"]
+        if id == "all":
+            all_sources = get_source_list()
+            total = len(all_sources)
+            count = 0
+            for source_id in all_sources:
+                try:
+                    source = Source.objects.get(id=source_id)
```
ahankinson: I presume all sources you want to update are already in the database, so why not loop over `Source.objects.all()`? Why read it from an external file? Then you don't need to check if the source doesn't exist, and you can pass the actual source object around your functions so that you're not doing at least three database lookups for every source!

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
lucasmarchd01: Right, type annotations would definitely help here as well as being more explicit with what is being returned.

django/cantusdb_project/main_app/management/commands/sync_source_creation_date.py
lucasmarchd01: The way I implemented it was just the first way I thought of since it would work similarly to other scripts in our management commands. But it definitely makes more sense to do it this way.

PR #904, opened 2023-07-31
Remove `<meta content="noindex">` from templates

### LINKED ISSUE #466: Remove `<meta name="robots" content="noindex">` from production branch/server
Assigning a label of `highpriority` not because it is urgent, but because this needs to happen before/when NewCantus replaces OldCantus. We want google/etc. to index the website once the handover occurs, so we'll need to remove `<meta name="robots" content="noindex">` from both [`main_app/templates/full_index.html`](https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/main_app/templates/full_index.html) and [`templates/base.html`](https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/templates/base.html).

jacobdgm:
Fixes #466. This is a good thing, because we are currently instructing Google/etc. not to crawl NewCantus Production.

### DIFF
--- a/django/cantusdb_project/main_app/templates/full_index.html
+++ b/django/cantusdb_project/main_app/templates/full_index.html
@@ -2,8 +2,6 @@
 <html>
 
 <head>
-    <!-- prevent search engines from indexing the staging website, remove for production -->
-    <meta name="robots" content="noindex">
     <!-- Bootstrap CSS -->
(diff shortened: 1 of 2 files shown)


lucasmarchd01 (approved):

ahankinson:
Someone (perhaps Debra) should set up an account at http://search.google.com/search-console to track crawling.

PR #908, opened 2023-08-01
content overview: add tabs for models and display all objects ordered by date created

### LINKED ISSUE #896: Content-Overview page - we should paginate results
In an email from Debra,

> I was wondering if I could get an option to see more than the latest 50 updated items.
> https://cantusdatabase.org/content-overview/
> 
> (Or is what is showing now all that has been updated on NewCantus, and if the page gets longer there will be an option to see 1-50, 50-100 items etc.?  - or however the view might look)

lucasmarchd01:
In this PR, modifications are made to the content-overview page. Previously, this page would only display the last 50 updated objects in our database. The way this information was fetched was inefficient, making it impossible to scale to use all the objects in our database without affecting the page loading time. The solution implements tabs at the top of the page that represent the relevant models in our database. When the user clicks on one of these tabs, we will display all of the objects corresponding to this model ordered by "-date_created". The user can now paginate through all of these objects with 100 objects per page. Fixes #896 

image

### DIFF
--- a/django/cantusdb_project/main_app/templates/content_overview.html
+++ b/django/cantusdb_project/main_app/templates/content_overview.html
@@ -5,79 +5,98 @@
 
 <div class="mr-3 p-3 col-md-12 mx-auto bg-white rounded">
     <h3>Content Overview</h3>
-    <table class="table table-sm small table-bordered">
-        <small>Displaying the <b>50</b> most recently updated objects</small>
-        <div style="height:10px;"></div>
-        <thead>
-            <tr>
-                <th scope="col" class="text-wrap" style="text-align:center">Title / Manuscript Full Text / Name</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Type</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Creation Date</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Creator</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Last Updated Date</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Last Updated By</th>
-                <th scope="col" class="text-wrap" style="text-align:center">Operations</th>
-            </tr>
-        </thead>
-        <tbody>
-            {% for object in objects %}
+
+    <div class="row justify-content-center">
+        {% for model_name in models %}
+            <div class="col-auto">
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/views/views.py
```diff
@@ -794,30 +794,39 @@ def project_manager_check(user):
 def content_overview(request):
     objects = []
     models = [
-        Century,
+        Source,
```
jacobdgm, line 797: Are we sure we don't want to include Century in this list?

django/cantusdb_project/main_app/templates/content_overview.html
```diff
@@ -5,79 +5,98 @@
...
+                                    {{ object.created_by }}
+                                </a>
                             {% endif %}
                         </td>
-                    {% elif object.full_name %}
+                        <td class="text-wrap" style="text-align:center">{{ object.date_updated|date:'Y-m-d H:i' }}</td>
                         <td class="text-wrap" style="text-align:center">
-                            <a href="{{ object.get_absolute_url }}"><b>{{ object.full_name }}</b></a>
+                            {% if object.last_updated_by is None %}
+                                {{ object.last_updated_by }}
```
jacobdgm, line 78: Maybe it would be clearer to just put `None` rather than `{{ object.last_updated_by }}` (since we already know it's `None`)

jacobdgm (changes requested):
This looks good! I left one comment to check that `Century` models haven't been overlooked, and one suggesting a possible simplification to the code.

One other thing: do any tests need to be updated to match these changes? If not, maybe now is a good time to write a good set of tests for this view.

django/cantusdb_project/main_app/views/views.py
lucasmarchd01: Century is still included, I just changed the ordering of this list and Century was moved further down.

django/cantusdb_project/main_app/templates/content_overview.html
lucasmarchd01: Makes sense!

lucasmarchd01:
> One other thing: do any tests need to be updated to match these changes? If not, maybe now is a good time to write a 
> good set of tests for this view.

Thanks for the suggestion, will do.

jacobdgm (approved):

PR #913, opened 2023-08-01
admin: increase source title widget length

### LINKED ISSUE #907: Admin Chant Edit page: we should make "title" field longer
In an email from Debra:

> https://cantusdatabase.org/admin/main_app/source/123611/change/
For a page like this one, could I please have a longer box for the source title? The width of the "provenance note" field would be about right, I think, so that I can see the whole title without having to scroll.

lucasmarchd01:
Fixes #907 by updating the widget of the title field in the source admin form.


lucasmarchd01:
Note: I wasn't sure what Debra meant by "provenance note" field. I can't seem to find that anywhere.

jacobdgm (approved):

Issue #879, opened 2023-07-24
Make improvements to the Users admin page


lucasmarchd01:
There are a few things that we should do to improve the admin user page to make it more understandable.

- [x] We should display the full name on the list page
- [ ] We should move the users with is_indexer=True to a new tab under Users (probably called Indexers) that will hide them from the regular users list page
- [x] For indexers, we should display their full name instead of the fake email

jacobdgm:
> We should display the full name on the list page

This is an easy "yes, let's do this"

> We should move the users with is_indexer=True to a new tab under Users (probably called Indexers) that will hide them from the regular users list page

I know I was the one to suggest this, but I'm not 100% sure that this is the right thing to do. @annamorphism, do you have any feelings about this one? Should we check with Debra?

> For indexers, we should display their full name instead of the fake email

If indexers are on a separate page, then yes; otherwise, I'm not sure that hiding the email address is the right choice. This one depends on point number 2.

jacobdgm:
the remaining item in the checklist above feels complicated, and doesn't strike me as necessary. setting this to priority: low.

PR #930, opened 2023-08-03
Improve the user admin area

lucasmarchd01:
This PR makes some changes to the user admin area:
- ad "id" field to search and list display (source admin)
- add custom change user form to display email required widget
- order sources assigned to user alphabetically by siglum
- add "is_indexer", "is_superuser", "is_staff" to list filter
- add indexer to user edit and user add form
- order users in list page by full name

Towards #800, #879

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -878,3 +878,21 @@ class AdminSourceForm(forms.ModelForm):
     complete_inventory = forms.ChoiceField(
         choices=TRUE_FALSE_CHOICES_INVEN, required=False
     )
+
+
+class AdminUserForm(forms.ModelForm):
+    class Meta:
+        model = get_user_model()
+        fields = "__all__"
+
+    first_name = forms.CharField(
+        required=False,
+        help_text="We have full name, first name, and last name defined in our User model which can be seen "
+        "in the admin area, but we are only using full name to represent users.",
+    )
+
+    last_name = forms.CharField(
+        required=False,
+        help_text="We have full name, first name, and last name defined in our User model which can be seen "
+        "in the admin area, but we are only using full name to represent users.",
+    )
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/forms.py
```diff
@@ -879,3 +880,17 @@ class Meta:
...
+        model = get_user_model()
+        fields = "__all__"
+
+    email = forms.CharField(
+        required=True,
+        widget=AdminTextInputWidget,
+    )
+    email.widget.attrs.update({"style": "width: 300px;"})
+
+    password = ReadOnlyPasswordHashField()
```
jacobdgm: what does this do?

django/cantusdb_project/users/admin.py
```diff
@@ -47,16 +55,19 @@ class UserAdmin(BaseUserAdmin):
                     "institution",
                     ("city", "country"),
                     "website",
-                )
+                ),
+                "description": "We have full name, first name, and last name defined in our User model which can be seen "
+                "in the admin area, but we are only using full name to represent users.",
```
jacobdgm: "...represent users _on the main site_"

jacobdgm: Perhaps something along the lines of "you can enter a user's first and last name, but these are only ever displayed in the Admin area - on the main site, users' full names are always used rather than their first and last name" would be clearer?

django/cantusdb_project/main_app/forms.py
lucasmarchd01: When I initially set up the user admin change form, the user's passwords would be displayed as text (hashed, but editable and unhidden). Using the ReadOnlyPasswordHashField maintains the use of the hashed and hidden widget instead. However, I notice now that it is missing the same help text that takes the user to the password change form, so I added it here as well.

jacobdgm (approved):

Issue #870, opened 2023-07-21
We need to set up cron job to renew https certificate from time to time


jacobdgm:

jacobdgm:
@kqct, I've lost the command we need to use in the cron job - can you add it here?

kqct:
`docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ --dry-run -d cantusdatabase.org`

note: the -d flag specifies domains, you can have multiple

also, i'd recommend updating your docker install to use docker compose V2 (`docker compose` instead of `docker-compose`

jacobdgm:
@timothydereuse and I spent a while looking at this this morning.

we noticed a likely problem with this line in `cron/cron.txt`:

```
50      4       *       *       0       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org -d www.cantusdatabase.org --renew-hook "nginx -s reload"
```

Since this command is run in the certbot container, there's no way that `nginx -s reload` will be able to run successfully, since that command needs to run in the _nginx_ container. There are two possible ways this is going wrong:
- the certificate is being updated successfully, but it isn't being displayed by Nginx because it isn't being reloaded
- since the `renew-hook` command is failing, the `certbot` command is rolling back to the previous version of the certificate (i.e., the hook itself is preventing the certbot command from updating our certificates)

I'm planning on updating this line to `/usr/local/bin/docker-compose run --rm certbot certonly --force-renew --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org -d www.cantusdatabase.org && /usr/local/bin/docker-compose run nginx nginx -s reload"` on Production, and setting it to run a few minutes after I update it. We should be able to check that the certificate was updated following the run of the command. If it worked, I'll open a PR with this updated command (without `--force-renew`).

ahankinson:
I’m not really a shell scripter but I think you probably want a && to separate the commands so that the second one only runs if the first one was successful?

jacobdgm:
That sounds like a good idea! (Although I don't think there's any harm done if we tell nginx to reload even if the first command returns an error code. But yes, `&&` is certainly closer to what we're aiming to do than `;`. Thank you!)

dchiller:
Further investigations into this: 
- The constantly restarting container is not properly configured and also should not have a restart policy of always! Since all we theoretically want to do is run this container periodically (eg. weekly).
[...]

Issue #740, opened 2023-06-14
We need to automate populate_next_chant, populate_is_last_chant_in_feast commands


jacobdgm:
We have a handful of management commands - `populate_is_last_chant_in_feast`, `populate_next_chant_fields`, perhaps one or two more - that serve to optimize bits of CantusDB's operation. 

**These scripts are meant to be run from time to time by a cron job. We haven't yet set up said cron job. At some point, we need to do this.**

It's worth noting that the fields these commands change are mostly used to make suggestions to users as they are creating new chants - it's not important that the details of these suggestions (e.g. precisely how many times chant B follows chant A in the manuscript) are exactly right - so as long as these fields are populated with the correct values for _many_ of the objects in the database, this functionality will work fine. So, when we automate these, we don't need to run them very often. And if it takes a while to set them up, it's fine, so long as they've been run once following the synchronization of data from OldCantus to NewCantus.

Issue #877, opened 2023-07-24
Cantus IDs sometimes get merged; Cantus Index has an API to show which Cantus ID have been merged


jacobdgm:
In an email from Jan:

>  In our CI form, we have the option to set Cantus IDs that need to be renamed across all databases. https://cantusindex.org/merged 
> This renaming process can be done immediately upon submitting the CI form (after calling some URL in CD), or alternatively, you can set up a regular cron job to periodically check JSON https://cantusindex.org/json-merged-chants and rename the Cantus IDs in CD. I'm not certain which approach you would prefer.

We should probably check in with Debra first, but this is a thing we should probably implement before too long.

jacobdgm:
Sounds like Debra wants/needs this updated soon. Increasing the priority.

jacobdgm:
A couple of weeks ago, I looked into the json-merged-chants API and https://cantusindex.org/merged and found some unusual things, including:
- duplicate mappings: for example, old a00556 maps to new a00555a twice.
- places where one Cantus ID maps to itself: for example, 008276:06
- places where one Cantus ID maps to multiple new Cantus IDs: for example, old 008276f maps to both 008276:06 and 008276:07

I asked Jan and Debra about this, and based on the response from Jan, it seems that the API lists a record of all merges that ever happened on CantusIndex. In instances where one old Cantus ID maps to multiple updated Cantus IDs, it seems that one merge occurred, then users continued to add chants with the old Cantus ID, and then a second merge occurred changing these new chants to a different updated Cantus ID.

Jan suggested that whenever we run whatever script we come up with, we keep track of the index/id of the last mapping we performed, and only apply mappings that follow this one.

I'll need time to investigate further and think about good approaches to handling this; maybe we can discuss it at lab meeting.

jacobdgm:
Another thought. Jan suggested that he could add the ID of the merge event in the API. It might make more sense to ask that a timestamp be attached to each record. That way, if we run a cron job every day (say), we can filter the JSON objects by checking their timestamp is after the last time we ran the cron job. And when we do an initial run of this script, we can only apply updates that occurred since we did the final data migration from OldCantus to New.

jacobdgm:
Update, based on the Zoom call with Debra we just finished:

[...]

PR #933, opened 2023-08-03
Create postgres backup scripts and set up cron jobs

### LINKED ISSUE #395: Periodically save a copy of database in case of disaster
Before NewCantus steps up to replace OldCantus, we should set up something to save a copy of the database at regular intervals, just in case disaster strikes and someone deletes all the chants in the database / etc.

jackyyzhang03:
Resolves #395, #870, #740. 

Created database backup scripts to be run outside container through cron. The scripts are currently set up to keep 7 daily backups, 8 weekly backups, and 12 monthly backups.

Runs `populate_next_chant_fields` and `populate_is_last_chant_in_feast` management commands monthly.

Updates certificates every 6 months.

### DIFF
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -1,0 +1,11 @@
+######################
+# CantusDB Cron Jobs #
+######################
+
+# Note: This is set up to run on the production server. If you want to run it on your local machine, you will need to change the paths.
+
+# min   hour    day     month   weekday command
+0       4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/daily.sh
+0       4       *       *       0       bash /home/ubuntu/code/CantusDB/cron/postgres/weekly.sh
+0       4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
+0       4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -10,4 +10,4 @@
 20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
 30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
 40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
-50      4       1       */6     *       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ --dry-run -d cantusdatabase.org
+50      4       1       */6     *       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -5,10 +5,10 @@
 # Note: This is set up to run on the production server. If you want to run it on your local machine, you will need to change the paths.
 
 # min   hour    day     month   weekday command
-0       4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/daily.sh
-0       4       *       *       0       bash /home/ubuntu/code/CantusDB/cron/postgres/weekly.sh
-0       4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
-0       4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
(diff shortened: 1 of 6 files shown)


postgres/scripts/monthly.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin.bash
+
+# This script creates a compressed SQL dump of the database in the /var/lib/postgresql/backups/monthly directory.
+# It then deletes any montly backups older than 7 days.
```
jacobdgm: this comment is incorrect - it deletes any monthly backups older than a year.

cron/postgres/monthly.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin.bash
+
+# This script creates a compressed SQL dump of the database in the /home/ubuntu/backups/postgres/montly directory.
```
jacobdgm: Here, and elsewhere in this file, "monthly" rather than "montly"

jacobdgm (changes requested):

jacobdgm:
Also, would it make sense to create yearly backups that we keep indefinitely?

I'm also thinking of where certbot things will fit into this: does it make sense to address #870 within this PR, or to merge this PR and create a new one for certbot? i understand this needs to be run a little bit more often than yearly (something like every 9 months?)

jackyyzhang03:
>Also, would it make sense to create yearly backups that we keep indefinitely?

I added another job to keep yearly backups for a decade.

>I'm also thinking of where certbot things will fit into this

Not sure, though if you think it is a good idea, we can do it in this PR along with #740 and maybe #877?

cron/cron.txt
```diff
@@ -0,0 +1,14 @@
...
+
+# Note: This is set up to run on the production server. If you want to run it on your local machine, you will need to change the paths.
+
+# min   hour    day     month   weekday command
+0       4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/daily.sh
+5       4       *       *       0       bash /home/ubuntu/code/CantusDB/cron/postgres/weekly.sh
+10      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
+15      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
+20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
+25      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields
```
jacobdgm: Good thought to ensure these don't all run at the same time! Can we run `populate_next_chant_fields` before `populate_is_last_chant_in_feast` (we need up-to-date data on which chants follow each other before we can accurately calculate which chants end feasts)?

jacobdgm (changes requested):
This most recent round of changes looks quite good! I left just one comment with a requested change. Thanks for taking the time to set this up!

cron/cron.txt
```diff
@@ -0,0 +1,13 @@
...
+######################
+
+# Note: This is set up to run on the production server. If you want to run it on your local machine, you will need to change the paths.
+
+# min   hour    day     month   weekday command
+0       4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/daily.sh
+10      4       *       *       0       bash /home/ubuntu/code/CantusDB/cron/postgres/weekly.sh
+20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
+30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
+40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields
```
jacobdgm: This appears to be back to calculating `is_last_chant_in_feast` before calculating `next_chant`? `is_last_chant_in_feast` depends on `next_chant`.

jacobdgm (approved):
To me, this looks like it does what we need it to! I'll wait before merging until @kqct has checked this over to verify that the certbot setup is correct.

meow10811:
The certbot setup looks correct, but I'd caution that SSL certificates expire in 90 days -- maybe run the renewal script every 2 months instead of every 6?

jacobdgm:
> maybe run the renewal script every 2 months instead of every 6?

@jackyyzhang03

kqct (approved):

ahankinson:
Just a note to say that you should run the certbot script much more frequently than that. You can run it every night, or at least once a week. If the certificate is not due for renewal then it will bail and nothing will happen, but if you only run it every two months and some transient error happens (certbot servers are down for maintenance; your servers have a network glitch) then you've lost your window and your certificate will expire.

There's no downside to running it daily or weekly.

jacobdgm:
Thanks, @ahankinson - done! #956

PR #942, opened 2023-08-07
Case insensitive ordering by feasts name

### LINKED ISSUE #894: List of Genres: alphabetization should be case-insensitive
The Lists of Feasts (this includes the list of all feasts as well as only the temporale and only the sanctorale) are ordered by alphabetic order of feast name, except that capital letters are currently taking precedence; a capital "Z" comes before a lowercase "a" and that muddles the system somewhat. Behold:

NewCantus temporale list

You can see that the alphabet thing is solid until Ad Processionem, after which the third letter becomes lowercase (Ad aquam benedictio) and the alphabet starts all over again. Maybe this is preferable, but OldCantus doesn't take capitalization into account, so I thought I'd mention it.

This is the same list on OldCantus, for reference: 

OldCantus temporale list

jackyyzhang03:
Resolves #894.

Use case insensitive ordering by name in feasts list.

### DIFF
--- a/django/cantusdb_project/main_app/views/feast.py
+++ b/django/cantusdb_project/main_app/views/feast.py
@@ -1,4 +1,5 @@
 from django.views.generic import DetailView, ListView
+from django.db.models.functions import Lower
 from main_app.models import Feast, Source
 from extra_views import SearchableListMixin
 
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/feast.py
```diff
@@ -96,7 +97,8 @@ def get_ordering(self):
         # default to ordering by name if given anything else
         if ordering not in ["name", "feast_code"]:
             ordering = "name"
-        return ordering
+        # case insensitive ordering by name
+        return (Lower(ordering) if ordering == "name" else ordering, )
```
jacobdgm: curious about the `,` at the end of `ordering,` - does `get_ordering` need to return a tuple? If so, maybe a type annotation on `get_ordering` would clarify things, as well as wrapping `ordering,` in parentheses: `return (... else (ordering,))`.

django/cantusdb_project/main_app/views/feast.py
```diff
@@ -1,4 +1,5 @@
 from django.views.generic import DetailView, ListView
+from django.db.models.functions import Lower
```
jacobdgm, line 2: I don't think I would have found this! If it does what I think it does, this is excellent!

jacobdgm (approved):

PR #943, opened 2023-08-07
Use dropdown for Office/Mass field in chant search

### LINKED ISSUE #531: Chant search page - why is Office/Mass field not a <select> dropdown?
Screen Shot 2023-01-27 at 4 25 14 PM

We only have 17 offices/etc. - why is "Office/Mass" not a dropdown, like "Genre" is? The current implementation works when a user enters a value like "L" or "M", but no results are returned when a user enters "Lauds", which is a problem.

Making it a dropdown may make optimizing search times easier as well.

(There's no point of comparison with OldCantus - it doesn't have an "Office/Mass" field on its search page.)

jackyyzhang03:
Resolves #531.


jacobdgm (approved):
This is very good - thanks for setting this all up!

Issue #486, opened 2022-12-21
We need to create Delete Views for Source, Sequence and Chant


annamorphism:
Conclusions from the comments below:
> we should add source-delete, chant-delete and sequence-delete views. People should not see links to these views, and be blocked from accessing these views, unless they are a `project manager`, or if the source/chant/sequence in question was `created_by` the current user.

Original post:
While investigating #481 I checked out how deleting items works in OldCantus for mere "contributors". It seems they can currently delete any chants OR sources that they created themselves, which corresponds to those items that they have edit access for. Basically if  https://cantusdatabase.org/node/711349/edit is available to them (not 403), then https://cantusdatabase.org/node/711349/delete gives the option of deleting it, and if not then not. (Haven't checked yet if this also holds for "editors", "proofreaders" etc but I don't know why it wouldn't.)
<img width="792" alt="image" src="https://user-images.githubusercontent.com/67451875/208976160-6558fffd-96d7-49ce-ae31-747e4a516938.png">

<img width="1085" alt="image" src="https://user-images.githubusercontent.com/67451875/208976107-bb3ab673-1686-4270-821f-f94ba9be93e3.png">

It seems like in NewCantus the equivalent option to delete chants is done through chant-delete, but there is no equivalent for deleting sources. Wondering if that is intentional. 
<img width="664" alt="image" src="https://user-images.githubusercontent.com/67451875/208976738-71ffd22e-efff-4171-bdaa-557a4cd441d1.png">

jacobdgm:
Whoa - I've never seen the "Delete" button on OldCantus. Jiwoo set up a chant delete page last summer, which was just for ease of testing as far as I know (It's set to be removed by #481), but evidently it may be worth keeping it around after all. We'd also need a source-delete and a sequence-delete page, yes?

annamorphism:
I didn't check sequences (since the Bower database is basically fixed and not added to), but yes, both sources and chants are delete-able.

jacobdgm:
Right re: sequences - we do have an Sequence Edit page, presumably for making quick corrections in case of discovered typos etc., but if we need to make drastic changes like deleting a sequence, we can do it in the admin panel.

jacobdgm:
#481 has been booting around for long enough - we should figure out what we want to do with this, one way or another.

First option: we add a delete button to all Model Edit pages. This way, if you have access to edit a thing, you have access to delete it. We would also add simple pages for the user to confirm they actually want to delete the thing.

[...]

PR #951, opened 2023-08-08
Create Source Delete page

### LINKED ISSUE #804: Cry for help: deleting one of "My Sources" is not obvious (in fact... I can't figure it out)
As an editor on Cantus, I can "Add new sources," and add chants to those sources. This function works, because I have successfully done so. However, I now can't make my source go away.

I can delete individual chants (see below), but I can't find where I can delete the entire source.

Deleting a chant

So I guess there are two possible issues here:
1. If it's actually not possible to delete a source once you've created it, it'd be great for that to be possible.
2. If it is possible to delete a source, then perhaps the "Delete" button should be somewhere more obvious to dumb users like me.

jacobdgm:
This PR implements a couple of things related to the Source Delete page:
- Implements a new Source Delete page (i.e. fixes #804)
- On the Source Edit page, adds a link to the Source Delete page
- Makes some tweaks to the Chant Delete page, cleaning it up and making it resemble the new Source Delete page.
- Updates the path for the Chant Delete page, from `/chant-delete/<id>` to `/chant/<id>/delete/` (the Source Delete path is constructed similarly).
- Adds some tests for the Source Delete page, updates some to match the new Chant Delete path, and refactors some to follow best practices (i.e. using `reverse` rather than hard-coding paths)


jacobdgm:
In addition to fixing #804, this PR brings us a step closer to resolving #486 - all we need now is a Sequence Delete page modelled on our existing Chant Delete page

django/cantusdb_project/main_app/views/source.py
```diff
@@ -218,6 +224,40 @@ def form_valid(self, form):
...
+    """
+
+    model = Source
+    template_name = "source_confirm_delete.html"
+
+    def test_func(self):
+        user = self.request.user
+        source_id = self.kwargs.get(self.pk_url_kwarg)
+        source = get_object_or_404(Source, id=source_id)
+
```
lucasmarchd01: You could just do `return user_can_edit_source(user, source)` here although the function name doesn't exactly match.

django/cantusdb_project/main_app/views/source.py
```diff
@@ -268,7 +308,7 @@ def form_valid(self, form):
         source = form.save()
 
         for old_editor in old_current_editors:
-            old_editor.sources_user_can_edit.remove(source)
+            old_editor.sources_3.remove(source)
```
lucasmarchd01: what is `sources_3`?

jacobdgm: oh! I don't recall making this change - something must have happened while sorting out the merge conflict. Thanks for catching this!

django/cantusdb_project/main_app/views/source.py
jacobdgm: good! I made this substitution, and the tests still pass.

lucasmarchd01 (approved):

Issue #936, opened 2023-08-04
Allow user to adjust Pagination


jacobdgm:
Users of CantusDB should be able to adjust how many results to see per page, e.g. 10, 50, 100, all, etc.

Ideally, this preference will persist throughout a session, or for a given user between sessions.

PR #952, opened 2023-08-08
Add functionality to pagination

### LINKED ISSUE #917: sources list could paginate better
When looking at the list of all sources, I find it annoying to have to page through half the sources to get to one in, say, the Netherlands--in OldCantus I could scroll all the way through, but now that things are paginated (good!) I have to load up pages of sources in Canada, Switzerland, France, etc. to get to NL-whatever. Of course, I _could_ simply search for the siglum in the search bar, so this is a minor inconvenience, but I am a creature of habit and it would be nice to have the ability to go straight to page 4 of 6 (or whatever).

lucasmarchd01:
Currently the pagination only displays the current page and options to skip to the next/previous page or the first/last page. Now, we display links to the next and previous 5 pages (we can change this number). This will allow the user to jump to a section further in the pagination instead of having to click through each next page.

Fixes #917 as long as the number of pages displayed on left and right seems reasonable, we still need to implement #936

### DIFF
--- a/django/cantusdb_project/main_app/templates/pagination.html
+++ b/django/cantusdb_project/main_app/templates/pagination.html
@@ -7,12 +7,12 @@
             <a href="?{% url_add_get_params page=page_obj.previous_page_number %}">previous</a>
         {% endif %}
 
-        {% if page_obj.number > 5 %}
+        {% if page_obj.number > 6 %}
             <span class="ellipsis">...</span>
         {% endif %}
 
         {% for num in page_obj.paginator.page_range %}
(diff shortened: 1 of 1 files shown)


jacobdgm (approved):
This looks reasonable. A screenshot might have been helpful, but I'll assume this looks good as it is.

lucasmarchd01:
> This looks reasonable. A screenshot might have been helpful, but I'll assume this looks good as it is.

@jacobdgm I definitely took screenshots as I was making the PR, somehow I forgot to include them!
image

PR #956, opened 2023-08-09
Update cron.txt so it runs certbot script every week

jacobdgm:
This PR updates `cron/cron.txt` so it runs the certbot script every week, instead of every 2 months.

### DIFF
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -10,4 +10,4 @@
 20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
 30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
 40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
-50      4       *       *       0       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org; /usr/local/bin/docker-compose exec -it nginx nginx -s reload
+50      4       *       *       0       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org --renew-hook "nginx -s reload"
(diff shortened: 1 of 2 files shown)


jackyyzhang03 (approved):
Looks good to me!

ahankinson:
You also have to reload/restart your webserver so that it picks up on the new certificate...

jacobdgm:
> You also have to reload/restart your webserver so that it picks up on the new certificate...

@kqct: we were just about to add build/down/up commands like so:
```/usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org; /usr/local/bin/docker-compose build; /usr/local/bin/docker-compose down; /usr/local/bin/docker-compose up -d```

...but then we realized that we have a volume already set up for certificates. This means the certificates will be accessible in the containers as soon as `certbot` finishes running, doesn't it? Or will we still need to restart the nginx container to make the updated certificate stick?

ahankinson:
a) the webserver itself needs to restart to pick up on the new certificate since they load them at server start
b) you should use a renew hook on certbot to only trigger the restart when the certificate correctly renews; otherwise you're restarting the service for nothing every time the script is run.

kqct:
You could also run `nginx -s reload` (or equivalent) to have the webserver pick up the certificates with no downtime. That might be a better option.

jacobdgm:
@kqct `/usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org; /usr/local/bin/docker-compose exec -it nginx nginx -s reload`?

jacobdgm:
testing `/usr/local/bin/docker-compose exec -it nginx nginx -s reload` locally, it seems to do what we want it to do. I pushed a new commit with this update; requesting reviews of this from @kqct and from @ahankinson

kqct (approved):

ahankinson:
I find docker makes everything so complicated. 

Why not:

`/usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org --renew-hook "nginx -s reload"`

This will only run the `nginx -s reload` command when the certificate is actually renewed. Otherwise you're still just running the reload every time the cron job runs.

jacobdgm:
I've updated the `cron.txt` to use the command with `--renew-hook` that you suggested, @ahankinson. I also noticed that I had set the command to run every hour on Sundays, rather than just once a day on Sundays, so I fixed that too.

timothydereuse (approved):

PR #961, opened 2023-08-09
Create redirects for old document paths

### LINKED ISSUE #960: The paths to several documents are different from OldCantus to NewCantus
can you make sure that links in the old CantusDB is correctly mapped to the new CantusDB documents?

For example, the following link should be discoverable by changing the domain name:
https://cantus.uwaterloo.ca/sites/default/files/documents/2.%20Volpiano%20Protocols.pdf

A user who changes the above link to this:
https://cantusdatabase.org/sites/default/files/documents/2.%20Volpiano%20Protocols.pdf

It should lead to here:
https://cantusdatabase.org/static/documents/2.%20Volpiano%20Protocols.pdf

Since django stores its static files differently than OldCantus did, we'll need to set up redirects from the old paths to the new paths.

jacobdgm:
This PR creates a function-based view to redirect old document paths (most of them are of the form `/sites/default/files/documents/the-documents-filename.pdf`) to their new locations in NewCantus (mostly in the form `/static/documents/the-documents-filename.pdf`). It fixes #960.

I also wrote a test for the view. I spent a while trying to make the test fully rigorous but was getting different results than I expected; I figured I was spending more time on it than it was worth, so I left a comment in the code describing the difficulty I encountered and inviting improvements from future developers.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -338,7 +338,7 @@ class PermissionsTest(TestCase):
 
         response = self.client.get(f"/source/{assigned_source.id}/delete")
         self.assertEqual(response.status_code, 403)
-        
+
         # Content Overview
         response = self.client.get(reverse("content-overview"))
         self.assertEqual(response.status_code, 403)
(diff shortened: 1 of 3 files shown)


lucasmarchd01:
The only thing that is concerning to me is the last test that you were trying to implement. I think it's important to know what response we are expecting for this redirect, but as long as it's working as expected I don't see a problem in merging this for now.

lucasmarchd01 (approved):

jacobdgm:
I agree the not-quite-complete test is not ideal. But yes, I've tested it manually in the browser a bunch, and it feels quite good to go.

PR #970, opened 2023-08-10
Add autocomplete fields for source create/edit pages

### LINKED ISSUE #450: adding editors
1) All the editors to add are displaying in one long list, more-ore-less alphabetical (some weird Anonymour User accounts at the top, and lowercase/special letters after the Zs.) It might be nice if you could start typing your name instead of having to scroll past hundreds of others.

2) What does this actually do? I tried adding a few editors to a source, and it doesn't seem to display anywhere who has already been added as an editor.

lucasmarchd01:
This PR utilizes the django-autocomplete-light (DAL) package to add autocomplete widgets on the source create page and source edit page. Previously, it was difficult to select users from the "current editors" and "melodies entered by" widgets because it wasn't searchable or sorted in any useful way. The source create and source edit pages were also missing the following fields belonging to indexer contributions, so now these fields can be edited from the main site instead of the admin area.
- "inventoried_by"
- "full_text_entered_by"
- "proofreaders"
- "other_editors"

From the DAL docs, for better understanding of the implementation:
Autocompletes are based on 3 moving parts:
- widget compatible with the model field, does the initial rendering,
- javascript widget initialization code, to trigger the autocomplete,
- and a view used by the widget script to get results from.

Fixes #450 

Source Create page:
image

Source Edit page:
image


jacobdgm (approved):
This looks good - great that this is set up now!

PR #972, opened 2023-08-11
Source Edit: update Rism Siglum field to use ModelChoiceField

### LINKED ISSUE #971: Source Create/Edit - RISM siglum field broken
In an email from Debra:

> For the RISM field in sources (such as this one: https://cantusdatabase.org/source/123664), I see 4100 entered instead of a RISM siglum. Is there a source for this number? (Like, is it in a table somewhere?)
> That's the place where I used to put "B-TO", and then in the Siglum field I would copy that again plus add the shelfmark. I'm wondering how I should fill out the RISM field in the future. I think we're likely changing this for New-NewCantus to be the RISM code in one field and then just the shelfmark in another, but what should we do in the meantime?

Currently, this field uses the ID of the RismSiglum object in the database - not at all useful! OldCantus has an autocomplete-style widget for this field (the user begins typing, and a list of matching values appears below the field); this may be a good place for a filterable list widget.

jacobdgm:
This PR changes the RISM siglum field on the Source Edit page to use a ModelChoiceField widget (previously, a user needed to know the ID of the rism siglum to assign a source to a rism siglum, which was not very useful). It was already set up properly on the Source Create page, so I copied the code from there.

Going to say this fixes #971 - it doesn't exactly solve the issue as described there currently, but this PR at least makes it useable; I'll open a new issue for making it a filterable list field, and mark it as an enhancement.

with the new changes:
Screenshot 2023-08-11 at 1 02 50 PM

previously, it was:
Screenshot 2023-08-11 at 1 10 30 PM

for reference, this is how the source create page looks (it works here, so I just took the code from this page):
Screenshot 2023-08-11 at 1 11 42 PM

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -212,6 +212,7 @@ class SourceCreateForm(forms.ModelForm):
         ]
         widgets = {
             "title": TextInputWidget(),
+            "rism_siglum": TextInputWidget(),
             "siglum": TextInputWidget(),
             "provenance_notes": TextInputWidget(),
             "date": TextInputWidget(),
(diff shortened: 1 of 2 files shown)


lucasmarchd01:
@jacobdgm If you'd like, I can probably implement the filterable list today to skip this step!

dchiller (approved):

dchiller (approved):

PR #975, opened 2023-08-11
Add autocomplete widget for Rism Siglum

### LINKED ISSUE #973: Source Create/Edit pages: RISM siglum field could be a Filterable List widget
Spinning off of #971, I fixed the immediate problem, but I still think it could be good to convert the RISM siglum field to be a Filterable List.

lucasmarchd01:
Sorry for being just a little bit late! 
This PR uses an autocomplete widget for the source rism siglum field on the source create and edit pages.

image

Fixes #973

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -223,6 +223,7 @@ class SourceCreateForm(forms.ModelForm):
             "fragmentarium_id": TextInputWidget(),
             "dact_id": TextInputWidget(),
             "indexing_notes": TextAreaWidget(),
+            "rism_siglum": autocomplete.ModelSelect2(url="rismsiglum-autocomplete"),
             "current_editors": autocomplete.ModelSelect2Multiple(
                 url="current-editors-autocomplete"
             ),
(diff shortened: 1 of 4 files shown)


django/cantusdb_project/main_app/forms.py
```diff
@@ -515,6 +509,7 @@ class Meta:
         ]
         widgets = {
             "title": TextInputWidget(),
+            "rism_siglum": autocomplete.ModelSelect2(url="rismsiglum-autocomplete"),
```
jacobdgm, line 512: from the `SourceCreateForm`, you removed
```python
rism_siglum = forms.ModelChoiceField(
        queryset=RismSiglum.objects.all().order_by("name"), required=False
    )

rism_siglum.widget.attrs.update(
        {"class": "form-control custom-select custom-select-sm"}
    )
```
Do you need to do the same below, within `SourceEditForm`?

lucasmarchd01: Yes, done! Thanks for finding that.

jacobdgm (approved):
This looks good!

Issue #977, opened 2023-08-14
500 error after syllabifying?


annamorphism:
From an email today:

"I tried to transcribe hymn Quia defecerunt sicut fumus dies from [FIN-Hy F.m.I.85](https://cantusdatabase.org/source/712830) to Cantus Database. The automatic syllabification didn’t spell all the syllables correctly, so I bravely tried the new “Edit syllabification” option below Volpiano. I made the changes, saved the chant, and got the “Chant updated successfully” -message. However, when I tried to edit the chant again, the software opened a new window displaying just a message “Serve error 500”.
I could have just deleted the chant and transcribed it again, but I started to wonder what the reason behind the error might be. Do you have any ideas what might have caused it?"

I haven't yet reproduced the error, but the chant in question does indeed throw an error: https://cantusdatabase.org/chant/715252.

jacobdgm:
I don't have time to properly investigate this right now, but in case the chant is changed/deleted, here are some current values (taken from the admin area):

full text std: `Quia defecerunt sicut fumus dies mei et ossa mea sicut in frixo-[rio confrix]-a sunt`
ms full text: `Quia defecerunt sicut fumus dies mei et ossa mea sicut in fixo-{rio confrix}-a sunt`
syllabized full text: `Qui-a de-fe-ce-runt si-cut fu-mus di-es me-i et os-sa me-a si-cut in fix-o- {ri-o con-frix} -a sunt`
volpiano: `1---d--da7---cde--d--dff-ghf-fe-gf-fe-ed-efef--ed---d--fdf---gfghgfe7--egfe-fed-ed---f--g---fgfdc-fgfdc-fed-gf-fed--dfd-dc7---d---ef--d---dcdfg--fe-fg-gf---c--df---f---f--ef-ghf777---6------6---d---dc-dff-ghgef-efhgf-fed---4`

annamorphism:
thanks, that is helpful! I suspect it has to do with how the curly brackets interact with volpiano. I created a dummy text with brackets which worked just fine as text until I tried to glue it to some notes, and then it also gave me a 500 and refused to let me edit the chant. 
(Though it did eventually give me an interesting triplicate success message, which is fun:
<img width="760" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/5150c74a-1eda-4249-8846-7fc065567245">
)

annamorphism:
[...]

PR #981, opened 2023-08-15
Fix IndexError when aligning pre-syllabified text with volpiano

jacobdgm:
This PR solves #977, filtering to remove empty syllables created when we try to split a pre-syllabified text into syllables.

(if a word in the pre-syllabified text ended with a `-`, an empty syllable would be created when we split the word on `-`. Later, when we checked for a particular character at `word[-1][-1]`, an IndexError was being raised, leading to the 500 response noted in #977.)

In discussion on #977, it was noted that the suggested syllabification on the Edit Syllabification includes unnecessary leading spaces. This PR fixes that too.

(When syllabifying a text, we currently add an empty word at the beginning, to correspond to the clef. But in our `syllabize_text_to_string` function, we would attach all syllables to each other, adding spaces between these syllables. This led to two unnecessary leading spaces in our suggested syllabification)

### DIFF
--- a/django/cantusdb_project/align_text_mel.py
+++ b/django/cantusdb_project/align_text_mel.py
@@ -358,4 +358,8 @@ def syllabize_text_to_string(text, pre_syllabized):
     """
     syls_text = syllabize_text(text, pre_syllabized)
     human_readable_text = " ".join(["".join(word) for word in syls_text])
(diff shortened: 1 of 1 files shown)


dchiller (approved):
Implementing the new alignment and syllabification process will get rid of some of the convolution of this.

lucasmarchd01 (approved):

PR #990, opened 2023-08-16
Move user permission checks to one location in code

### LINKED ISSUE #962: Code for permissions checks could all be in one place - opportunity for refactoring
Currently, code dictating which users have access to which pages, as well as code dictating which links should be shown to whom, is currently scattered through the code base. Whenever I want to know who has access to a page, I need to think really hard about all the places I need to look to answer what should be a simple question.

It could be helpful to put test functions for all our different views in one place - perhaps a single file in `main_app/`.

If we did this, we could simply call the relevant function in each of our views' `test_func` method, and we could call the same function in the `get_context_data` method in our other views, to decide whether or not to display a link to a given page. This way, we could be confident that no one is being shown a link that would lead to a 403/401 page.

lucasmarchd01:
This PR focuses on enhancing the maintainability of our codebase. Previously, locating the sections of code responsible for checking user permissions before rendering views or managing specific links was challenging. In response, we've implemented a restructuring effort.

In this update, we've taken all the code related to user permissions and consolidated it into a dedicated file named `permissions.py`, located under the `main_app/` directory. This aims to make it much clearer and easier to identify the parts of our codebase that handle user permissions.

Note: we still use the `has_group` helper tag in `base.html` and `default.html` to handle some user permissions to display certain links in the admin navigation and my sources sidebar. Since these templates don't have a view, we can either leave it as is or set up a function in `context_processors.py` that uses `permissions.py` to handle these cases as well, but I'm unsure if this would make things more complicated.

Resolves #962

### DIFF
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -69,7 +69,7 @@ def user_can_view_unpublished_chant(user: User, chant: Chant) -> bool:
     """
     source = chant.source
     user_is_authenticated: bool = user.is_authenticated
-    return (source.published) or (user_is_authenticated)
+    return (source is not None) or (source.published) or (user_is_authenticated)
 
 
 def user_can_view_unpublished_sequence(user: User, sequence: Sequence) -> bool:
@@ -79,7 +79,7 @@ def user_can_view_unpublished_sequence(user: User, sequence: Sequence) -> bool:
     """
     source = sequence.source
     user_is_authenticated: bool = user.is_authenticated
-    return (source is None) or (source.published) or (user_is_authenticated)
+    return (source is not None) or (source.published) or (user_is_authenticated)
 
 
 def user_can_edit_sequence(user: User) -> bool:
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -115,14 +115,12 @@ def user_can_edit_source(user: User, source: Source) -> bool:
     is_editor: bool = user.groups.filter(name="editor").exists()
     is_contributor: bool = user.groups.filter(name="contributor").exists()
 
-    if (
+    return (
         (is_project_manager)
         or (is_editor and assigned_to_source)
         or (is_editor and source.created_by == user)
         or (is_contributor and source.created_by == user)
-    ):
-        return True
-    return False
+    )
 
 
 def user_can_view_user_detail(viewing_user: User, user: User) -> bool:
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -1,7 +1,10 @@
 from django.db.models import Q
+from typing import Optional
+from main_app.models import Source
+from users.models import User
 
 
-def user_can_edit_chants_in_source(user, source) -> bool:
+def user_can_edit_chants_in_source(user: User, source: Optional[Source]) -> bool:
     """
     Checks if the user can edit Chants in a given Source.
     Used in ChantDetail, ChantList, ChantCreate, ChantDelete, ChantEdit,
@@ -28,7 +31,7 @@ def user_can_edit_chants_in_source(user, source) -> bool:
     )
 
 
-def user_is_proofreader(user) -> bool:
+def user_is_proofreader(user: User) -> bool:
     """
     Checks if the user is a proofreader.
     Used in UserSourceListView.
@@ -36,7 +39,7 @@ def user_is_proofreader(user) -> bool:
     return user.groups.filter(Q(name="project manager") | Q(name="editor")).exists()
 
 
-def user_can_proofread_chants_in_source(user, source_id) -> bool:
+def user_can_proofread_chants_in_source(user: User, source_id: int) -> bool:
     """
     Checks if the user can access the proofreading page of a given Source.
     Used in ChantProofreadView.
@@ -53,7 +56,7 @@ def user_can_proofread_chants_in_source(user, source_id) -> bool:
     return (user_is_project_manager) or (user_is_editor and user_is_assigned_to_source)
 
 
-def user_can_view_unpublished_source(user, source) -> bool:
+def user_can_view_unpublished_source(user: User, source: Optional[Source]) -> bool:
     """
     Checks if the user can view an unpublished Source or Chant/Sequence belonging
     to an unpublished Source on the site.
@@ -69,7 +72,7 @@ def user_can_view_unpublished_source(user, source) -> bool:
     return True
 
 
-def user_can_edit_sequence(user) -> bool:
+def user_can_edit_sequence(user: User) -> bool:
     """
     Checks if the user has permission to edit a Sequence object.
     Used in SequenceDetail and SequenceEdit views.
@@ -77,7 +80,7 @@ def user_can_edit_sequence(user) -> bool:
     return user.groups.filter(name="project manager").exists()
 
 
-def user_can_create_source(user) -> bool:
+def user_can_create_source(user: User) -> bool:
     """
     Checks if the user has permission to create a Source object.
     Used in SourceCreateView.
@@ -88,7 +91,7 @@ def user_can_create_source(user) -> bool:
     ).exists()
 
 
-def user_can_edit_source(user, source) -> bool:
+def user_can_edit_source(user: User, source: Source) -> bool:
     """
     Checks if the user has permission to edit a Source object.
     Used in SourceDetail, SourceEdit, and SourceDelete views.
@@ -112,7 +115,7 @@ def user_can_edit_source(user, source) -> bool:
     return False
 
 
-def user_can_view_user_detail(viewing_user, user) -> bool:
+def user_can_view_user_detail(viewing_user: User, user: User) -> bool:
     """
     Checks if the user can view the user detail pages of regular users in the database or just indexers.
     Used in UserDetailView.
@@ -120,7 +123,7 @@ def user_can_view_user_detail(viewing_user, user) -> bool:
     return viewing_user.is_authenticated or user.is_indexer
 
 
-def user_can_manage_source_editors(user) -> bool:
+def user_can_manage_source_editors(user: User) -> bool:
     """
     Checks if the user has permission to change the editors assigned to a Source.
     Used in SourceDetailView.
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -82,7 +82,7 @@ def user_can_view_sequence(user: User, sequence: Sequence) -> bool:
     return (source is not None) or (source.published) or (user_is_authenticated)
 
 
-def user_can_edit_sequence(user: User) -> bool:
+def user_can_edit_sequences(user: User) -> bool:
     """
     Checks if the user has permission to edit a Sequence object.
     Used in SequenceDetail and SequenceEdit views.
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -5,7 +5,7 @@ def user_can_edit_chants_in_source(user, source):
     """
     Checks if the user can edit Chants in a given Source.
     Used in ChantDetail, ChantList, ChantCreate, ChantDelete, ChantEdit,
-    ChantEditSyllabification, and SourceDetail views
+    ChantEditSyllabification, and SourceDetail views.
     """
     if user.is_anonymous or (source is None):
         return False
@@ -29,7 +29,7 @@ def user_can_edit_chants_in_source(user, source):
 def user_is_proofreader(user):
     """
     Checks if the user is a proofreader.
-    Used in UserSourceListView
+    Used in UserSourceListView.
     """
     if user.groups.filter(Q(name="project manager") | Q(name="editor")).exists():
         return True
@@ -39,7 +39,7 @@ def user_is_proofreader(user):
 def user_can_proofread_chants_in_source(user, source_id):
     """
     Checks if the user can access the proofreading page of a given Source.
-    Used in ChantProofreadView
+    Used in ChantProofreadView.
     """
     if user.is_anonymous:
         return False
@@ -57,7 +57,7 @@ def user_can_view_unpublished_source(user, source):
     """
     Checks if the user can view unpublished Sources or Chants/Sequences belonging
     to unpublished sources on the site.
-    Used in ChantDetail, SequenceDetail, SourceDetail
+    Used in ChantDetail, SequenceDetail, and SourceDetail views.
     """
     display_unpublished = user.is_authenticated
     if (
@@ -84,7 +84,7 @@ def user_can_edit_sequence(user):
 def user_can_create_source(user):
     """
     Checks if the user has permission to create a Source object.
-    Used in SourceCreateView
+    Used in SourceCreateView.
     """
 
     is_authorized = user.groups.filter(
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -35,14 +35,6 @@ def user_can_edit_chants_in_source(user: User, source: Optional[Source]) -> bool
     )
 
 
-def user_is_proofreader(user: User) -> bool:
-    """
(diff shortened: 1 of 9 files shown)


django/cantusdb_project/main_app/permissions.py
```diff
@@ -125,9 +125,9 @@ def user_can_view_user_detail(viewing_user, user):
     Checks if the user can view the user detail pages of regular users in the database or just indexers.
     Used in UserDetailView.
     """
-    if not (viewing_user.is_authenticated or user.is_indexer):
-        return False
-    return True
+    if viewing_user.is_authenticated or user.is_indexer:
+        return True
+    return False
```
jacobdgm: `return (viewing_user.is_authenticated or user.is_indexer)`

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,140 @@
+from django.db.models import Q
+
+
+def user_can_edit_chants_in_source(user, source):
```
jacobdgm: Can we annotate all of these functions indicating they return a boolean? `def user_can_edit_chants_in_source(user, source) -> bool:`, and similarly for all the rest?

lucasmarchd01: Yes, will do.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Checks if the user can edit Chants in a given Source.
+    Used in ChantDetail, ChantList, ChantCreate, ChantDelete, ChantEdit,
+    ChantEditSyllabification, and SourceDetail views.
+    """
+    if user.is_anonymous or (source is None):
+        return False
+
+    source_id = source.id
+    user_is_assigned_to_source = user.sources_user_can_edit.filter(id=source_id)
```
jacobdgm: `user.sources_user_can_edit.filter(id=source_id).exists()` would be clearer.

We could annotate variables throughout this PR. `user_is_assigned_to_source: queryset = user.sources_user_can_edit.filter(id=source_id)` is an accurate description of what this variable contains, but it also reveals that the variable's name doesn't reflect what's there. Better would be `user_is_assigned_to_source: bool = user.sources_user_can_edit.filter(id=source_id).exists()`

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+    if user.is_anonymous:
+        return False
+    user_is_assigned_to_source: bool = user.sources_user_can_edit.filter(
+        id=source_id
+    ).exists()
+
+    user_is_project_manager: bool = user.groups.filter(name="project manager").exists()
+    user_id_editor: bool = user.groups.filter(name="editor").exists()
+
+    if (user_is_project_manager) or (user_id_editor and user_is_assigned_to_source):
```
jacobdgm: `return (user_is_project_manager) or (user_id_editor and user_is_assigned_to_source)`

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+    Used in ChantProofreadView.
+    """
+    if user.is_anonymous:
+        return False
+    user_is_assigned_to_source: bool = user.sources_user_can_edit.filter(
+        id=source_id
+    ).exists()
+
+    user_is_project_manager: bool = user.groups.filter(name="project manager").exists()
+    user_id_editor: bool = user.groups.filter(name="editor").exists()
```
jacobdgm: should this be `user_is_editor` - `is` instead of `id`?

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+        return True
+    return False
+
+
+def user_can_view_unpublished_source(user, source) -> bool:
+    """
+    Checks if the user can view unpublished Sources or Chants/Sequences belonging
+    to unpublished sources on the site.
+    Used in ChantDetail, SequenceDetail, and SourceDetail views.
+    """
```
jacobdgm: The function name suggests this function does one thing, while the docstring suggests it does something else.

> Checks if the user can view unpublished Sources or Chants/Sequences belonging to unpublished sources on the site.

If this is true, why
- is it called `user_can_view_unpublished_source` rather than `user_can_view_unpublished_sources`?
- does it have `source` as a parameter, if it dictates whether the user can view unpublished sources in general?

django/cantusdb_project/main_app/permissions.py
jacobdgm: this can be used in all places where you currently end a function with
```python
if condition:
    return True
return False
```
With `return condition`, you can accomplish the same thing with the same clarity on one line rather than three. I'll stop commenting on this whenever I come across it.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+    user_is_authenticated: bool = user.is_authenticated
+    if (
+        (source is not None)
```
jacobdgm: I feel like this is not the right place for this check - we should check that `source` is actually a source before we pass it in to this function.

Annotations on parameters for all these functions could help. If we had `user_can_view_unpublished_source(user: User, source: Source)`, it would be clear to future developers that they can't pass a `None` as an argument (or, if it's intended that you can pass a `None` in, it should be `source: Optional[Source]`).

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+        and (source.published is False)
+        and (not user_is_authenticated)
+    ):
+        return False
+    return True
+
+
+def user_can_edit_sequence(user) -> bool:
```
jacobdgm: this should either be `user_can_edit_sequences(user)` or `user_can_edit_sequence(user, sequence)`

Looking at the logic, it appears it should be the former. In which case, it should probably actually be `user_can_edit_sequences(user: User)`

django/cantusdb_project/main_app/templates/user_source_list.html
```diff
@@ -16,13 +16,13 @@ <h3>My Sources</h3>
                                     <b>{{ source.siglum }}</b>
                                 </a>
                             </td>
-                            {% if source.chant_set.all%}
```
jacobdgm, line 19: Does it work to change this to `{% if source.chant_set.exists %}`?

django/cantusdb_project/main_app/permissions.py
lucasmarchd01: My understanding is that it should be optional. In the develop branch we have this check in SequenceDetail
```
        # if the sequence's source isn't published,
        # only logged-in users should be able to view the sequence's detail page
        if (
            (source is not None)
            and (source.published is False)
            and (not self.request.user.is_authenticated)
        ):
            raise PermissionDenied()
```

And for chant and source detail we have something like
```

        # if the chant's source isn't published, only logged-in users should be able to
        # view the chant's detail page
        source = chant.source
        display_unpublished = user.is_authenticated
        if (source.published is False) and (not display_unpublished):
            raise PermissionDenied()
```
My guess is that it's because some sequences in the database don't belong to a source, which would explain the extra check for source is not None.

jacobdgm: recommending:
```python
def user_can_view_source(user: User, source: Source) -> bool:
    return source.published or user.is_authenticated
```

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+    return (
+        (user_is_project_manager)
+        or (user_is_editor and user_is_assigned_to_source)
+        or (user_is_editor and source.created_by == user)
+        or (user_is_contributor and user_is_assigned_to_source)
+        or (user_is_contributor and source.created_by == user)
+    )
+
+
+def user_is_proofreader(user) -> bool:
```
jacobdgm: Do we use "proofreader" anywhere else in the code base? I thought this was only used in OldCantus, and that OC "proofreaders" are now NC "editors".

A better name for this function might be `user_can_access_proofreading_pages` or something similar

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+def user_can_edit_sequence(user) -> bool:
+    """
+    Checks if the user has permission to edit a Sequence object.
+    Used in SequenceDetail and SequenceEdit views.
+    """
+    return user.groups.filter(name="project manager").exists()
+
+
+def user_can_create_source(user) -> bool:
```
jacobdgm: This is fine, but maybe `user_can_create_sources` would be better. The current wording slightly suggests that the function should take a `source` argument as well as a `user`

django/cantusdb_project/main_app/permissions.py
jacobdgm: Or maybe it would make sense to get rid of this function, and use `user_can_proofread_chants_in_source` in all situations? Not sure if that works.

django/cantusdb_project/main_app/permissions.py
lucasmarchd01: I did try this at first locally but there were Sequences in the database that didn't belong to a source which would lead to an error when trying to get to the SequenceDetailView, which is why we need the extra check for Source is not None.

The other option would be to add another function to check if a user can view a sequence belonging to an unpublished source.

django/cantusdb_project/main_app/views/user.py
```diff
@@ -109,6 +110,7 @@ def get_context_data(self, **kwargs):
         page_obj = paginator.get_page(page_number)
 
         context["user_created_sources_page_obj"] = page_obj
+        context["user_is_proofreader"] = user_is_proofreader(self.request.user)
```
jacobdgm: Looking at `user_detail.html`, I can't see `user_is_proofreader` anywhere. Is it necessary to add it to `context`?

lucasmarchd01: It's used in `user_source_list.html`, so it should be added to `context` in this case

jacobdgm (changes requested):
Thanks for putting this together! Not much of substance to change - a couple of functions could be renamed, type annotations would be an excellent idea in as many places as possible, and the logic could be tightened up in a couple of places.

One final thought - this might be a natural time to ensure our permission check tests are comprehensive. But I expect them to be sufficient as-is, so no need to double-check these unless you really want to.

django/cantusdb_project/main_app/templates/user_source_list.html
lucasmarchd01: yes!

django/cantusdb_project/main_app/views/user.py
jacobdgm: Ah, okay! Somehow I thought this was part of a different view. Very good!

django/cantusdb_project/main_app/permissions.py
jacobdgm: Huh. I just checked now, and it does seem we can have chants/sequences without sources. Surprising, but I guess this is an OldCantus legacy data thing.

In this case, it sounds like we want a `user_can_view_chant` (and possibly `..._sequence`) function, and handle the case where a chant/sequence doesn't have a source there. If we've got a function called `user_can_view_source`, we should use it to check whether a particular user can view a particular source detail page, and not as a substitute to answer another question like whether a particular user can view a particular sequence detail page.

jacobdgm: Also - we may want to check in with Debra about whether to display chants/sequences without sources. In OC, there was a boolean `visible` property (sources could definitely have it, and possibly chants/sequences could too), which we got rid of, rolling everything into our `published`. Since all our visibility-related logic is based on a chant/sequence's source, we may not want to default to displaying chants to everyone on the grounds that they don't belong to an unpublished source.

django/cantusdb_project/main_app/permissions.py
lucasmarchd01: This is possible now, needed to do a bit of reworking in `UserSourceListView` to add a list of tuples (source, permission for source) to the pagination object.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,139 @@
...
+        id=source_id
+    ).exists()
+
+    user_is_project_manager: bool = user.groups.filter(name="project manager").exists()
+    user_is_editor: bool = user.groups.filter(name="editor").exists()
+
+    return (user_is_project_manager) or (user_is_editor and user_is_assigned_to_source)
+
+
+def user_can_view_unpublished_source(user: User, source: Source) -> bool:
```
jacobdgm: I don't understand this function name - I'd expect `user_can_view_source`. In the last line, we `return (source.published) or (user_is_authenticated)`. We shouldn't need to check that `source.published == True`: since the function name is `user_can_view_unpublished_source`, we already know the source is unpublished.

Based on both the function name and the docstring, anyone who passes a published source into the function is not using the function properly. The function, as it currently exists, clearly accepts only unpublished sources.

We can apply similar reasoning to the two other `user_can_view_unpublished_...` below.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+def user_can_view_unpublished_chant(user: User, chant: Chant) -> bool:
+    """
+    Checks if the user can view a Chant belonging to an unpublished Source on the site.
+    Used in ChantDetail, SequenceDetail, and SourceDetail views.
+    """
+    source = chant.source
+    user_is_authenticated: bool = user.is_authenticated
+    return (source is not None) or (source.published) or (user_is_authenticated)
```
jacobdgm, line 72: But source will never be None - based on the docstring, this function accepts only chants that belong to unpublished sources. Checking for `source is not None` only makes sense for a function named `user_can_view_chant`

You get the point.

django/cantusdb_project/main_app/permissions.py
lucasmarchd01: Makes, sense. Based on the context of how they are used in the view, we should name it instead to `user_can_view_....` because the function should accept both published and unpublished sources.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    Checks if the user can view a Sequence belonging to an unpublished Source on the site.
+    Used in ChantDetail, SequenceDetail, and SourceDetail views.
+    """
+    source = sequence.source
+    user_is_authenticated: bool = user.is_authenticated
+    return (source is not None) or (source.published) or (user_is_authenticated)
+
+
+def user_can_edit_sequence(user: User) -> bool:
```
jacobdgm: function names are much better now! This one, though, might be better named `user_can_edit_sequences` - `user_can_edit_sequence` suggests there should be a `sequence` parameter.

django/cantusdb_project/main_app/views/user.py
```diff
@@ -95,20 +100,27 @@ def get_queryset(self):
...
         )
+        user = self.request.user
+        sources_with_proofread_permissions = [
+            (source, user_can_proofread_chants_in_source(user, source))
+            for source in my_sources
+        ]
 
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
+        user_sources_paginator = Paginator(sources_with_proofread_permissions, 10)
```
jacobdgm, line 109: I see we've replaced `paginate_by = 100` with `Paginator(..., 10)`. Just verifying that this 100 -> 10 substitution is intentional.

jacobdgm (approved):
This is really good - thanks again for putting this together! I'll let you merge this after you look at the pagination question I posed, but on the whole, this is excellent!

django/cantusdb_project/main_app/views/user.py
lucasmarchd01: It was, I thought it made sense to be something lower than 100 for this page but maybe something like 20 or 30 could be best?

jacobdgm: Paginating by 10 is fine - as long as this change was intended, and not a typo.

For future changes like this, it might make sense to make a separate commit with a message indicating a change has taken place - if there's a bug with pagination in the future, it will be easier to track down the commit that caused the substitution if it has a descriptive commit message.

PR #997, opened 2023-08-17
Add Feast autocomplete for chant create/edit pages

### LINKED ISSUE #987: Feast dropdown could use autocomplete
The feast selector dropdown select on the chant edit/create page could use a searchable autocomplete widget so the user won't have to scroll through hundreds of items.

lucasmarchd01:
Resolves #987. Now feasts come in a searchable dropdown widget where the user can search by feast name or feast code. This makes it easier and faster to find the desired feast rather than having to scroll through hundreds of list items in a regular dropdown.

image


jacobdgm (approved):
Nice!

PR #1011, opened 2023-08-22
Insert "add chant" links to chant-proofread and chant-edit pages

### LINKED ISSUE #1004: question: why is "add new source" on the edit-chants page?
On the full-text editor/"edit-chants" page, you get the option to create a new source after the explainer text of how to click the edit button. I know the short answer is that this is there because it was in OldCantus but...what I actually always want from this page is an easy way to make new CHANTS, and today I wondered if this button (which I have always ignored, because I never want to go making new sources from within a source) was actually intended to be that once upon a time?

lucasmarchd01:
Resolves #1004. We now display a link to edit chants in a given source on the full text & volpiano edit page as well as the chant-proofread page.

image

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_edit.html
+++ b/django/cantusdb_project/main_app/templates/chant_edit.html
@@ -235,18 +235,20 @@
                     <dd><small>3) The <b>fulltext</b> and <b>Volpiano</b> fields should appear in this area</small></dd>
                     <dd><small>4) Edit the fields <b>according to the manuscript, following the fulltext guidelines created by Cantus</b></small></dd>
                     <dd><small>5) Click <b>"SAVE"</b></small></dd>
-                </dl> 
-                <a href="{% url 'source-detail' source.id %}" title="{{ source.title }}" style="display: inline-block; margin-top:5px;">
-                    {{ source.siglum }}
-                </a>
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/templates/chant_edit.html
```diff
@@ -236,13 +236,17 @@ <h3>Full text &amp; Volpiano edit form</h3>
...
                 </dl> 
-                <a href="{% url "source-create" %}" style="display: inline-block; margin-top:5px;">
-                    <small><b>&plus; Add new source</b></small>
-                </a>
-                <br>
                 <a href="{% url 'source-detail' source.id %}" title="{{ source.title }}" style="display: inline-block; margin-top:5px;">
                     {{ source.siglum }}
                 </a>
+                <br>
+                <a href="{% url "chant-create" source.pk %}" style="display: inline-block; margin-top:5px;">
```
jacobdgm: why are we using both `<br>` and `margin-top`? It might be cleaner to put these links in a `<div>` (each link might need its own div - not sure) and apply `margin-top` to each. This would eliminate the need for the `br`, and we may not need `inline-block` either...

jacobdgm (approved):

PR #1013, opened 2023-08-24
Make migrations to remove null=True and blank=True in source ForeignKey

lucasmarchd01:
This PR makes the necessary migrations to remove the `null=True` and `blank=True` settings on our chant and sequence models for the `source` ForeignKey. Trying to remove them both immediately results in an database error:
```
It is impossible to change a nullable field 'source' on chant to non-nullable without providing a default. This is because the database needs something to populate existing rows.
Please select a fix:
 1) Provide a one-off default now (will be set on all existing rows with a null value for this column)
 2) Ignore for now. Existing rows that contain NULL values will have to be handled manually, for example with a RunPython or RunSQL operation.
 3) Quit and manually define a default value in models.py.
```
We first need to provide a default value (such as an empty string "") and then remove the column. I'm not sure why django makes you do this even though we made sure none of the entries in our database have a null source entry.

Now, users should not be able to make chants that don't belong to a source. Existing chants and sequences that didn't have a source have been moved here ([Unknown Source (Chants)](https://cantusdatabase.org/source/1000003)) and here ([Unknown Source (Sequences)](https://cantusdatabase.org/source/1000004))

**Update:** added management command that will create the unknown sources for chants and sequences that can be run locally and on staging.


ahankinson:
> We first need to provide a default value (such as an empty string "") and then remove the column. I'm not sure why django makes you do this even though we made sure none of the entries in our database have a null source entry.

You will need to provide a default value of the same type as the  column type. For foreign keys this is usually an integer. Furthermore some databases impose a restriction on foreign keys that they must exist in the foreign table. (See: https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK)

Django imposes this because migrations deal with the structure of the database, and not the content. You can make a migration on a dev machine and then deploy it in production with different data. The migration doesn’t know anything about the contents, it just needs to know what to do when it finds a null that can no longer be there.

jacobdgm:
> I'm not sure why django makes you do this even though we made sure none of the entries in our database have a null source entry.

Is it the case that every single sourceless chant and sourceless sequence on both Staging and Production is now assigned to an "Unknown Source" source?

This might have been a good place to write a management command/script to create our Unknown Sources and assign sourceless chants/sequences to them - before someone runs these migrations, they can create the relevant dummy sources and be confident that nothing weird will happen when the migrations are run. Without this, I'm going to need to figure out how to create these sources and assign the right chants/sequences to them on my own local machine, instead of running a simple script. Same goes for running migrations on Staging.

(update: I just checked, and there are some chants on Staging without sources. So yes, can you please write this management command?)

lucasmarchd01:
> (update: I just checked, and there are some chants on Staging without sources. So yes, can you please write this management command?)

Yes, will do. This was an oversight on my part, it definitely makes sense to have the management commands.

django/cantusdb_project/main_app/management/commands/add_unknown_source.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        cantus_segment = Segment.objects.get(id=4063)
+        calvin_segment = Segment.objects.get(id=4064)
+        chants = Chant.objects.all().filter(source__isnull=True)
+        sequences = Sequence.objects.all().filter(source__isnull=True)
+
+        if len(chants) > 0 and len(sequences) > 0:
```
jacobdgm: should this be two separate `if` statements?

jacobdgm (approved):
Approving. Where I left a comment, I don't understand why these two conditions are rolled up in a single `if` statement, but with the current state of our data, it's likely that it doesn't matter. Feel free to change this, or not, before merging.

PR #1016, opened 2023-08-29
Ensure Short Video Introduction opens in new tab

### LINKED ISSUE #1012: Enhancement idea: youtube tutorial opens in a new tab
PSA: this is a super not urgent and very specific idea.

In the "About" dropdown menu, there's a "Short video introduction (Youtube)." Clicking on that opens the link in the same tab, which makes the user lose Cantus. I think that having the tutorial open in a new tab would be more convenient, and also match what I would expect from that sort of hyperlink.

jacobdgm:
fixes #1012


jacobdgm:
wait what - that's way more commits than should be part of this. Let me create a new PR to try to fix this.

Issue #815, opened 2023-07-06
Suggested chant ID button not working


lucasmarchd01:
I am copying this issue mentioned in Kathleen's email from notes on the production server. I was not able to replicate this issue, the ID button works fine for me, but this is worth investigating what might be causing this.

"For both the newly added source and the previously started source, Production site brings up the suggested chants that might follow BUT clicking on the desired ID button does not work. The Input tool works, so I used that, but having the suggestions available to select if correct is better!"

jacobdgm:
Trying this on production now (as I did last month), I'm not able to recreate this bug. Testing in Safari, Chrome and Firefox, it works on all three browsers. So I'm not sure what to do about this.

@fujinaga, do you have any suggestions? Should we email Kathleen and ask her to try again, and give us more information on browser/OS/etc. if it's still broken for her? Or should we just close this issue?

fujinaga:
We should always reply to users immediately then when things are fixed, you should check with them to make sure that things look good from their end.

jacobdgm:
I emailed Kathleen saying that this should be working now, and asking her to send us more information if it still seems broken. Closing this for now; we can reopen it if Kathleen still finds this not to be working.

PR #1019, opened 2023-08-30
Chant Create: fix bug in chant suggestions

### LINKED ISSUE #998: some chant suggestions don't work?
In trying to create some test chants to investigate #996 , I found that the chant suggestions sometimes autofill but not always.  For example, having created 
image
 the suggestion for g00556.Tp doesn't populate the fields, while all the other suggestions do. And for an earlier chant, the first suggestion didn't work but the others did: 
image
Still investigating why this is.

jacobdgm:
This PR fixes a bug on the Chant Create page in the Suggested Chants feature (i.e. fixes #998, #815). Previously, we created a bunch of functions, each named after the Cantus ID of one of the suggested chants/cantus_ids, that we would call when the corresponding button was pressed. If the Cantus ID had a period in it, however, the function name would be invalid, preventing the button from working.

This PR:
- changes how these functions are named, ensuring the function name does not contain any periods or colons
- adds `title` attributes to various elements in the Suggested Chants panel (this way, if a Cantus ID is too long for its button, or if the full text is truncated with an ellipsis, the user can mouse over these elements to see the entire value)
- renames a variable in the templates: previously, we iterated through each of the suggestions with `{% for chant in suggested_chants %}`. But each suggestion was really a Cantus ID rather than a chant, so this has been changed to `{% for cantus_id in suggested_chants %}`

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_create.html
+++ b/django/cantusdb_project/main_app/templates/chant_create.html
@@ -249,10 +249,10 @@
                         Cantus ID: <a href="https://cantusindex.org/id/{{ previous_chant.cantus_id }}" target="_blank">{{ previous_chant.cantus_id }}</a><br>
                             {% if suggested_chants %}
                                 {% for chant in suggested_chants %}
-                                    <input type="button" style="width: 80px" value="{{ chant.cid }}" onclick="autoFill{{ chant.cid }}()"></input>
+                                    <input type="button" style="width: 80px" value="{{ chant.cid }}" onclick="autoFill{{ chant.cid_clean }}()"></input>
                                     <strong>{{ chant.genre }}</strong> - {{ chant.fulltext|truncatechars:100 }} (<strong>{{ chant.count }}x</strong>)<br>
                                     <script>
-                                        function autoFill{{ chant.cid }}() {
+                                        function autoFill{{ chant.cid_clean }}() {
                                             document.getElementById('id_cantus_id').value = "{{ chant.cid }}";
                                             document.getElementById('id_genre').value = {{ chant.genre_id }};
                                             document.getElementById('id_manuscript_full_text_std_spelling').value = "{{ chant.fulltext }}";
--- a/django/cantusdb_project/main_app/templates/chant_create.html
+++ b/django/cantusdb_project/main_app/templates/chant_create.html
@@ -248,14 +248,14 @@
                         {{ previous_chant.folio }} {{ previous_chant.c_sequence}} <a href="{% url 'chant-detail' previous_chant.id %}" target="_blank">{{ previous_chant.incipit }}</a><br>
                         Cantus ID: <a href="https://cantusindex.org/id/{{ previous_chant.cantus_id }}" target="_blank">{{ previous_chant.cantus_id }}</a><br>
                             {% if suggested_chants %}
-                                {% for chant in suggested_chants %}
-                                    <input type="button" style="width: 80px" value="{{ chant.cid }}" onclick="autoFill{{ chant.cid_clean }}()" title="{{ chant.cid }}"></input>
-                                    <strong>{{ chant.genre }}</strong> - <span title="{{ chant.fulltext }}">{{ chant.fulltext|truncatechars:100 }}</span> (<strong>{{ chant.count }}x</strong>)<br>
+                                {% for cantus_id in suggested_chants %}
+                                    <input type="button" style="width: 80px" value="{{ cantus_id.cid }}" onclick="autoFill{{ cantus_id.cid_clean }}()" title="{{ cantus_id.cid }}"></input>
+                                    <strong>{{ cantus_id.genre }}</strong> - <span title="{{ cantus_id.fulltext }}">{{ cantus_id.fulltext|truncatechars:100 }}</span> (<strong>{{ cantus_id.count }}x</strong>)<br>
                                     <script>
-                                        function autoFill{{ chant.cid_clean }}() {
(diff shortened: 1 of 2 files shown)


jacobdgm:
I initially removed all the `.`s from function names and thought everything was solved. But then I looked through a couple of the documents on the Cantus Index site, and found that hymn verses sometimes have `:`s, and these break chant suggestion buttons, so I had to account for these too. @annamorphism, can you confirm no other interesting characters ever appear in Cantus IDs? `#`?

dchiller:
Could you not just have the function names be something like the number index of the suggested chant? Then you don't have to worry about the cantusid characters now or any being added in the future....

Also, are you sure that taking out these characters will never cause conflicts (ie. are `12345.6` and `123456` not both valid cantus id's?)

jacobdgm:
> Could you not just have the function names be something like the number index of the suggested chant? Then you don't have to worry about the cantusid characters now or any being added in the future....
> 
> Also, are you sure that taking out these characters will never cause conflicts (ie. are `12345.6` and `123456` not both valid cantus id's?)

This is a good point. I'm pretty sure that every Cantus ID needs to be 6 digits before any `.` or `:` suffixes, but I'm not certain of this. I'll make this change.

annamorphism:
> Could you not just have the function names be something like the number index of the suggested chant? Then you don't have to worry about the cantusid characters now or any being added in the future....
> 
I'm not sure what you mean by this...
> Also, are you sure that taking out these characters will never cause conflicts (ie. are `12345.6` and `123456` not both valid cantus id's?)

The base Cantus ID is six digits (sometimes preceded by a letter); in some cases the 6 digits are followed by a period or colon, and then one or two digits or letters designating a variant, hymn verse, or language.  There are some exceptions (the h prefix gets 5 digits, the Lamentations chants are "Lamen4:16), but none that rely on the period or colon to distinguish between otherwise identical digit strings. in other words, 12345.6 is simply not a valid Cantus ID because it's not long enough. BUT it may be wise to avoid this problem anyway, in case seven-digit strings become valid IDs in the future.

dchiller:
> > Could you not just have the function names be something like the number index of the suggested chant? Then you don't have to worry about the cantusid characters now or any being added in the future....
> 
> I'm not sure what you mean by this...

Rather than have functions like `autoFill123456` where 123456 is the Cantus ID, just have `autoFill1` be the function for the first suggested cantus id, `autoFill2` be the suggestion for the second cantus id, etc.

ahankinson:
Or just have a single autofill function and pass the id in as a parameter… they al do the same thing so making a duplicate function for each one is … inefficient.

ahankinson:
Or just have a single query selector select all the buttons and subscribe them to the `click` event. With this you get the current button as the event "target" so you can get any attributes from the element. So you can remove the `onclick` attribute from the buttons altogether, and manage the subscription and the behaviour all in one place.

jacobdgm:
> Or just have a single query selector select all the buttons and subscribe them to the `click` event. With this you get the current button as the event "target" so you can get any attributes from the element. So you can remove the `onclick` attribute from the buttons altogether, and manage the subscription and the behaviour all in one place.

Sorry for taking a while to get back to this.

Currently, in our template, `for` every `chant` in `suggested_chants`, we're making a new function where the suggested fulltext/etc. is hard-coded into the function declaration. If I understand your suggestion, @ahankinson, we could instead use a single function. Each button could call the function with a unique argument, the CantusID. The function would know which suggested fulltext to use because we attach the relevant fulltext/etc. to each button using custom attributes, along the lines of `<input type="button" style="width: 80px" value="{{ chant.cid }}" onclick="autoFill({{ chant.cid }})" cantus_id={{ chant.cid }} fulltext={{ chant.fulltext }} genre_id={{ chant.genre_id }}></input>`.

Is this correct? A little bit of googling suggests that we can create our own custom attributes. Is everyone confident that such custom attributes work with all browsers, and/or are there any other gotchas we need to be careful of?

ahankinson:
Not quite. 

In JS, you can subscribe elements to events, if you know how to target them. If you use `querySelectorAll`, for example, with a class name you can query the DOM for all buttons with a given class and subscribe them to a custom `click` event.

See here: https://fjolt.com/article/javascript-multiple-elements-addeventlistener

When you subscribe them, your click handler gets an argument, (usually called `e` or `evt`) which is the event itself. This has `e.target`, which is the element instance being clicked.

In the Django template, you create data attributes with the data you want to pass.

`<input type="button" value="{{ chant.cid }}" data-cid="{{ cantus.cid }}" data-fulltext="{{ chant.fulltext }}" data-genre="{{ chant.genre_id }}"></input>`

So now in your click event, `e.target.dataset` holds all of the values for your data attributes, e.g., `e.target.dataset.cid`. You don't need the onclick attribute at all.

Just be aware that you need to subscribe your elements only after they've been added to the DOM, so you probably want to wrap your subscription code in a block that runs on the `DOMContentLoaded` event. 

https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event

If you really want to keep the onclick attribute, passing `this` gives you the element too.

`<input onclick="someFn(this)" />`

Then `this.dataset` works the same way.

jacobdgm:
In the interests of fixing this quickly, I just pushed another commit ensuring that the function names will always be unique by replaced dots and colons instead of removing them.

I'll open an issue flagging this (creating a single function) as a good opportunity for refactoring. Thanks for the feedback and suggestions, @dchiller, @annamorphism and @ahankinson!

dchiller (approved):
Sounds like a plan!

PR #1027, opened 2023-09-07
Merge Staging into Production, 2023-Sept-07 edition

jacobdgm:
New code is running on the staging server!

Pages that have been changed:
- [x] All pages, header: Short Video Intro now opens in a new tab
- [x] Admin: Chant List - now allows searching/filtering by id
- [x] Admin: Sequence List - now allows searching/filtering by id
- [x] Chant Create: fixed bug where Suggested Fulltext buttons sometimes wouldn't work
- [x] Chant Edit: in the main body of the page, when not editing a specific chant, we now have a link pointing to the Chant Create page.
- [x] Chant Detail: We now display the `differentia_new` ("Differentiae Database") field
- [x] Chant Proofread: in the main body of the page, when not editing a specific chant, we now have links pointing to the Chant Create and Source Create pages.
- [x] Chant/Sequence Create/Edit pages: it is now impossible to create a chant or sequence without a source (existing chants/sequences without a source have been assigned to an unpublished dummy "Unknown Source" source). (Jacob notes that it's difficult to test this on these pages themselves, since they've got the source ID in the URL. But I'm indeed unable to create sourceless chants/sequences in the Admin area, so we're good)


annamorphism:
"now allows searching/filtering by id" ---this is CantusID, correct?

annamorphism:
"we now display a link in the sidebar pointing to the Chant Create page"--there's "add new chant" in the body of a source's edit-chants page, is that what you mean?

annamorphism:
I'm a wee bit confused about the chants in Unknown Source. For example http://206.12.88.113/chant/688446 seems to include itself in its concordances, with a siglum (presumably the one connected to that chant when it was made). But clicking on this siglum takes you back to the chant page and not a source (I'm thinking this maybe shouldn't be linked?). Meanwhile I can't tell if the other chants from Unknown Source show up as their own concordances--I think they don't. Which is fine, since they're anomalous, I'm just unsure what the intended behaviour is here.

jacobdgm:
> "now allows searching/filtering by id" ---this is CantusID, correct?

no - if you know the ID (i.e. primary key) of the chant, you can enter it and that specific chant will come up.

jacobdgm:
> "we now display a link in the sidebar pointing to the Chant Create page"--there's "add new chant" in the body of a source's edit-chants page, is that what you mean?

Oops, you're right. I'll edit my initial comment.

annamorphism:
> > "now allows searching/filtering by id" ---this is CantusID, correct?
> 
> no - if you know the ID (i.e. primary key) of the chant, you can enter it and that specific chant will come up.
oops, realized that my tooling around on admin was still only on production (not sure I can access this on staging?).

jacobdgm:
everything looks good! I'll merge this tomorrow and update the production server. If anyone wants to approve this before merging, that might be best practice ^^

annamorphism (approved):

lucasmarchd01 (approved):

PR #1043, opened 2023-09-14
chant admin: hide volpiano_notes and volpiano_intervals fields

### LINKED ISSUE #1041: Admin Area: Hide "Volpiano notes" and "Volpiano interval" fields
We currently display these fields in the admin area, but they are only for internal use. Instead, we should hide these fields (i.e., exclude, instead of readonly_fields).

lucasmarchd01:
Fixes #1041


jacobdgm (approved):
Very good!

PR #1047, opened 2023-09-14
Merge chant-proofread and chant-edit pages

### LINKED ISSUE #1032: Make proofread boxes available in edit pages
From Debra (email 12 Sept 2023):

> if possible I would like the little blue checkboxes that I click when I proofread a record visible in my "edit" view. (They can also be any version of a "yes/no" box or fill-in, as long as I can mark a record proofread as I go through the source.)

Screenshot from chant edit page in OldCantus

image

According to Debra, these boxes (or something like them) are available in the Taxonomy Manager (I don't have permission to view, so have not seen what they look like there myself), but not in the edit pages, making it a two-window task to proofread/edit and then mark as proofread.

### LINKED ISSUE #1034: We could merge the Chant Edit and Chant Proofread pages
related to #1032, it might be a good idea to merge the Chant Edit and Chant Proofread pages. Currently, the Proofread page has all the same fields as the Edit page, plus checkboxes for the three "proofread" booleans and a dropdown allowing the user to select a "proofread by" (aside: I wonder whether there is any reason that a user would select anyone other than themself as a "proofread by").

It would be simpler to include all the fields on the Chant Edit template, but only display the "proofread" checkboxes if the user has permission to proofread that particular chant. If we did this, we could get rid of the Proofread page entirely.

I suggested this to Debra as a thing to discuss when she's in town in mid-October.

Are there any reasons not to merge these two pages? Is there any situation where a user will want access to _both_ the Proofread and the Edit page for a particular chant?

lucasmarchd01:
Fixes #1032, Resolves #1034. 

This PR addresses concerns about having a chant-proofread page and a chant-edit page. Previously, the chant-proofread page contained checkboxes where a user with proofread access could edit the proofread fields. Now, we merge the chant-proofread page and chant-edit page so that the chant-edit page will display these checkboxes (along with other fields that were exclusively shown on the chant-proofread page). Here is a summary of the changes:

- Add   "chant_range", "manuscript_full_text_std_proofread", "manuscript_full_text_proofread", "volpiano_proofread", and "proofread_by" fields to chant-edit page
- permission check is added to chant-edit view so that certain fields will be displayed based on proofreading access.
- Remove chant-proofread template and view, along with any links or references to this page.
- Add autocomplete widget for the proofread_by field in chant-edit
- update test suite by removing TestChantProofreadView and moving any of it's relevant tests to TestSourceEditChantsView


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1660,6 +1659,10 @@ def get_chants_with_folios(chants_in_feast):
             )
             context["syllabized_text_with_melody"] = text_and_mel
 
+        user = self.request.user
+        context["user_can_proofread_chant"] = user_can_proofread_chants_in_source(
+            user, source
+        )
```
jacobdgm: the two of us just checked this - we no longer use `user_can_proofread_chants_in_source` anywhere, so this function (in `permissions.py`) should really be renamed `user_can_proofread_chant` and should take (User, Chant) as arguments rather than (User, Source).

jacobdgm (approved):

PR #1063, opened 2023-09-21
Fix "Feasts that follow" links

### LINKED ISSUE #1050: "Feasts that follow" don't link
Below the feast selector on chant-create there are helpful suggestions about "Feasts that follow":
image
These look like they should fill in the "feast" field, but they don't seem to--nor do they link out to anything else.

lucasmarchd01:
Since updating the Feast select field on the chant create page, the autoFillFeast function did not work correctly anymore. Clicking on one of the Feasts that follow links would not autopopulate the Django Autocomplete Light field as it did with the default dropdown select. 

 In django autocomplete with select2, items are loaded async. So if you want to programatically select, you need to add the value first. This is also explained in the [select2 documentation](https://select2.org/programmatic-control/add-select-clear-items#create-if-not-exists) 

Now, the autocomplete widget acts as expected, matching the behaviour of OldCantus and NewCantus before updating the autocomplete widgets. Fixes #1050

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_create.html
+++ b/django/cantusdb_project/main_app/templates/chant_create.html
@@ -101,7 +101,18 @@
                                     <a href="javascript:" onclick="autoFillFeast{{ feast.id }}();">{{ feast.name }}</a> ({{ count }}x) |
                                     <script>
                                         function autoFillFeast{{ feast.id }}() {
-                                            document.getElementById('id_feast').value = "{{ feast.id }}";
+                                            var feast_id = {{ feast.id }};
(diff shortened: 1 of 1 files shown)


jacobdgm (approved):
At first glance, I might expect this to cause the feast to appear twice in the list. But if this is the recommended approach (as it appears to be, looking at the documentation), and it doesn't in fact cause duplication in the list, then this looks good. Nice!

lucasmarchd01:
I wondered this as well, and it seems that clicking on another feast doesn't add any duplicates to the list.

PR #1064, opened 2023-09-21
Prevent users from appearing twice in proofreader fields

### LINKED ISSUE #1062: Admin Area, Chant/Sequence Change pages - duplicate names appearing
A few weeks ago, we fixed the issue where users' names would appear twice among the "Available indexers" etc.

The same issue occurs on the Chant and Sequence Change pages, in the "Available proofread by" field. We need to apply the same treatment to these models/forms.

lucasmarchd01:
Fixes #1062. Using `.distinct()`, we can ensure that users will not appear twice in the proofreader widgets. Previously, they would filter based on user groups, sometimes presenting two of the same user if they belong to two different groups at the same time.


jacobdgm (approved):
Looks good!

PR #1068, opened 2023-09-21
Chant admin: fix manuscript full text std spelling widget size

### LINKED ISSUE #1065: Chant admin "Manuscript full text std spelling" widget is small
The Manuscript full text std spelling widget should be the same size as the Manuscript full text and Manuscript syllabized full text widgets below in the Admin chant detail page.

image

lucasmarchd01:
Fixes #1065. Matches the size of the other larger text area input widgets.

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -556,6 +556,9 @@ class AdminChantForm(forms.ModelForm):
         'the first word of each chant, and the first word after "Alleluia" for '
         "Mass Alleluias. Punctuation is omitted.",
     )
+    # Django's default text area widget selection for form inputs is non-intuitive
+    # and manual updates to fields (e.g., changing required=True) affect widget properties unexpectedly;
+    # this workaround is our current best solution.
     manuscript_full_text_std_spelling.widget.attrs.update(
         {"style": "width: 610px; height: 170px;"}
     )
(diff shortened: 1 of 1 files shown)


jacobdgm (changes requested):
Does this change work even when the window is narrow, for example? If all the other similar fields look one way and this one looks different, I'd prefer we figure out what the difference is rather than hard-coding the formatting to match the other fields.

lucasmarchd01:
I've discovered that Django's default widget selection for text area inputs in forms isn't straightforward. Manual updates to fields (such as changing required=True) also changes widget properties and behaves strangely. We have come to the conclusion that this is the best solution for now.

django/cantusdb_project/main_app/forms.py
```diff
@@ -556,6 +556,9 @@ class Meta:
         'the first word of each chant, and the first word after "Alleluia" for '
         "Mass Alleluias. Punctuation is omitted.",
     )
+    manuscript_full_text_std_spelling.widget.attrs.update(
+        {"style": "width: 610px; height: 170px;"}
+    )
```
jacobdgm, line 564: > I've discovered that Django's default widget selection for text area inputs in forms isn't straightforward. Manual updates to fields (such as changing required=True) also changes widget properties and behaves strangely. We have come to the conclusion that this is the best solution for now.

Can you add a comment in the code to this effect?

jacobdgm (approved):

PR #1073, opened 2023-09-22
Add `add_cantus_ultimus_links` management command

### LINKED ISSUE #1072: We need a workflow for attaching CantusUltimus links to records
a subset of #940, and a slightly simpler problem, because we know that URLs on CantusUltimus correspond to URLs on CantusDB in a consistent way.

This can probably be put together as a management command - whenever a manuscript is "published" on CU, we can run this command on the corresponding source in CantusDB and it will update the image links for all the chants in the source.

I'll get to work on this presently - should not be very difficult.

jacobdgm:
Closes #1072

In action, it looks like this:

```
root@3898da677ca2:/code/django/cantusdb_project# python manage.py add_cantus_ultimus_links 123684

Updating image link for source: [D-Mbs Clm 4306] München, Bayerische Staatsbibliothek, Clm 4306 (123684)
Source image link updated successfully!

Updating image links for 2050 chants...
  just finished chant 100
  just finished chant 200
  just finished chant 300
  just finished chant 400
  just finished chant 500
  just finished chant 600
  just finished chant 700
  just finished chant 800
  just finished chant 900
  just finished chant 1000
  just finished chant 1100
  just finished chant 1200
  just finished chant 1300
  just finished chant 1400
  just finished chant 1500
  just finished chant 1600
  just finished chant 1700
  just finished chant 1800
  just finished chant 1900
  just finished chant 2000
All chant image links updated successfully!
```

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/add_cantus_ultimus_links.py
+++ b/django/cantusdb_project/main_app/management/commands/add_cantus_ultimus_links.py
@@ -1,0 +1,63 @@
+from main_app.models import Source
+from django.core.management.base import BaseCommand
+
+# This management command goes through all the chants belonging to the specified manuscript
+# and adds a link to the corresponding page on Cantus Ultimus where a user can view the folio
+# of that manuscript. If the source is available on Cantus Ultimus, this script assumes its
+# ID there is the same as its ID on CantusDB.
+
+# This command only works for sources containing chants (i.e. sources in the CANTUS segment,
+# and not sources in the Bower segment which contain sequence objects as opposed to chant
+# objects).
+
+# run with `python manage.py add_cantus_ultimus_links <source_id>`
+
+
+def update_image_link_for_chant(chant):
+    folio = chant.folio
+    number = chant.c_sequence
+    source = chant.source.id
+    cantus_ultimus_link = (
+        f"https://cantus.simssa.ca/manuscript/{source}/?folio={folio}&chant={number}"
+    )
+    chant.image_link = cantus_ultimus_link
+    chant.save()
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "source_id",
+            type=int,
+            help="This command attaches Cantus Ultimus links to chants in the source with the specified ID",
+        )
+
+    def handle(self, *args, **kwargs):
+        source_id = kwargs["source_id"]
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/management/commands/add_cantus_ultimus_links.py
```diff
@@ -0,0 +1,63 @@
...
+# and not sources in the Bower segment which contain sequence objects as opposed to chant
+# objects).
+
+# run with `python manage.py add_cantus_ultimus_links <source_id>`
+
+
+def update_image_link_for_chant(chant):
+    folio = chant.folio
+    number = chant.c_sequence
+    source = chant.source.id
```
dchiller, line 19: There would be no place where this was different than the value in line 36, right? In other words, this will be the same for all chants in a single run of this command...

dchiller (approved):

django/cantusdb_project/main_app/management/commands/add_cantus_ultimus_links.py
dchiller: Approved...since I think this value will always be the same, I think a refactor would be clearer, but not critical.

dchiller:
Oh, also I think `stdout.write` is preferred to print statements in this case.

django/cantusdb_project/main_app/management/commands/add_cantus_ultimus_links.py
jacobdgm: Yes, `source` will always be the same for a given run of the command. This thought occurred to me (mostly from the point of view of optimizing the function to run more quickly), but I erred on the side of writing a function with fewer parameters - shaving a few seconds off the runtime is not that important for a command that I imagine will be run once every few weeks at most.

jacobdgm:
ooh, not only did I just learn about `self.stdout.write`, but `self.style.SUCCESS` and `self.style.ERROR` add all sorts (i.e. mostly green and potentially but hopefully not red) of pretty colours to the terminal. This management command will be much more enjoyable to run now :)

Good catch, @dchiller!

ahankinson:
> Oh, also I think stdout.write is preferred to print statements in this case.

The `logging` module is an even better way of doing it, since you can redirect to a file, control formatting, etc.

PR #1018, opened 2023-08-30
Create broken link checker cron jobs

zhannaklimanova:
Broken link checker cron job that will run at 08h08 every day and check for broken links on flatpages and created articles obtained from https://cantusdatabase.org/flatpages-list/ and https://cantusdatabase.org/articles-list/ websites.

PR #1093, opened 2023-10-20
Add updated google analytics tag to base.html

### LINKED ISSUE #520: Set up Google Analytics
In an email from Debra, she said:

> We are hoping to have Google Analytics on the site that is on the DRA server.
> 
> At this point we have the old GA on the UW server, but since this is changing we need to update to GA 4 anyways.
> 
> We have started to set up the property in our cantusdatabase@gmail.com account on our side, so let me know what to do next! (Currently it's registered with the wrong domain .... so maybe we need to start over.)

This is an issue to be solved only once we are fully production-ready.

jacobdgm:
This PR adds updated GA4 tags to all our pages (which will only be displayed when PROJECT_ENVIRONMENT is PRODUCTION). In doing so, it resolves #520.

This PR formalizes changes already made on our production server (i.e. the changes made in this PR were made as a hotfix on Production, where the changed code is already running), which @ahankinson and I set up together yesterday. All indications are that it is already working as expected.

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -5,15 +5,14 @@
 
 <head>
     {% if PROJECT_ENVIRONMENT == "PRODUCTION" %}
-        {# To ensure site is tracked in Google Analytics #}
         <!-- Google tag (gtag.js) -->
-        <script async src="https://www.googletagmanager.com/gtag/js?id=G-YSGM9EQBGL"></script>
+        <script async src="https://www.googletagmanager.com/gtag/js?id=G-JXEY1VRT2W"></script>
         <script>
             window.dataLayer = window.dataLayer || [];
             function gtag(){dataLayer.push(arguments);}
             gtag('js', new Date());
-            
-            gtag('config', 'G-YSGM9EQBGL');
+
+            gtag('config', 'G-JXEY1VRT2W');
         </script>
     {% endif %}
     {% if not PROJECT_ENVIRONMENT == "PRODUCTION" %}


dchiller:
I trust you and Andrew, so happy to approve, but not sure if the events of this weekend have implications for this...

ahankinson:
What are the events of this weekend?

dchiller (approved):

dchiller:
> What are the events of this weekend?

I will describe them because I mentioned them but upon further thought it's not actually relevant to this (so, side note, I'm approving). 

On Friday night/Saturday morning, Jan got in touch saying that he had to block Cantus DB from making requests to Cantus Index because of an unusually high number of requests. The only place Cantus DB really interacts with the Index is when a chant page is viewed (e.g., https://cantusdatabase.org/chant/294915), when we make a request to an Index endpoint for the concordances in other databases for chants with the same Cantus ID. Since this only should happen when a user actually navigates to a chant detail page, the traffic level should be minimal. One guess for the spike in traffic (h/t @annamorphism ) was that Google was indexing the site and therefore visiting all chant detail pages, and initiating these requests to Cantus Index, at a high frequency. We ultimately "solved" the problem by shutting off the concordance fetching altogether. 

We want the analytics tag regardless, hence my change of heart to "this isn't relevant". We just need to solve the issue another way: figure out a better way to handle concordances, figure out a way to "turn off" concordances when it's an indexer or other bot accessing the page, or figure out, if this was caused by something else entirely, what cause it.

ahankinson:
There is also the option to script something to cache the results periodically. That way you can control the rate. The data doesn’t change often enough to necessitate an on the fly lookup, so a weekly refresh is probably enough.

Google doesn’t generally result in a big load like that, but other crawlers might. What do your logs say happened on Friday night?

ahankinson:
PS: https://goaccess.io is a useful and fairly easy log analysis tool.

dchiller:
> There is also the option to script something to cache the results periodically. That way you can control the rate. The data doesn’t change often enough to necessitate an on the fly lookup, so a weekly refresh is probably enough.

Definitely

> What do your logs say happened on Friday night?

Unfortunately, CantusDB does not (not ideal, I know) have any set-up to persist logs, so we are limited to logs in this case to the last time nginx was restarted (so here that's after a quick fix was made for this on Saturday). 

I'm looking now at the logs since it was restarted on Saturday (we know the request volume was elevated when things were restarted on Saturday, so we might be able to find something useful if behaviour before and after restart was similar. Something more quantitative to come, but I'm noticing a number of bots making requests (Amazonbot, Turnitinbot, Semrushbot, Bytespide). I'm also noticing a lot of requests coming in from `lychee` which upon further investigation is what was used to set up the link checker...wondering if that link checker is running somewhere? And if so if it is paging through every chant for some reason? To investigate.

dchiller:
More details from my investigations...

In the first ~3 hours after the change was made (again, to draw any conclusions we'd have to assume that behaviour was roughly similar in this period immediately following the fix as it was in the period leading up to the fix), CantusDB logged ~ 130,000 `200` requests.

The most common requester was "Bytespider" with 65,000 requests in this period. Only about 428 were to `/chant/<id>` pages (so once every 26 seconds or so). However, another 5,500 requests (once every 2 seconds) were to the oddly structured `/login/?next=%2Fchant%2F<id>` url. I believe this just returns the log-in page and so probably is not responsible for sending requests to the Cantus network.

The lychee link checker is responsible for 43,000 of the requests in this period. It's mostly requesting static pages (`/about`, `/genres`, `/indexers`, etc.) For example, it requested the `/genres` page about once every two seconds for the entire period. The one exception is that it requested `/chant/230882` a dozen or so times. I imagine this is mentioned in an article or another static page.

I think we start getting to a culprit when we look at a bot with fewer requests (~10,000 over the period), but that send 6,000 requests to `/chant/<id>/` urls (that's once every 1.3 seconds): Amazonbot.

Add to that DataForSeoBot, which sent another 2,500 to those urls (once per 4.5 seconds) and Semrushbot, which sent another 1,200 (once per 9 seconds), and you're averaging a request to Cantus Index for concordances every 1.1 seconds at least. 

Maybe others have a theory of why this just started to be a problem? And of course longer-term logs would be necessary to see if any of this is different than "normal". While there are other solutions in the concordance problem in the works, we may want to take steps to mitigate these anyway (e.g., why is our link-checker, if it is our link-checker, sending that many requests!?!)

jacobdgm:
Thanks for looking into this! Should we try closing #1018 at this point and see if that stops the lychee requests?

We should also find a way to make our logs persist. I'll maybe open an issue for this.

dchiller:
> Should we try closing https://github.com/DDMAL/CantusDB/pull/1018 at this point and see if that stops the lychee requests?

I think it probably should be closed (unless it is going to get merged at some point), but there have been no runs of the action in the past two months, so that's not responsible for the process we saw on Saturday.

dchiller:
Aha, here's what's causing the lychee thing, I think: https://github.com/zhannaklimanova/cron-test-2/actions

jacobdgm:
@zhannaklimanova

jacobdgm:
looking through https://github.com/zhannaklimanova/cron-test-2/actions, it appears that it's only checking flatpages and articles, and not the whole site. While i don't think it's necessary that this run every 15 minutes and it would be nice if it stopped, I don't think this is responsible for all the visits to `/chant/<id>` pages.

Following the nginx logs now, I'm seeing a lot of traffic from one bytespider.

PR #1104, opened 2023-10-26
Add differentiae database melodic transcriptions to chant detail page

### LINKED ISSUE #774: Chant Detail Page: OldCantus has a "Differentia Database" field not present in NewCantus
related to #693, related to #701

OldCantus:
Screenshot 2023-06-21 at 1 49 35 PM

NewCantus:
Screenshot 2023-06-21 at 1 49 24 PM

### LINKED ISSUE #1030: We should display Differentiae Database volpiano snippets
A follow-on to #1014 - I sent an email to Jan asking about what API CD uses to get volpiano snippets from Differentiae Database, and it turns out there's no API - Jan has an extra table in OldCantus where these snippets are stored. He sent us a database dump with all the information he has.

It seems that this table was created once and has rarely/never been touched since. I asked Debra, and it sounds like the data in DD doesn't change very often (for better or for worse - I understand @annamorphism may have some opinions on this).

I'm thinking the best approach could be to create a new `Differentia` model in our database to store volpiano snippets for display on the Chant Detail page. From time to time (perhaps once a year), we can sync data with DD, perhaps by asking them for a CSV.

The model would look something like this (based on what I can see at https://differentiaedatabase.ca/index):

```python
class Differentia(BaseModel):
    differentia_id = models.CharField()  # values like `129a` get stored here
    melodic_transcription = models.CharField()  # values like `1--k-k-l-j-k-h--4` get stored here
    # mode - differentia database stores information on mode, but we don't expect to ever need to display it
    # saeculorum - DD stores a volpiano snippet about the "saeculorum" part of the differentia, but we don't expect to ever need to display it
    # amen - DD stores a volpiano snippet about the "amen" part of the differentia, but we don't expect to ever need to display it
```

**Does creating a new model like this seem like a reasonable course of action?** In the absence of any dissenting voices within the next ~24h, I'll go ahead and start implementing this.

lucasmarchd01:
This PR adds the differentia database melodic transcriptions to the chant detail page and adds a new model to our database `Differenita`.

To do this, we have added a txt file that contains the data from [differentiaedatabase](https://differentiaedatabase.ca/). The added sync script creates a new Differentia objects that store the differentia database ID's along with their melodic transcription and modes. Then, the ID's that are currently stored in the `differentiae_database` field in the Chant model are mapped to the corresponding Differentia object with the same ID. A ForeignKey `diff_db` is added to the Chant model that stores the appropriate Differentia for that chant. 

Additionally, chant create, chant edit, and chant detail pages are modified to display the `diff_db` field instead of the `differentiae_database` field.

To populate these changes onto your local development branch / staging / production:
- perform database migrations
- run the sync script add_differentiae.py

image

Fixes #1030, Fixes #774


django/cantusdb_project/main_app/management/commands/add_differentiae.py
```diff
@@ -0,0 +1,89 @@
...
+# the differentia ID. Now, we change this field to a ForeignKey that points to the new Differentia model that contains
+# the ID, melodic transcription, and mode.
+
+
+DIFFERENTIA_DATA = "differentia_data.txt"
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        differentia_database = get_differentia_data(DIFFERENTIA_DATA)
```
jacobdgm: type annotations would be helpful, here and elsewhere!

django/cantusdb_project/main_app/management/commands/add_differentiae.py
```diff
@@ -0,0 +1,89 @@
...
+                differentia.melodic_transcription = melodic_transcription
+                differentia.mode = mode
+                differentia.save()
+            else:
+                differentia = Differentia(
+                    differentia_id=differentia_id,
+                    melodic_transcription=melodic_transcription,
+                    mode=mode,
+                )
+                differentia.save()
```
jacobdgm, line 36: I don't really understand why we need an if/else here... are there duplicate IDs in `differentia_data.txt`? In what situations is `differentia` expected to evaluate to `True`?

jacobdgm: Maybe, in the `if` block, an error message would be more appropriate than allowing this unexpected situation to pass silently and just overwrite?

django/cantusdb_project/main_app/management/commands/add_differentiae.py
```diff
@@ -0,0 +1,89 @@
...
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        differentia_database = get_differentia_data(DIFFERENTIA_DATA)
+        total = len(differentia_database)
+        count = 0
+        for item in differentia_database:
+            differentia_id, melodic_transcription, mode, _ = item
+            differentia = Differentia.objects.filter(
+                differentia_id=differentia_id
+            ).first()
```
jacobdgm: can we just use `Differentia.objects.get()` here? If we somehow end up with multiple differentiae with the same differentia_id, I'd want an exception to be raised, rather than just picking one arbitrarily as is currently done.

django/cantusdb_project/main_app/management/commands/add_differentiae.py
```diff
@@ -0,0 +1,89 @@
...
+        while start_index <= chants_count:
+            print("processing chunk with start_index of", start_index)
+            chunk = chants[start_index : start_index + CHUNK_SIZE]
+
+            # Update Chant model to use foreign key
+            for chant in chunk:
+                try:
+                    differentia_id = chant.differentiae_database
+                    # Check if differentiae_database is null
+                    if differentia_id:
```
jacobdgm: or, `if not differentia_id is None:`, and we don't need a comment!

Type annotations earlier may also help clarify why this check is needed.

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -129,14 +129,17 @@ <h3>{{ chant.incipit }}</h3>
...
                     {% endif %}
 
-                    {% if chant.differentiae_database %}
+                    {% if chant.diff_db %}
                         <div class="col">
                             <dt>Differentia&nbsp;Database</dt>
                             <dd>
-                                <a href="https://differentiaedatabase.ca/differentia/{{ chant.differentiae_database }}" target="_blank">
-                                    {{ chant.differentiae_database }}
+                                {% if chant.diff_db.melodic_transcription %}
```
jacobdgm, line 136: if a chant's `diff_db` is `None`, do we not get an error message when the template tries to display `None.melodic_transcription`? Not certain of this, but do make sure this works for a chant with no `diff_db`.

jacobdgm (changes requested):
This is pretty close to being approval-ready. Please do check on the possible bug in the template, and take a look at the comments I left in `add_differentiae.py`.

django/cantusdb_project/main_app/management/commands/add_differentiae.py
lucasmarchd01: This is checking if a `Differenita` object already exists with that ID. Say, if you ran the script more than once, you don't want duplicate objects in the database.

django/cantusdb_project/main_app/templates/chant_detail.html
lucasmarchd01: If `diff_db` is None this line will not be evaluated since line 132 already checks for this situation I believe

django/cantusdb_project/main_app/management/commands/add_differentiae.py
jacobdgm: got it - we might update `differentia_data.txt` and run the script again. Thanks.

jacobdgm (approved):

PR #1111, opened 2023-10-27
Create script to remap user IDs

### LINKED ISSUE #1000: strange user assignments?
As I was looking at the sources in #999 I noticed some VERY suspicious user-source relationships. I happen to know that the sources in Helsinki are being worked on by users 251662, 251663 and 251678--who I don't see at all in the recently modified sources list. Instead I see 251612 and 251613, who I didn't think had been logged in in some time, and aren't to my knowledge involved in that project, and all the Finnish sources seem to have been "last modified by" @jacobdgm. 

I find this odd...

jacobdgm:
Fixes #1000.

After several rounds of bugs, @annamorphism and I have looked at my local database following the running of this script, and everything we're seeing is as expected. We should be good to test this on staging!


dchiller (approved):
Look fine to me ... 

I'm sure the logic is sound since it's both @jacobdgm and @annamorphism approved, but I'll just ask because I any extra caution around changing ID's around is warranted: the reason we don't need to worry about chants and sources being `created_by` an Indexer (someone whose has a dummy account with the ID's in the keys of the dictionary here) is because sources and chants that involved indexers (ie. those entered in the before times) will have those users (the dummy users created for Indexers in NewCantus) credited in other fields (`inventoried_by`, `proofread_by`, etc.)?

jacobdgm:
I'm not sure I understand your question, @dchiller. To answer what I think you're asking: Indexers / dummy users cannot have created objects, because they were not Users in OldCantus, and their accounts cannot be logged into in NewCantus.

We created new user accounts for OldCantus indexers, at the same time as new (real) user accounts were created in OldCantus. Those (real) users made edits, but their account IDs overlapped with those of NewCantus's new indexer users. So when we synced, we matched chants/sources based on User IDs, which led to what should have been impossible - objects that were created by users whose accounts could not be logged into.

dchiller:
You answered it: essentially, no sources or chants should actually be "created by" the dummy user ID's...contributions by the Indexers represented by those dummy user's are recognized in other fields.

Issue #1098, opened 2023-10-21
we applied a hotfix to production for `touch_ci_json_con_api`


jacobdgm:
and we need to implement a proper fix. This issue under construction.

annamorphism:
is this the issue with the concordances? depending on the time frame of this fix, can we have the interim message display something like "the concordances feature is currently under construction" rather than "Cantus encountered an error...". One makes it look like something is wrong (I have had somebody report it to me) while the other is a little clearer that we know about it.

jacobdgm:
we can apply another hot fix if you think it's important. I hope we can get to this soon, so I might favour properly fixing this as soon as we can, rather than making more changes directly on the production machine.

annamorphism:
that's fine! I would rather have things properly fixed, but if the solution does end up being "not do concordances until things are sorted at Cantus Index" then we should signal that to users.

jacobdgm:
I think we'll be fine to fetch concordances from CI before too long (the cached ones - we definitely need a more robust solution for requesting refreshes).

Thinking about this over the "weekend", one component issue of the heavy traffic we were passing on to CI is that we were not waiting for CI to recalculate concordances before returning a response:
- if, when someone requested a chant detail page, we waited for CI to update its stored concordances before returning a response, then whatever bot was crawling our site would have had to move more slowly.
- with the approach we had implemented until recently, however, we returned a response much more quickly than it took for the CI databases to process their API requests. This allowed the crawler to quickly visit many pages, causing all the requests to CI to pile up.

For requesting updated concordances, I think there are two approaches that could work similarly well:
- First is that we implement a "refresh concordances" button that sends a request to CI's /refresh endpoint, waits for a response, and then reloads the page (or, dynamically updates the content in the table). Debra said this is less than ideal, so maybe let's consider a second approach.
- Second, we keep a local copy of concordances for all CantusIDs, which we automatically update from time to time. If we want to do this, we should get to work on #410. I'm going to bump the priority on this one from low to high.

jacobdgm:
Some notes to myself: I've been thinking that, as an initial step, it would be a good idea to configure nginx to do some rate limiting (i.e., throttling) of requests from any particular IP address. Looking at the network requests while loading the homepage, I see 13 requests made for resources (e.g., images etc.) hosted on cantusdatabase.org. So we'd need to ensure that bursts of traffic are allowed - based on the number of requests I observed, 20-25 could be a good number of requests to allow in a burst.

[...]

PR #1115, opened 2023-10-31
Attempt to implement rate limiting via nginx

jacobdgm:
This draft PR is an attempt to implement rate limiting via nginx, toward addressing #1098. In its current form, it doesn't work: I got this error message - 
```
2023-10-31 13:07:04 2023/10/31 17:07:04 [emerg] 1#1: "limit_req_zone" directive is not allowed here in /etc/nginx/conf.d/cantusdb.conf:5
2023-10-31 13:07:04 nginx: [emerg] "limit_req_zone" directive is not allowed here in /etc/nginx/conf.d/cantusdb.conf:5
2023-10-31 13:07:17 2023/10/31 17:07:17 [emerg] 1#1: "limit_req_zone" directive is not allowed here in /etc/nginx/conf.d/cantusdb.conf:5
2023-10-31 13:07:17 nginx: [emerg] "limit_req_zone" directive is not allowed here in /etc/nginx/conf.d/cantusdb.conf:5
```
\- when I tried to test this with a modified-for-local-development version of this (The first `server` block was deleted and the second server block began with `listen 80;` instead of `listen 443 etc`. `limit_req_zone $binary_remote_addr zone=default:10m rate=2r/s;` was the line identified in the error message.)

ChatGPT suggested that I wrap the entire file in an `http` block, and move `limit_req_zone` outside of the server block, but I don't know enough about nginx to know whether this is a reasonable idea or something that will cause problems.

People who know nginx (@kqct? @ahankinson? @alastair?), would one of you be able to offer some guidance as to how to proceed?

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -19,7 +19,10 @@ server {
     ssl_certificate /etc/nginx/ssl/live/cantusdatabase.org/fullchain.pem;
     ssl_certificate_key /etc/nginx/ssl/live/cantusdatabase.org/privkey.pem;
 
+    limit_req_zone $binary_remote_addr zone=default:10m rate=2r/s;
+
     location / {
+        limit_req zone=default burst=30;
         proxy_pass http://django:8000;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header Host $host;


ahankinson:
What’s the reason for doing rate limiting? Generally this is used for API systems or things where programmatic access is the primary mode of use. (And popular websites that might be the target of DDoS)

Are you doing this to limit crawlers? If so, look at beefing up your robot.txt. Or block the bad user agent / offending IP address outright. 

rate limiting is a tricky thing to get right so you should exhaust other options first.

jacobdgm:
> What’s the reason for doing rate limiting?

So in the days/hours leading up to the project meeting, some agent was visiting many chant detail pages on the website in very quick succession (~10r/s). Whenever someone requests a chant detail page, we ask for information on concordances from CantusIndex, and asynchronously we ask CI to update its cached information on concordances. Since we were able to return responses faster than CI and related CantusNetwork sites could recalculate lists of concordances, some of these other sites were sagging under the traffic that was being generated.

Jan sent us an angry email telling us to stop our "aggressive bot" and telling us that what we were doing was immoral, so we applied a couple of hotfixes to simply not request any information from CI. Thus, several features of the site that depend on CI are not working at the moment.

We're not sure what caused all the requests to CD, or why the approach that we had been using since the launch of NewCantus suddenly began causing problems two weeks ago. Initially, we thought that a Google crawler was causing all the additional traffic (something having to do with us having just set up Google Analytics). And over the summer, there were several experiments to periodically check all the pages on the site for broken links, which the developer forgot to turn off. But looking in more detail into these, neither of them seems to explain the heavy traffic Jan was reporting. When we applied the hotfix, we lost the logs nginx had generated up to that point. So we don't know who or what was causing all this traffic, only that it was happening.

We're working on implementing a solution where we can send a request to a single endpoint on CI every few hours and cache the response, but that's not ready yet, either on CD's end or on CI's. For the time being, however, any updates to Production will necessitate rolling back the hotfixes, so we're hoping to find a temporary solution that would allow us to start displaying concordances again while ensuring we don't send too many requests to CI.

> rate limiting is a tricky thing to get right so you should exhaust other options first.

I think all we need is something that blocks the most egregious high-traffic situations, rather than anything fine-tuned. Maybe setting our rate to 5r/s would be more appropriate than 2r/s? (We know that loading a single page necessitates ~15 requests for images, css, etc., hence setting the `burst` to 30.)

jacobdgm:
It sounds like the general sentiment is that this would not be a good idea. Closing.

ahankinson:
I think you need to know what the problem is before trying to fix it.

10r/s is not high for sites. You should definitely consider that “normal” traffic and engineer for that.

It sounds like you’re working on keeping your nginx logs and analyzing them. I think that’s the first course of action. Use a tool like goaccess.

jacobdgm:
I agree, though ideally we can anticipate problems and proactively work to prevent them, rather than playing whack-a-mole. To me, rate limiting feels like a proactive solution, but if you think it is likely to create more problems than it will solve, I believe you.

We appreciate you sharing your instincts/hunches surrounding this - many thanks!

ahankinson:
>~~instincts/hunches~~ experience 

I tried rate limiting for DIAMM. It was difficult to configure, and difficult to verify it was actually working.

PR #1121, opened 2023-11-02
Change viewing permissions for proofread fields on chant detail page

### LINKED ISSUE #1100: feature request: holding fields until proofread
At our meeting last week @lucasmarchd01 suggested (if I remember right) that chants could become "unpublished" when they have been modified, and published again once they have been proofread. In that discussion it was decided that could become unwieldy (e.g. if every time a typo were fixed, the chant would have to be re-proofread, things could get out of hand quickly.)

After some discussion, though, it turns out that a variation on this idea would be really useful. Here's the problem:
A chant might originally have been created, proofread, and published, but only for (e.g.) Standardized spelling. Later, a new project adds MS Spellings and volpiano melodies to their source. As it is, these new additions will show up in the published chant, but they have not been proofread to the same level as the others. 
Is there a way that we could leave the proofread _fields_ published/viewable, but hide some fields (like volpiano) until they have been proofread?

lucasmarchd01:
On the chant detail page, we make several modifications to field viewing permissions based on if they have been proofread or not. Specifically:
- logged in users will be able to see proofread and non-proofread fields
- for logged in users non-proofread fields, we will show a message indicating they are not yet proofread
- non-logged in users will only be able to see fields if they have been proofread.

Resolves #1100 (special debra request!)

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -167,23 +167,69 @@
                     {% endif %}
                 </div>
 
-                {% if chant.manuscript_full_text_std_spelling %}
-                    <dt>Manuscript Reading Full Text (standardized spelling)</dt>
-                    <dd>{{ chant.manuscript_full_text_std_spelling }}</dd>
+                {% if user.is_authenticated %}
+                    {% if chant.manuscript_full_text_std_spelling %}
+                        {% if chant.manuscript_full_text_std_proofread %}
+                            <dt>Manuscript Reading Full Text (standardized spelling)</dt>
+                            <dd>{{ chant.manuscript_full_text_std_spelling }}</dd>
+                        {% else %}
+                            <dt>Manuscript Reading Full Text (standardized spelling)
+                                <span style="font-weight:normal"><i> (not proofread)</i></span>
+                            </dt>
+                            <dd>{{ chant.manuscript_full_text_std_spelling }}</dd>
+                        {% endif %}
+                    {% endif %}
+                {% else %}
+                    {% if chant.manuscript_full_text_std_spelling and chant.manuscript_full_text_std_proofread %}
+                        <dt>Manuscript Reading Full Text (standardized spelling)</dt>
+                        <dd>{{ chant.manuscript_full_text_std_spelling }}</dd>
+                    {% endif %}
                 {% endif %}
 
-                {% if chant.manuscript_full_text %}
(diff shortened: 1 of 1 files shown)


jacobdgm (approved):
There are currently three sets of `{% if user.is_authenticated %} ... {% else %} ... {% endif %}`...

I'm approving because this is fine, and it's possible that this is clearest way to arrange things. But I'm guessing that things would be clearest if all three fields were wrapped in a single `{% if user.is_authenticated %} ... {% else %}` block.

Please consider this before merging. If you decide to do this, please re-request my review.

jacobdgm:
Current line of thinking: **I'm going to roll this change back** to clear the way for #1146, since @annamorphism and I are discovering a bunch of edge cases - I had initially thought this would be a simple change, but it's proving not to be as simple as I thought when I gave my approval.

First off, there is the issue of making sure that the melody information on the Chant Search page matches the melody information on the Detail page: #1147.

Second, there's the issue of having to handle the text-and-melody preview: #1148.

While I was working to address the second of these, it occurred to me that we currently have a series of fallbacks for choosing which text to use as part of the text-and-melody preview, and these fallbacks don't consider whether a given field is proofread or not. At the very least, we'll need to rework `Chant`'s `get_best_text_for_syllabizing` method to account for whether the requesting user is logged in. (We don't expect there to be many situations where volpiano is proofread but the MS Spelling fulltext isn't, but we need to account for this situation)

@annamorphism also pointed out that there will exist many old fulltexts/volpianos in the database which are practically proofread, but which aren't marked as proofread (i.e. the option to check this box didn't exist when the proofreader went through and verified them). We may decide to simply hide them from anonymous users because they aren't marked as proofread, but this also merits more thought. Perhaps @annamorphism can weigh in and expand on this a bit.

So in summary, I'm planning on rolling this back to allow other changes in #1146 to go to production.

(When we re-implement this, it might be worth putting the logic in the view, and passing this to the template as context variables, e.g. `{% if display_ms_fulltext %}` `{% if display_volpiano %}`, `{% if display_text_melody_preview %}`, and so on.)

annamorphism:
These examples of edge cases are somewhat related--most of the edge cases about non-proofread texts and melody preview come about because of old workflows where texts weren't marked proofread (e.g. it was proofread in some spreadsheet offline) or the manuscript only has incipits (which wouldn't be possible now). Putting the logic in the view feels like  it could be a good approach--since then you wouldn't have to change which text is the "best text," but only decide under what conditions it gets shown. But then, making sure those conditions were right might also take some thought...I will look into what the possible cases are!

PR #1141, opened 2023-11-09
Add `CantusIdentifier` model

### LINKED ISSUE #410: We could create new CantusIdentifier model
I suggest that we add a new model named `CantusIdentifier`. This new model would address two issues:

First, "a way to check that all the genres in the Cantus Database are identical to the Cantus Index Cantus ID # entries" has been requested in the wish list (further: "Sometimes errors slip in and human eyes don't see that the genre has been entered incorrectly. (I.e., Can ID 202838 should always have the genre "A". Each Can ID # should have only 1 genre assigned with it, and tell me if you find any that do not match so I can fix them).") This type of information could conveniently be stored in a CantusIdentifier object.

Second, it would help optimize some things related to the local storage of concordances retrieved from Cantus Index etc. (as discussed in #405). For some Cantus IDs, we have dozens of chants associated with that ID. By having a single object associated with each individual ID, we can store the concordance only once for each ID (rather than dozens of times; this is only a small concern, though, because it doesn't take a lot of data to store concordances), and when we update our stored concordances, we only need to send a single request to Cantus Index (rather than dozens).

lucasmarchd01:
Here, we are setting up a `CantusIdentifier` model that can be used to store concordances and CI information. Here are several reasons why we are making this change, _Originally posted by @jacobdgm in https://github.com/DDMAL/CantusDB/issues/410#issuecomment-1731370051_

- CantusIDs would be a controlled vocabulary - when creating/editing chants, users would have to select from a list of existing CantusIDs instead of entering one freehand (thus limiting the possibility of typos)
  - we can also ensure that when a user selects a Cantus ID, the chant's Genre is automatically set to the genre associated with the CantusID
  - we would need to talk with Debra about how new CantusIDs are created - if a user is indexing a manuscript, will they ever need to invent a new CantusID? Would such a CantusID need to be added to CantusIndex first before it can be selected in CantusDB
- It would be easy/efficient to cache concordances from CantusIndex, since these concordances belong to specific CantusIDs and not to specific chants. This would also allow us to continue to display concordances even in case of a temporary outage on CantusIndex
- For the Input Tool, this would allow us to store canonical texts and genres for all CantusIDs
- For the previous two points, performance would probably also be improved, because we won't need to send requests to CantusIndex in order to load our pages - we'd just need to query our own database.

Resolves #410


jacobdgm (approved):
This looks good! Thank you!

django/cantusdb_project/main_app/models/cantus_identifier.py
```diff
@@ -0,0 +1,18 @@
+from main_app.models import BaseModel
+from django.db import models
+
+
+class CantusIdentifier(BaseModel):
+    cantus_id = models.CharField(blank=False, null=False, max_length=255)
```
ahankinson: - `blank` and `null` are False by default, so you don't need to specify them. 
- 255 chars for an identifier is pretty long, no?

jacobdgm: > `blank` and `null` are False by default, so you don't need to specify them.

This is a good point. We've got this boilerplate in a bunch of places in our models, iirc - might be worth cleaning up at some point.

> 255 chars for an identifier is pretty long, no?

This is a good point - @lucasmarchd01, 15=(2**4)-1 might be a better number?

jacobdgm:
The thought occurs to me: What are we going to do with chants that have an erroneous Cantus ID?

When we switch the `cantus_id` field from a CharField to a ForeignKeyField, we expect nearly all the chants to have Cantus IDs that are present in CI, but I imagine there will be several that are typos. So as not to lose that data, we'll have to create new CantusIdentifier objects to accommodate these, but we should be able to easily identify them so that chant specialists can go in and fix them at some point. Also, these erroneous Cantus IDs should probably not show up as suggestions in our controlled vocabulary when creating/editing chants.

Should we create a BooleanField called `present_in_ci`, to mark whether the object was created when syncing with CI or whether we created it internally? Can anyone think of a better way to approach this?

jacobdgm:
Also, would we want a `ci_uri` class method for the model, which returns the page for the Cantus ID on CI? I don't imagine us using it in views, but it could be handy for developer use, for writing tests, etc.

annamorphism:
> When we switch the `cantus_id` field from a CharField to a ForeignKeyField, we expect nearly all the chants to have Cantus IDs that are present in CI, but I imagine there will be several that are typos. 

We also know (from a recent Debra email) that there will be a few correct IDs that are not in CI, namely those for unknowable/unidentifiable texts.
Also, out of curiosity--how does this handle cantusIDs that are in CI but currently not attached to a chant in CantusDB?

annamorphism:
>       * we would need to talk with Debra about how new CantusIDs are created - if a user is indexing a manuscript, will they ever need to invent a new CantusID? Would such a CantusID need to be added to CantusIndex first before it can be selected in CantusDB

Yes, users do need to create new CantusIDs from time to time. These should be added to CantusIndex first, which is a better authority for which numbers have already been assigned and how certain IDs relate to other IDs (variants, related chants, etc.). We do NOT want to be creating IDs in both places at once (this happens for feasts currently and is a real pain.)
>

jacobdgm:
> Also, out of curiosity--how does this handle cantusIDs that are in CI but currently not attached to a chant in CantusDB?

We haven't set up any syncing between the two databases yet, but how I imagine it working:
- if a Cantus ID is in CD but not yet on CI, the Cantus ID will get its own, mostly-empty, CantusIdentifier object without any concordances, fulltext, etc.
- if a Cantus ID is on CI but doesn't have any associated chants in CD, a CantusIdentifier object will be created, which won't be associated with any chants. It will show up in the Cantus ID selector on Chant Edit/Create pages.
- if a Cantus ID is on both sites (e.g. if it was initially created in CD and only later makes its way to CI), then whenever we do our periodic sync with CI, its information (concordances/fulltext/etc.) will be overwritten with whatever CI gives us

ahankinson:
I'm wondering if creating a full model for this is going quite far in the opposite direction of what is needed. Before, all the info was coming from CI on-the-fly; now it seems you're effectively creating your own CI model structure and formally integrating it into your data models. 

This means the lines between the CI authority, and your own local authority, start getting quite fuzzy. Which is the "source of truth" here -- your models, or their data? If you start making FK relationships between models, you're starting to assert a very strong connection with your own data models. 

This can be solved by simply *caching* CI info locally. Functionally, a cache has the ability to be completely wiped and re-created at any moment. That means that there are no hard links to your data models, but your cache simply replaces the HTTP call to CI with something local. 

So what about this?

 - Create a "CICache" model with two fields: `path`  and `json`. For every URL path in CI you need to call, instead just do a lookup for that path in your model, and save the JSON that it returns. The function of your cron job is to gather the URLs that need to be cached, and requests them at a rate that makes Jan happy. 
- If something happens and you don't have the path cached (e.g., someone adds a reference to a new cantus ID number), then request it from CI and store the response in your cache. Subsequent lookups will then use the cached value, and bypass the HTTP request. 
- If something happens and things go completely squirrely, wipe all entries in the CICache table and re-fetch the data. Again, use a request rate (1 r/s?) that doesn't overwhelm CI. 
- Or, rather than periodically refreshing with cron, you could store a `created` timestamp on your CICache entries, and simply re-fetch the data from the endpoint once a certain window expires. Only re-fetch the whole thing with cron if you've invalidated all caches.
- If someone puts in a chant ID that doesn't exist in CI (the request for info from CI returns an error), then politely refuse to store it and direct them to CI to create the chant they need.  (This would jive with @annamorphism 's take). This should only happen to CD indexers, who should know better, and will have zero impact on the general users of the database.
- As a bonus, you could use this cache for other sites as well. Add a third field, `data_source` with an indicator of the database that you want to get the data from. Then when you want to look up the data from another source, you've already got that locally too. 
- As an added benefit, if you store the data as a `JSONField` you might also be able to [query it directly](https://docs.djangoproject.com/en/4.2/topics/db/queries/#querying-jsonfield) should you need to. Not sure what use case that would cover, but it may be handy to know that it's possible.

Thoughts?

---

A few red flags are raised with @jacobdgm 's responses.

> its information (concordances/fulltext/etc.) will be overwritten with whatever CI gives us

Overwriting your data from another source if it's actually FK linked sounds like a Really Bad Idea™ to me. You can think in the best case that it will work, but what about the non-best-case? What if CI gets hacked and records start getting deleted? Will this start automatically removing chant relationships from Cantus? How will they be restored if they're deleted? 

> if a Cantus ID is in CD but not yet on CI, the Cantus ID will get its own, mostly-empty, CantusIdentifier object without any concordances, fulltext, etc.

Since you need to supply a value for `cantus_id` this means that you are, in effect, creating your own authority. Don't underestimate the ability for someone to add their own value just to get the thing to save, and then forget about it completely. Trying to unpick this later can be a real nightmare. 

> if a Cantus ID is on CI but doesn't have any associated chants in CD, a CantusIdentifier object will be created,

This means you're basically creating a mirror of CI, but with the ability of adding records to your own local copy that don't exist in the remote. Will any of the "local" records get deleted if they are re-assigned to a CI chant? Probably not. So over time you're going to get a lot of cruft in your `CantusIdentifiers` table.

jacobdgm:
Thanks for these thoughts, @ahankinson! A couple of thoughts:

- I think I was unclear about the situations where I imagined CD and CI diverging and how to handle them. The only time when this should happen is when we do the initial "launch" of the model. There will be (i.e., there are) some things in the current `cantus_id` field that don't exist on CI, so if we implement a CantusIdentifier model, we would have to create CantusIdentifier objects to accommodate these values. My thought was that these CantusIdentifier objects would not be modifiable/creatable by any CD user, so following the initial launch, all data for CantusIdentifier objects will come from CI (with a few "legacy" objects, which will be flagged so chant researchers can go through and assign those chants to "real" Cantus IDs, and the legacy objects can be removed over time).
  - in other words, the main purpose of the new model would be to store information we've received from CI, with a secondary purpose of creating a controlled vocabulary (which we could enforce in other ways, but having a model allows us to enforce this at the database level, which will be simple and robust). Creating extra objects to accommodate existing non-conforming values will be necessary, even as it's not ideal: In a perfect world, all the CantusIdentifier objects would have corresponding Cantus IDs on CI, along with their own set of concordances and fulltext.
- On the whole, the method you outlined feels pretty similar to the one I initially envisioned (i.e., initially I thought we'd be working with the CI APIs as they exist currently), except that we unpack the responses from CI at request time (your suggestion) vs. cache time (my suggestion).
  - This changes now, however, since it sounds like Jan is well along in his work creating a single API that will return a list of concordances + other relevant data for all the Cantus IDs at once. There's nothing stopping us from parsing this endpoint's JSON at request time, but I feel like doing this would be inelegant.
- Another thought about caching requests: there are places in our code where we fetch data from CI, but in a really hacky, workaround sort of way. I'm thinking in particular of our [CI Search view](https://github.com/DDMAL/CantusDB/commit/7b18c99a911bb0aac88886842af1ef011d1c6359#diff-6656acfd99a9582e0f9947e6c38fc8e833ac78f621c88a748c01f0964613935cL1065) (recently removed), where we would send a request to https://cantusindex.org/search and then parse the HTML we got back. This is part of the reason I welcomed the news when Jan suggested a new API, which we now expect to include the information we have been scraping from CI's Search page.

In summary, I like your suggestion, and think that it would have made a lot of sense at one point. But given we are at the place we are right now (i.e., Jan nearly finished an API that will return information on all Cantus IDs at once, and us still lacking access to a necessary API for one of our views), we're going to need to work with what CI provides us, so I think it will make more sense for us to organize the data we're going to get from CI's new API in a meaningful way at cache time.

jacobdgm:
> Overwriting your data from another source if it's actually FK linked sounds like a Really Bad Idea™ to me. You can think in the best case that it will work, but what about the non-best-case? What if CI gets hacked and records start getting deleted? Will this start automatically removing chant relationships from Cantus? How will they be restored if they're deleted?

I was unclear. If CantusIndex gives us new information for a Cantus ID, we trust that information to be accurate and update our local object. If CantusIndex doesn't give us information about a Cantus ID, our CantusIdentifier would sit there, unmodified (as would happen for a placeholder object for a non-conforming Cantus ID, or if the Cantus ID was removed from CantusIndex). I didn't mean to imply, if an ID was removed from CantusIndex, that we would delete our own CantusIdentifier object.

This is an important thing to consider, though. We'll need to keep it in mind while implementing #877 (and this will be true whether we use CantusIdentifier objects or the current `cantus_id` CharField).

jacobdgm:
Another thought occurred to me: The approach taken in this PR is similar to the approach we took in #1104 - we're creating a model to store data taken from another database. For the `Differentia` model, caching simply wasn't an option - we got the data from a spreadsheet Jan gave us, and we had to organize the data in a way that was logical and accessible.

If we decide that a `CantusIdentifier` model is not the right approach, to what extent does this mean we need to reconsider our `Differentia` model as well?

(In hindsight, for #1104: @lucasmarchd01, is it possible that we will lose certain `diff_db` values when we mapped chants onto our new Differentia model and then deleted the value in the old field? At a glance, I'm not seeing a provision for `diff_db` values that don't have a corresponding `Differentia` object, other than a message that would print to the terminal. I don't recall seeing such a message printed to the terminal when I ran this locally...)

ahankinson:
Thanks @jacobdgm ! That helps me understand a bit better. I realize that I didn't quite understand everything, but I still have a few doubts about the way this is currently implemented. 

- I remain sceptical about this approach: "There will be (i.e., there are) some things in the current `cantus_id` field that don't exist on CI, so if we implement a `CantusIdentifier` model, we would have to create `CantusIdentifier` objects to accommodate these values."  I would definitely not make CI models for "bad" data in CD. Flag them for review, put them in a log, and pass them off to a person to verify and fix. But creating "fake" data just to make the process work, even if only once, sounds wrong. Do you know how widespread this problem is?
- "except that we unpack the responses from CI at request time (your suggestion) vs. cache time (my suggestion)" and "If CantusIndex gives us new information for a Cantus ID, we trust that information to be accurate and update our local object." I guess this is what got me thinking about it more like a cache, and less like a local model. In a cache, the "upstream" will change, and may actually change quite often. So I see the current method (what you're calling "cache time") more like an "import", primarily because parsing the data starts to couple the two data sources together. For example, adding a m2m key between the CI "cache" and the CD "genres". If it were really a cache, then the incoming data would stand entirely on its own. 
- If you're going to be able to get *all* data from CI in one request, then I think it makes even more sense to simply cache the response locally. I'm not sure how you're planning your request structure to get the CI data into your views, but I'm guessing it will require either a) a hardcoded FK between something in your CD structure to the `CantusIdentifier` table, or b) a dedicated query to the `CantusIdentifer` model for every mention of a `cantus_id`. If the former, I don't think that's a good idea -- tying CD data to CI data directly. If the latter, that's an even worse idea since it will result in an explosion of SQL queries. But, if you cache the result as a single blob, then it's just one query to get the blob from a JSONField -- then it's a Python dictionary. That's pretty easy to parse.

jacobdgm:
> I would definitely not make CI models for "bad" data in CD. Flag them for review, put them in a log, and pass them off to a person to verify and fix. But creating "fake" data just to make the process work, even if only once, sounds wrong. Do you know how widespread this problem is?

This is a good point, and I'm inclined to agree that it would be better to review and correct the data first. In a random sample of 100 cantus IDs from the database, 98 of them have pages on CI and 2 don't. We have ~16000 unique Cantus IDs in the database, so we can expect ~300 non-conforming Cantus IDs to exist (with very wide error bars). Guessing this would affect maybe up to 1000 chants, and possibly a fraction of that number (assuming that my small sample was representative, and that the non-conforming Cantus IDs are all one-offs)

Let me look into this more.

> But, if you cache the result as a single blob, then it's just one query to get the blob from a JSONField -- then it's a Python dictionary. That's pretty easy to parse.

One possible concern: this dictionary will have about 10<sup>4</sup> items (one for each Cantus ID), each of which will have about 10<sup>2</sup> concordances. I'm guessing, but I'm not certain, that our VM will not have issues with memory if we have a dictionary containing 10<sup>6</sup> concordances - are we safe to assume this?
- I know that we currently need to break up our 5*10<sup>5</sup> chants in some of our management scripts ([example](https://github.com/DDMAL/CantusDB/blob/871ef9e3873f0c8ac357a06dcb619c8b7dd4c39d/django/cantusdb_project/main_app/management/commands/populate_chant_fields.py#L26)), process them in batches, and delete the batches from memory for them to run to completion. It's not entirely clear to me why deleting these from memory causes these to work, but it suggests this is a thing to bear in mind)
  - update: thinking back, upon running such scripts, the python interpreter in our container would print "Killed" to the terminal and terminate the process after a length of time. So maybe this was a time thing rather than a memory thing.

jacobdgm:
> primarily because parsing the data starts to couple the two data sources together

You've expressed something along these lines a few times, and I don't fully understand your concern. Do you mean to say that this is a bad idea, or just that we need to tread carefully to ensure that our data remains intact if something happens on CI's end?

Whether or not we want CD's data and CI's data to be coupled, the nature of the relationship between the sites (Cantus IDs are integral to the purpose of CD, and CI is The Authority on Cantus IDs) means that this coupling happens to at least some degree. I'm thinking particularly of https://github.com/DDMAL/CantusDB/issues/877, where we're being asked to change a bunch of our data when a change occurs on CI.

I understand that creating a new model and using FKs to link it with other models creates new ways for things to go wrong... 
- e.g., data that was directly attached to a chant might now be inadvertently changed if our new model changes

...even as it makes some things easier/clearer.
- thinking mostly of the enforcement of controlled vocabularies here. A new model would also make it clear that things like canonical fulltexts and concordances are associated to Cantus IDs rather than with individual chants, which also feels like a plus to me - this is a thing that it took me a while to wrap my head around when I began work on CantusDB, which led to me proposing caching of information from CI within individual chant objects (clearly wrong, looking back on it now) back when I was starting.

I'm guessing that you just have a clearer idea of the potential downsides of creating new models in situations like this, while I don't yet appreciate all the risks that this course of action may introduce and how to safeguard against them.

ahankinson:
> You've expressed something along these lines a few times, and I don't fully understand your concern

I've been trying to find a way of expressing this concern, so apologies if it was a bit vague. 

I think it boils down to concerns about [referential integrity](https://en.wikipedia.org/wiki/Referential_integrity). With a method that uses actual database FKs to implement the model, bringing in external data sources (over which there is very little control) and tightly integrating them at the database level, seems like it has the potential to cause problems. 

I'm still not entirely clear how the `CantusIdentifier` model will integrate with the data in the Cantus Database. Is the plan to store FKs to `CantusIdentifier` instances in places where it calls for a Chant ID? If so, then this is effectively requiring referential integrity between the two bits of data but, since one side of the reference depends on an entirely different database, then it is impossible to actually guarantee the integrity of the link. 

We have to assume that, at some point, an identifier in CI will change in ways that are, at the very least, difficult to manage. Merges seem to be one such way that seems to be on the top of the list. My concern is that with a regular mass import, the data in the `CantusIdentifier` model will be changing in a way that is bypassing the "normal" referential integrity of the underlying database. So you will be relying on automated processes, minimally supervised, to modify your Cantus Database records in response to changes in the Cantus Index records which you may not have anticipated. This, to me, seems like it will lead to someone noticing problems in the data only after it's too late to do anything about it. With a tightly integrated data model, you will be *required* to make these changes, probably through automated methods, since Django will refuse to even function if it encounters a FK relationship that it cannot resolve.

"Weak" references, such as storing the identifier as a string, are much more robust to failure. Django does not care if it can't resolve a string value in an external database, so the site will continue to function, more-or-less, even if something goes catastrophically wrong with CI. Users may be annoyed that the concordance function is broken, but at least the data is intact and unchanged. If something happens and CI gets completely wiped out, or changed in ways you cannot control, then you will at least still have the last known-good value stored. More importantly, it means that you're not reliant on automated processes to keep your database integrity intact. 

So to sum up: If you are thinking of implementing this in such a way that does not require your syncing process to touch or modify the core CD data in any way, go ahead. If, however, you're going to rely on the automated import processes to create, delete, or change the core CD data, however "safe" you might think that change may be, then I would strongly advise against that approach.

ahankinson:
> I know that we currently need to break up our 5*10^5 chants... 

500,000 records is not a large dataset -- there is absolutely no reason, apart from unoptimized code, that this should cause problems. `Killed` is an OS-level message that means the script was taking too much of some resource and the OS needed to step in otherwise the whole system would crash. Since you're running this in Docker, however, there is an extra layer of resource limits that need to be managed. So whether it was the OS that killed it, or Docker, is something that you need to determine.

For comparison, the RISM Online indexer can run parallel processes for hours (depending on the number of CPUs) and never get "Killed." This will process almost 6 million records. 

Tracing the calls on your example, I wouldn't be surprised if your simple `.save()` call is producing thousands of individual queries, since none of them seem to use `.select_related()`. You might consider running a few saves directly in the Django Debug Toolbar's shell: https://django-debug-toolbar.readthedocs.io/en/latest/commands.html

That should tell you how many database queries are actually triggered on a save. 

> I'm guessing, but I'm not certain, that our VM will not have issues with memory if we have a dictionary containing 10^6 concordances - are we safe to assume this?

In normal systems, absolutely. In your dockerized setup, possibly not. Only testing will tell you.

lucasmarchd01:
> (In hindsight, for #1104: @lucasmarchd01, is it possible that we will lose certain `diff_db` values when we mapped chants onto our new Differentia model and then deleted the value in the old field? At a glance, I'm not seeing a provision for `diff_db` values that don't have a corresponding `Differentia` object, other than a message that would print to the terminal. I don't recall seeing such a message printed to the terminal when I ran this locally...)

When I set up this script, the `diff_db` values that were given in the spreadsheet were created and mapped to the chants that contained these `diff_db` values. Any value that wasn't in the list but was in our database created a new `diff_db` object. We didn't set up any way of distinguishing these objects, however, we never deleted the old `differentiae_database` values, so we could use those still as reference.

jacobdgm:
It sounds as though the thing we need to be most concerned about is the merging of Cantus IDs (https://github.com/DDMAL/CantusDB/issues/877) and other places where Cantus IDs may be automatically changed. @lucasmarchd01 and I have come up with an idea about CantusID mergers - I'll open an issue/discussion and link to it here soon. But it seems that having a CantusIdentifier model vs. having a CharField doesn't make this problem better or worse. In both cases, an external site may/will cause the data in CantusDB to change, so we need to tread very carefully.

---

When I initially thought of a CantusIdentifier model, the idea was to create one object for each unique Cantus ID in the database, and that for each object, its `cantus_id` would never change - the only things that would come from Cantus Index would be canonical fulltexts, lists of concordances, and similar.

I'm thinking now that rather than starting with the list of Cantus IDs that we get from Cantus Index, we should make a point of starting with the data already in the database. If Cantus Index doesn't have fulltext/concordance data on a particular Cantus ID, then the fulltext/concordance fields will stay blank, and that's fine (i.e., I'm suggesting we don't make an effort to clean up erroneous Cantus IDs before implementing this). But, if there are some Cantus IDs that exist in Cantus Index, I don't see harm in creating additional CantusIdentifier objects for them. They can sit in the table in our database, appearing in the list of valid Cantus IDs and waiting for a chant scholar to attach chants to them at some point.

ahankinson:
I actually don't think we're very far apart in implementation. Where I think we still might differ is in how the incoming data is **thought about**. I think you're thinking of it as a "core" component of the site -- data to be used and related locally. I'm thinking of it as largely ephemeral -- data that exists only because we can't always go upstream to the real source. 

If I'm wrong in that summary, please let me know.

> we should make a point of starting with the data already in the database.

Yes. This is the cache idea. Intercept requests for an ID that you have "locally" in Cantus Database, request it once, then cache it for all future requests. There is no need to keep a copy of 10,000 Spanish chants if you don't need to.

> But, if there are some Cantus IDs that exist in Cantus Index, I don't see harm in creating additional CantusIdentifier objects for them. 

You should be able to wipe out your local copy and re-fresh the cache. You can create as many `CantusIdentifier` objects as you want, but you should also design it so that they can also be deleted, possibly even on a regular basis.

The main reason for this work, as far as I can see, is to cut down on the amount of traffic sent upstream.

jacobdgm:
> I actually don't think we're very far apart in implementation. Where I think we still might differ is in how the incoming data is **thought about**. I think you're thinking of it as a "core" component of the site -- data to be used and related locally. I'm thinking of it as largely ephemeral -- data that exists only because we can't always go upstream to the real source.

I think that the Cantus IDs themselves are a core component, but everything related to them as data that we're fetching from elsewhere and storing it in a way that makes it easy to use. I'm not sure that I would have chosen the word "ephemeral", but I see where you're coming from.

> > But, if there are some Cantus IDs that exist in Cantus Index, I don't see harm in creating additional CantusIdentifier objects for them.
> 
> You should be able to wipe out your local copy and re-fresh the cache. You can create as many `CantusIdentifier` objects as you want, but you should also design it so that they can also be deleted, possibly even on a regular basis.

Would it be enough to reset the "ephemeral" data on all our CantusIdentifier objects, or do you think it's important that the objects actually be deleted and re-created? If the latter is what you mean, this will be hard to do if we implement a CantusIdentifier model.

> The main reason for this work, as far as I can see, is to cut down on the amount of traffic sent upstream.

This is the main reason, I agree (though this is becoming less necessary since Jan is implementing his One API to Rule Them All). I do think that there a couple of worthwhile side benefits of implementing such a model:
- it makes it easy to enforce controlled vocabularies. This can be done in other ways, yes, but this would allow us to approach the problem in exactly the same way we currently do with Feasts/Genres/Offices/etc.
- a lot of our code becomes cleaner and self-documenting. The fact that our CantusIdentifier model will have certain fields will make it very clear to any new developers that chants have concordances by dint of having a Cantus ID. And `concordances = chant.cantus_id.concordances` is both much shorter and clearer than several lines of code grabbing the Cantus ID, querying an API/cache, and then assigning the result to `concordances`.
  - One reason I don't like the idea of parsing the entire output of Jan's One True API whenever we need to find a canonical fulltext is that it will require more code in our views, whereas we could do this in a single place, i.e., when we query the One True API.
- in addition to reducing the amount of upstream traffic, it also allows us to use these values right away (without waiting for a response from upstream) thus improving load times, and allowing us to use these values even if upstream isn't working. (I recognize that a cache does both of these too)

ahankinson:
> I'm not sure that I would have chosen the word "ephemeral", but I see where you're coming from.

The data as it exists in Cantus Database is ephemeral, in the same way that any upstream cache is ephemeral. 

> do you think it's important that the objects actually be deleted and re-created?

If the objects are able to be deleted and re-created at will, then it means that they are not tightly tied to the Cantus Database objects. That's my primary concern -- that `chant.cantus_id.concordances` is the wrong move, since it ties in changing data from an external source into your own data models, and delegates control of this data to automated processes that may have minimal supervision.

They should be separate and separable. 

> it makes it easy to enforce controlled vocabularies ... in exactly the same way we currently do with Feasts/Genres/Offices/etc.

Correct me if I'm wrong, but these are *local* controlled vocabularies, no? As in, Cantus Database is the primary authority for them?

> a lot of our code becomes cleaner and self-documenting.

I would bracket this one, since I think that's a matter of style and competence. Any code can be clean and self-documenting, if it's written correctly. Parsing JSON doesn't need to be a rat's nest. 

>  it also allows us to use these values right away (without waiting for a response from upstream)

I think this is probably negligible. We're talking milliseconds, up to a second or two, for non-cached data on the first load. Subsequent loads are a hundred ms at worst.

Do you want to have a zoom call at some point to hash this out, rather than filling up this GH convo?

jacobdgm:
> Do you want to have a zoom call at some point to hash this out, rather than filling up this GH convo?

I'd be happy to set up a time to do this if you'd like, but I feel like we're close to converging on a solution. Personally, I find typing things up helps clarify my thoughts, and it probably doesn't hurt to have this record of pros/cons of different approaches kept here for posterity. But yes, happy to do a call if you'd like.

> Correct me if I'm wrong, but these are _local_ controlled vocabularies, no? As in, Cantus Database is the primary authority for them?

Only those Cantus IDs on CantusIndex should be available to select on CantusDB, so I don't think it would make sense to call it a local controlled vocabulary, no.

Okay, I think I'm coming around to your way of viewing things, regarding coupling of data, ephemeral data, and so on. **Current line of thinking:**

- If Jan is fine with us continuing to use [`cantusindex.com/json-cid/{cantus-id}`](https://cantusindex.uwaterloo.ca/json-cid/001234) with his new added `"info"` key, then we should implement a cache - all the information we need from CI will be available here, we can fetch the info per-Cantus-ID, and depending on how we implement it, we can use our Cache model to limit the number of requests we're sending to CI (we can at least limit the request rate per Cantus ID to 1 per several minutes, and there are only ~10<sup>5</sup> Cantus IDs)
- We can use the single big table to populate a list of all the available/valid Cantus IDs.
- If we need to use the single big table and don't have access to the individual per-Cantus-ID endpoints, I still might favour creating a model to break this data up a bit to store in a table in our DB (i.e., a CantusIdentifier model)
- in each case, we still store `cantus_id`s as strings attached to individual `Chant` and `Sequence` objects. Since we don't ever need to display information associated with multiple Cantus IDs at once, we can just do a `CantusIdentifier.get()` or `Cache.get()` query to fetch a single object when loading a particular view (we don't need to worry about lots of queries being generated for a single page, since we'll only ever do this for a single Cantus ID at a time)

ahankinson:
OK -- we can keep going here then.

> Only those Cantus IDs on CantusIndex should be available to select on CantusDB, so I don't think it would make sense to call it a local controlled vocabulary, no.

I meant that the examples you gave (Feast, Genres, Offices) were local controlled vocabs. That was to differentiate them from a "controlled" vocabulary that you would get from outside. (The scare quotes indicate that you don't actually control them, which is kinda the point...)

> we can at least limit the request rate per Cantus ID to 1 per several minutes, and there are only ~105 Cantus IDs

Consider storing an "updated" or "refreshed" timestamp on your entries, and requesting a new copy if it's too old. This will be easier to manage than some sort of global request timer. (i.e., 'user A sent something that triggered a request to CI less than one minute ago, so we can't ask for an update for user B because we haven't waited long enough.')

>  ... I still might favour creating a model to break this data up a bit to store in a table in our DB ... 

However you want to store it is fine... I don't see any benefit here, but it also won't hurt. Parting them out into fields also makes the cache less applicable to other API calls to other sites...

> in each case, we still store cantus_ids as strings attached to individual Chant and Sequence objects.

OK!

> ... Since we don't ever need to display information associated with multiple Cantus IDs at once, we can just do a CantusIdentifier.get() or Cache.get() query ...

If you do have a list of strings to look up [you can do that in a single query too](https://docs.djangoproject.com/en/4.2/topics/db/queries/#the-pk-lookup-shortcut).

```
my_cantuses = Cache.objects.filter(pk__in=["12345", "54321", "98765"])
```

PR #1132, opened 2023-11-03
Change concordance load error message

dchiller:
Took care of this one too, since I know you all are focused on how to actually get these up on the server without breaking the hotfix!

PR #1142, opened 2023-11-09
Remove all places where we send requests to Cantus Index

### LINKED ISSUE #777: Intermittent failure SourceEditChantsViewTest.test_provide_suggested_fulltext
From time to time, the following test fails. I'm learning to ignore it, because it's caused by a failed request to something on CantusIndex (I think. In any case, a failed request to _something_ external). For the benefit of future developers, we should fix this eventually.

```python
======================================================================
FAIL: test_provide_suggested_fulltext (main_app.tests.test_views.SourceEditChantsViewTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/django/cantusdb_project/main_app/tests/test_views.py", line 2072, in test_provide_suggested_fulltext
    self.assertEqual(suggested_fulltext, expected_suggestion)
AssertionError: '' != 'Puer natus est nobis alleluia alleluia'
+ Puer natus est nobis alleluia alleluia
```

### LINKED ISSUE #778: Intermittent failure of ChantCreateViewTest.test_suggest_one_folio
From time to time, the following test fails. I'm learning to ignore it, because it's caused by a failed request to something on CantusIndex (I think. In any case, a failed request to _something_ external). For the benefit of future developers, we should fix this eventually.

```python
======================================================================
ERROR: test_suggest_one_folio (main_app.tests.test_views.ChantCreateViewTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/django/cantusdb_project/main_app/tests/test_views.py", line 1838, in test_suggest_one_folio
    self.assertEqual("007450", response.context["suggested_chants"][0]["cid"])
KeyError: 'cid'
```

### LINKED ISSUE #1106: Chant suggestions are showing empty boxes
On the chant create page sidebar, suggestions based on the previous chant are currently not working because of our recent changes on production that stop our requests to Cantus Index.

We should either hide these suggestion boxes or display a message that says this tool is currently unavailable until we restore our communications with Cantus Index.

image

### LINKED ISSUE #1119: We should disable/remove all places in the code where we request info from Cantus Index
Planning on embarking on this task today.

As I go, I'll compile a list here of features I've properly disabled, so we can re-enable them once we're storing info on Cantus IDs in our own database.

These features should be gracefully disabled for the time being, and they'll need to be significantly reworked anyways when we implement our improved/sustainable solution.

jacobdgm:
This PR removes all places in the code where we send requests to Cantus Index. In doing so, it fixes #1119.

Some notes:
- One function that references CI remains: `parse_json_from_ci_api`.
  - It's been refactored to accept a path rather than a URL; eventually, we can use it to query the One True API that Jan is setting up.
  - If all our requests to CI go through this function, it gives us a single point where we can control (i.e., turn off) these requests in case something else goes wrong with CI in the future.
- This PR removes the CI Search tool, and removes references to it that had existed on the Chant Create page. Thus, it fixes #1106.
- we had several tests that would fail intermittently in local development, when requests to CI would time out. This PR has the nice side effect of preventing these failures. (i.e., fixes #777, fixes #778)
- This PR incorporates changes initially proposed in #1132, so that one can be closed. (thanks, @dchiller, for setting this up!)
- A list of things I removed can be found at https://github.com/DDMAL/CantusDB/issues/1119#issuecomment-1804035810 - they'll all need to be re-implemented


jacobdgm:
As so many files were touched as part of this PR, I'm thinking I'd like to have approving reviews from both @dchiller and @lucasmarchd01 before merging. Thanks!

lucasmarchd01 (approved):

dchiller:
I can review more thoroughly, but I am surprised that much of this is being deleted rather than commented (for example, keeping a function, but commenting out it's call with an explanation).

Will not a number of features (eg. a concordances table, suggested chants) be "turned back on" once we have figured out a solution to the problem?

Of course, you can copy and paste the relevant code back in from a previous commit, but given that turning off requests to Cantus Index is meant to temporarily mitigate what are essentially unrelated problems (the aggressive bots and the way that Cantus Index handles concordances), I would have thought you wouldn't delete.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -67,149 +65,21 @@
...
-        }
-        sorted_database_summaries.append(gregorien_database_dict)
-
-    return sorted_database_summaries
-
-
-def parse_json_from_api(url: str) -> Union[list, None]:
-    """Queries a remote api from cantusindex.org that returns a json object, processes it and returns
-    a list containing its information.
+def parse_json_from_ci_api(path: str) -> Union[list, None]:
```
ahankinson: `Union[list, None] == Optional[list]`

jacobdgm:
> I can review more thoroughly, but I am surprised that much of this is being deleted rather than commented (for example, keeping a function, but commenting out it's call with an explanation).
> 
> Will not a number of features (eg. a concordances table, suggested chants) be "turned back on" once we have figured out a solution to the problem?
> 
> Of course, you can copy and paste the relevant code back in from a previous commit, but given that turning off requests to Cantus Index is meant to temporarily mitigate what are essentially unrelated problems (the aggressive bots and the way that Cantus Index handles concordances), I would have thought you wouldn't delete.

My thought was that, since the plan is to store everything internally, everything that is being removed in this PR will need to be reworked when we reimplement it. We can refer to this PR to see the code that was removed, but I'm thinking of this as opportunity to build these features better.

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2920,82 +2919,18 @@ def test_post_error(self):
...
-        self.assertEqual(1, len(response.context["suggested_chants"]))
-        self.assertEqual("007450", response.context["suggested_chants"][0]["cid"])
-        self.assertEqual(
-            fake_R_genre.id, response.context["suggested_chants"][0]["genre_id"]
-        )
-
-    def test_fake_source(self):
-        """cannot go to input form with a fake source"""
-        fake_source = faker.numerify(
+    def test_nonexistent_source(self):
```
dchiller, line 2922: This addition is unrelated to the main PR right? Looks fine, just curious.

dchiller (approved):

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: yes. I came across it while addressing the rest of the tests, and was initially baffled by the method name. Once I figured out what it was meant to do, I figured it needed to be fixed.

jacobdgm:
I improved and added type annotations to our functions in `views/chant.py` - I was about to set into annotating all the views, but managed to restrain myself...

PR #1156, opened 2023-11-16
autocomplete: change genre, office, feast to use icontains search

### LINKED ISSUE #1155: Chant Create/Edit: Feast/Office/Genre autocompletes should do icontains search rather than istartswith
Testing changes currently on staging, on the Chant Create/Edit pages. It would be nice to be able to find text from the middle of a field, rather than just at the beginning:

Feast field:
- searching for "mariae", I expect "Conceptio Mariae" to come up. It doesn't.

Office field:
- When I search for "V", First and Second Vespers come up. But if I search for "vespers", neither does - I need to search for "first" in order for First Vespers to appear.

Genre field:
- Similarly to Office, I expect Antiphon Verse and Alleluia Verse to appear when I search "verse", but neither does.

This is not important enough to block the merging of #1146, but it would be nice to get it fixed before too long.

lucasmarchd01:
Fixes #1155 

Office and Genre autocomplete widgets are now using icontains search filter on `description` (`name` is still istartswith since they are the shorthand letter representation). Feast are now using icontains search filter on `name`.

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -1014,7 +1014,7 @@ class FeastAutocomplete(autocomplete.Select2QuerySetView):
         qs = Feast.objects.all().order_by("name")
         if self.q:
             qs = qs.filter(
-                Q(name__istartswith=self.q) | Q(feast_code__istartswith=self.q)
+                Q(name__icontains=self.q) | Q(feast_code__istartswith=self.q)
             )
         return qs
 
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/views.py
```diff
@@ -1014,7 +1014,7 @@ def get_queryset(self):
         qs = Feast.objects.all().order_by("name")
         if self.q:
             qs = qs.filter(
-                Q(name__istartswith=self.q) | Q(feast_code__istartswith=self.q)
+                Q(name__icontains=self.q) | Q(feast_code__istartswith=self.q)
```
jacobdgm: it occurs to me now - we're not displaying the `feast_code` anywhere currently (within the autocomplete widget), are we? We should probably not be displaying `feast_code` matches if the user can't see why it's being matched...

jacobdgm (approved):
This looks good. Thanks for setting this up!

PR #1177, opened 2023-11-28
Update json-cid API to conform to new specification

### LINKED ISSUE #1170: We need to rework our json-cid API
Jan recently sent an email saying that work is progressing on his One True API, and he requested that we update our /json-cid/{cantus_id} API with some new information required to make the new system work.

He included a specification:

> To obtain concordances and analysis data from the new CD, I would need a URL for a JSON export that includes all CD chants with specific columns. This file should be updated at least daily (we can later discuss the update frequency). There is also a "sequence" field necessary for suggesting new chants and feasts in the Cantus Input Tool.
> 
> List of columns:   
> 
> |   |   |
> |---|---|
> |siglum|varchar(255)|
> |srclink|varchar(255)|
> |chantlink|varchar(255)|
> |folio|varchar(255)|
> |sequence|smallint|
> |incipit|varchar(255)|
> |feast|varchar(255)|
> |genre|varchar(255)|
> |office|varchar(255)|
> |position|varchar(255)|
> |cantus_id|varchar(255)|
> |image|longtext|
> |mode|varchar(255)|
> |full_text|longtext|
> |melody|longtext|
> |db|varchar(255)|

jacobdgm:
This PR makes changes to our json-cid API, so that it conforms to the new standard specified by Jan (i.e., it fixes #1170)

The changes that occurred:
- In the API,
  - we now include a `"sequence"` key, which is an integer
  - we now include a `"cantus_id"` key, a string
  - our previous `"fulltext"` key is now called `"full_text"`
  - in the code, the order of these keys has been rearranged, and a comment has been removed (it was from back when we were just recreating what was on OldCantus, and were not working from a specification)
- `JsonCidTest.test_structure` and `.test_values` have been updated to reflect the new specification

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -4923,7 +4923,10 @@ class JsonCidTest(TestCase):
         chant = make_fake_chant(cantus_id="100000")
         expected_values = {
             "siglum": chant.source.siglum,
+            "srclink": f"http://testserver/source/{chant.source.id}",
+            "chantlink": f"http://testserver/chant/{chant.id}",
             "folio": chant.folio,
+            "sequence": chant.c_sequence,
             "incipit": chant.incipit,
             "feast": chant.feast.name,
             "genre": chant.genre.name,
@@ -4932,7 +4935,7 @@ class JsonCidTest(TestCase):
             "mode": chant.mode,
             "image": chant.image_link,
             "melody": chant.volpiano,
-            "fulltext": chant.manuscript_full_text_std_spelling,
+            "full_text": chant.manuscript_full_text_std_spelling,
             "db": "CD",
         }
         response_1 = self.client.get(
@@ -4961,7 +4964,14 @@ class JsonCidTest(TestCase):
         )
         json_for_one_chant_2 = response_2.json()["chants"][0]["chant"]
         for item in json_for_one_chant_2.items():
-            self.assertIsInstance(item[1], str)  # we shouldn't see any Nones or nulls
+            self.assertIsInstance(
+                item[1],
+                (
+                    int,  # ["sequence"] should be an int
+                    str,  # all other keys should be strings, and there should
+                    # be no Nones or nulls
+                ),
+            )
 
         chant.manuscript_full_text = "nahn-staendrd spillynge"
         chant.manuscript_full_text_std_spelling = "standard spelling"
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -4895,29 +4899,34 @@ def test_structure(self):
 
         first_chant = first_item["chant"]
         chant_keys = first_chant.keys()
-        expected_keys = [
+        expected_keys = {
```
dchiller, line 4902: Are you trying to test order here? I am wondering because you changed the order in the API response...but a set does not preserve order.

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -4955,7 +4964,14 @@ def test_values(self):
         )
         json_for_one_chant_2 = response_2.json()["chants"][0]["chant"]
         for item in json_for_one_chant_2.items():
-            self.assertIsInstance(item[1], str)  # we shouldn't see any Nones or nulls
+            self.assertIsInstance(
+                item[1],
```
dchiller: I feel like iterating through `json_for_one_chant_2.values()` would be clearer here.

dchiller:
Also, not part of my review because one wouldn't see changes to the wiki in a PR anyway, but a reminder that once this is merged, the wiki (https://github.com/DDMAL/CantusDB/wiki/APIs#json_cid) should be updated #386

django/cantusdb_project/main_app/tests/test_views.py
ahankinson: Also note that, while some APIs and code libraries do preserve order, [JSON Objects are not ordered](https://www.json.org/json-en.html).

> An object is an unordered set of name/value pairs. ...

So it's really in your best interests to not expect objects to be in any particular order, since you don't necessarily know if any particular library will preserve it.

dchiller: Ah! Thanks...

also, in that case, I'm not sure what this is testing that is different than `test_values` immediately below this function.

jacobdgm: I changed the ordering in the code so that it's clear how it corresponds to the specification. But since JSON Objects are unordered, I figured the test would be clearer if we worked with unordered sets - otherwise future developers might think that the ordering is important.

I see where you're coming from with your thought about `test_values`. When I initially implemented/recreated this API, it took me a while to wrap my head around all the levels of nesting in the final response. If I recall, `test_structure` was an effort to communicate how the request was meant to be structured (in a different way than the couple of other forms of documentation we also had). I guess it's not strictly necessary considering we also have `test_values`. If you think it impedes understanding on the whole, we can remove it - let me know what you think.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: Thanks for pointing this out! I just pushed an new commit, updating this to `for key, value in json_for_one_chant_2.items()`, which uses the key for identifying `subTest`s. I haven't been using subtests so far in places where I really should be - this felt like a good time to start!

dchiller (approved):

Issue #1165, opened 2023-11-23
remap_user_ids command is not working properly


jacobdgm:
I tried running our new `remap_user_ids` command on staging, and upon asking @annamorphism to verify that things ran properly, she pointed out several issues:

http://206.12.88.113/source/713037 now has no Contributor (compare to https://cantusdatabase.org/source/713037)

http://206.12.88.113/source/711225 - My Project Manager test account has been added as a contributor on this and perhaps other sources. This is clearly an error. (compare to https://cantusdatabase.org/source/711225)

One thing to do before testing this again is to download a copy of the database from Production and put it on Staging before testing again. (I hadn't done this, thinking the data was probably close enough to up-to-date. But relevant users have been making changes lately, so this should really be updated.)

jacobdgm:
The path forwards:

- We need to deploy certain changes currently on Staging to Production as soon as we reasonably can, so I think we should disable this script to ensure it doesn't get run on Production, and return to it sometime in the coming weeks. I'll open an issue for this.

lucasmarchd01:
I ran this script locally and found that (at least for the two sources mentioned in this issue) source 713037 has a contributor this time, and Jacobs project manager test account was not added to source 711225. I also went through the code in debug mode and everything seemed to be working as expected.

It's possible that the issue has to do with data discrepancies between production and staging. The next step will be to use a fresh copy of the database from production.

lucasmarchd01:
I have just run the `remap_user_ids` script on staging. From what I can tell, it has worked as expected. Can @jacobdgm or @annamorphism confirm this?

jacobdgm:
My working hypothesis is that within the past few weeks, I changed the contributor of http://206.12.88.113/source/711225 to something nonsensical (i.e. my "Project Manager" account) on staging, since it's staging and it doesn't really matter what happens with the data on staging, and then forgot about it. If this is the case, the wonky contributor assignment of this source had nothing to do with the script, and it's been overwritten since Lucas overwrote the database with a copy of what's on Production.

It doesn't explain the contributorless sources, though. Maybe the script didn't run to completion when I first ran it on the staging server?

annamorphism:
So, the correct users seem to be showing up now! This is good. 
[...]

Issue #1079, opened 2023-09-29
Source Detail Page/Sidebar: Allow for multiple image views (Cantus Ultimus and host institution)


jacobdgm:
In an email from Debra:

> For the Cantus Ultimus sources, Jennifer and I would like those to be one of the optional "views" in the RH menu. See the screenshot and imagine that "View on Cantus Ultimus" is one of the ways to see the material.
> 
> We might rename these links (because someone coming new to the site might not know the difference between them), but 1 would go to all chants in ms order in a view that shows the fields we have deemed more important with the benefit of a full text (can it be the MS spelling if it is included instead of the first full text?) and Volpiano melody; 2 would go to the "phonebook style" listing (for those of you who know what a phonebook was) where you can easily scan/scroll through the entire thing and see it all lined up (this remains one of the views by request of many researchers - we can't ever take this one away); and 3 would be a link to the Cantus Ultimus version (in a new window, of course, so the user can get back to CantusDB if they want).
> 
> For the other sources in CD & CU, I would like to keep the image libraries pointing to the archive where they are stored (i.e., Gallica, e-codices, etc.) so we continue to credit the correct people.
> 
> Does this all make sense?
> We want to interconnect CD & CU as much as possible now, especially since we have control over the coding of both sites. Back and forth between the two makes a lot of sense to me, and if we can get the additional menu option in CD for the CU sources, we can promote those within CD more, since I expect researchers who don't know about CU yet will love that they can get the Cantus data onscreen beside the images.

@annamorphism responded, summarizing thus:

> OK, so you're imagining the connectivity happening at the level of the source, but not with individual chant records linking to their Cantus Ultimus counterparts (except for Salzinnes).  That seems doable! We can make it a "feature request." (I think in principle—Jacob and Dylan can correct me here-- the chant records could also have an "image link" and a "view on Cantus Ultimus" link, if we wished, which might give even more interconnectivity—or would that be too much of a good thing? Or maybe it's harder to program than I imagine.)

jacobdgm:
It would be easy to add this field while keeping the "image gallery" link intact.

I agree with @annamorphism that it could be a good idea (and it would not be very difficult) to add a "view on Cantus Ultimus" link to Chant Detail pages (and thus to the Chant model).

dchiller:
On source detail page (see sidebar on upper right)

[...]

Issue #940, opened 2023-08-04
We need a workflow for attaching image links to records


annamorphism:
In an email, Debra requested "a method for me to submit or request that image links be attached to records".
"I used to send Jan a URL of a manuscript in a digital library, and he created the links. Depending on the urls, he could match our folio numbers to the correct images. (Sometimes we could just link to the first page, but for many we have direct links to each page.)"

I will note for posterity that on OldCantus the Admin area included a link simply labeled "Image Links", which leads to https://cantus.uwaterloo.ca/image-links-csv; it seems somebody (presumably Jan) could upload a CSV with all the links here. 

<img width="792" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/95325e36-0caf-4a89-97b1-39cc85080954">

This particular CSV seems to have the source id, folio #s in sequence, and a URL for their page view (which increment by 1 every folio)--presumably each CSV is/was individually created by inspection of the URLs and then gets fed into the image links for each chant view.
<img width="482" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/573151e4-2c33-4b41-be10-3d3108a68f2f">

jacobdgm:
We've let Debra know that she needs to send us (whoever's developing on CantusDB at the time) a link to the images and a link to the MS, and we'll add the proper image links to all the chants in the MS. We can set up tools to help us accomplish this however we want (since Jan was the only one to interact with this page on OldCantus, we think, we don't need to set this up the exact same way), so it might make sense to wait for Debra to send us a few of these. In going through the import process several times, it will become clear the best way to automate it.

annamorphism:
I think I have a list of manuscripts needing links somewhere, actually! Is what you want the Source ID# and the URL to e.g. folio 1r?

jacobdgm:
> I think I have a list of manuscripts needing links somewhere, actually! Is what you want the Source ID# and the URL to e.g. folio 1r?

yes, that should be all we need!

dchiller:
We have our first new example of this! 

[...]

Issue #429, opened 2022-11-23
On Chant Detail page, we should add links to corresponding Cantus Ultimus page


jacobdgm:
From an email from Debra:

> A really helpful item in the Cantus Database detail record (the view of a single chant in a source) would be the link to that chant/view in Cantus Ultimus.
> 
> Could we add this field right away... ?
> 
> Then, can the folio/page links for all the CU-processed sources be fed automatically into that new field?
> 
> I can imagine this link being directly under the Image Link field on both the data entry screen and the view for the user.

jacobdgm:
In talking with @dchiller about how optimally to implement this, some relevant points emerged:

- sources/manuscripts have different identifiers in Cantus Database and Cantus Ultimus (for example, see Einsie in [Cantus Database](https://cantus.uwaterloo.ca/source/123606) vs [Cantus Ultimus](https://cantus.simssa.ca/manuscript/74/))
- it would be possible to set up a mapping between sources in CD and those in CU, it require an amount of data entry
- eventually, for a number of reasons, it would make sense to harmonize the source/manuscript identifiers in CU with the source identifiers in CD.
- Even though its URLs sometimes include an `&chant=N` query parameter, Cantus Ultimus is not currently set up to show a specific chant on a folio when that query parameter is specified.
  - so, for any given chant in Cantus Database for which we want to generate a link to the corresponding page in Cantus Ultimus, all we would need to provide is
    - the proper manuscript ID, and
    - the chant's `folio`

jacobdgm:
@dchiller, you were working on renumbering sources a while ago, right? How is that going? This might be in a place where it will be trivial to implement - can you comment?

dchiller:
When the source is on Cantus Ultimus, we should have the image link to there; if not, to the host institution. But we don't need both (and also not another field) because we can construct the CU links from the chant folio and sequence #.

PR #1179, opened 2023-11-28
Delete management commands we no longer need

### LINKED ISSUE #1168: we may want to delete/disable many of our management commands
We have a bunch of management commands that have already served their purpose. They include commands to sync data from OldCantus, commands to populate fields with initial values which are now kept up-to-date by signals, and so on.

With #1167, I disabled a command we're working on to ensure it isn't run.

Some of our commands would simply hit errors if we tried to run them now (e.g., they assume fields that no longer exist on our models). Some would cause serious problems if run (e.g., the sync scripts). For each command, I can guess which would happen, but I'm not sure.

To ward off the possibility of someone running one of these commands in error, we should consider disabling the commands we no longer need.
- we could either delete the commands entirely, which would lead to a slimmer codebase. In case we ever need to check what these scripts did, there will always be a record in the git history.
- we could raise `DeprecationWarning`s instead of running each command. The commands will stay in the codebase for easy reference, but a developer cannot accidentally run the command without making a change to the code itself.

I favour the first of these two approaches - our codebase is big enough already.

jacobdgm:
This PR deletes a number of commands in `management/commands` that are no longer needed. In doing so, it fixes #1168.

It renames and slightly reworks one command (`populate_chant_fields` has become `touch_all_chants`), deletes many commands, and retains just a few. For justification for deleting each of the deleted commands, see https://github.com/DDMAL/CantusDB/pull/1179#issuecomment-1830949388 below. For justification for retaining each of the retained commands, see https://github.com/DDMAL/CantusDB/pull/1179#issuecomment-1830952088 below.


jacobdgm:
Commands deleted:
- update_differentia_new
  - we created a field, `differentia_new` (now `differentiae_database`), which hadn't been there when our chant model was initially set up. This command looked at chants' `json_info` field and took the relevant value from there to populate the new field.
  - this command only needed to be run once, and it has. It's done its job!
- sync_sources
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- sync_source_creation_date
  - we previously hadn't kept track of the time of creation of the sources in CantusDB. This script got this information from one of the APIs on OldCantus and added it to the existing sources on NewCantus. It's been run, and its job is complete.
- sync_sequences
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- sync_indexers
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- sync_genres
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- sync_feasts
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- sync_chants
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- replace_indexer_fields
  - this was a necessary step to merge indexers and users. This merger is complete; this command has done its job
- populate_volpiano_new_fields
  - there are several volpiano-related fields (`volpiano_notes`, `volpiano_intervals`) that are used on the Melody Search page. These fields are now kept up-to-date by `update_volpiano_fields` in `signals.py`
- populate_next_chant_fields
  - the `next_chant` field is used to optimize the calculation of feasts that follow a given feast. its value is kept up-to-date by `signals.py`, and this script no longer needs to be run.
- populate_is_last_chant_in_feast
  - the `is_last_chant_in_feast` field is used to optimize the calculation of feasts that follow a given feast. its value is kept up-to-date by `signals.py`, and this script no longer needs to be run.
- populate_chant_fields
  - this PR renames this command from `populate_chant_fields` to `touch_all_chants`. The command goes through every chant and saves it, triggering `on_chant_save` in `signals.py`. It was meant to update an existing copy of the database while adding new behavior to `signals.py`. Since this behavior may still be useful in the future as we make further changes to `signals.py`, it has been retained, and renamed to reflect its purpose.
- populate_chant_and_melody_counts
  - this populates the number of chants and number of chants with melodies for each source in the database. These values are now kept up to date in `signals.py`, and this command should never need to be run again.
- merge_indexer_user
  - The indexer-user merger is now complete. This command has done its job!
- add_unknown_source
  - it used to be that some chants/sequences had no source, but Source is now a required field for those models. This command created two "Unknown Source" sources, one for chants and one for sequences, and assigned sourceless chants/sequences to these sources. This command has already run to completion on Production, and running it would both cause problems (creating duplicate "Unknown Source" sources) and pointless (since all chants/sequences already have sources)
- add_prefix
  - this command added feast codes and prefixes to each Feast in the database. The feast code functionality was added to the `sync_feasts` command, and the prefix field is kept up-to-date by a signal in `signals.py`. This command is now several levels of out-of-date.
- add_all_users
  - one of the scripts for syncing data from OldCantus to New. It's done its job!
- add_differentiae
  - This just ran to completion on Production, and seems to have worked. Its job is done!

jacobdgm:
Commands that I left:
- add_cantus_ultimus_links
  - this is implicated in #429, #940, #1079
- remap_user_ids
  - this is currently in the workshop, as it's needed to fix #1165
- touch_all_chants
  - this command was renamed from `populate_chant_fields`. See https://github.com/DDMAL/CantusDB/pull/1179#issuecomment-1830949388 for further details.

django/cantusdb_project/main_app/management/commands/touch_all_chants.py
```diff
@@ -18,10 +23,14 @@ def handle(self, *args, **kwargs):
         chants_count = chants.count()
         start_index = 0
         while start_index <= chants_count:
-            print("processing chunk with start_index of", start_index)
+            self.stdout.write(f"processing chunk with {start_index=}")
```
dchiller, line 26: What does the equal sign do here?

dchiller (approved):
Approving...everything makes sense.

PR #1183, opened 2023-11-30
exclude fields from admin change page

### LINKED ISSUE #1181: `created_by` and `last_updated_by` fields should be read only on source admin change page
They should be readonly fields instead. I don't recall seeing them at the top of the page like they are now. Maybe previously they were hidden and now they are visible.

lucasmarchd01:
The base `EXCLUDE`, which is used in most of our other admin panels, was not included in the source admin. I suspect this happened when `source_status` was added as an exclude field in #1116. For reference:
```
EXCLUDE = (
    "created_by",
    "last_updated_by",
    "json_info",
)
```

This might also be a good time to consider having the `created_by` and `last_updated_by` fields as readonly instead of excluded. This information might still be useful to view on the admin change pages. @jacobdgm, let me know if I should make this change as well.

Fixes #1181

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -172,7 +172,7 @@ class SequenceAdmin(BaseModelAdmin):
 
 
 class SourceAdmin(BaseModelAdmin):
-    exclude = ("source_status",)
+    exclude = EXCLUDE + ("source_status",)
 
     # These search fields are also available on the user-source inline relationship in the user admin page
     search_fields = (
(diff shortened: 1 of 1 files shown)


jacobdgm (approved):
> This might also be a good time to consider having the `created_by` and `last_updated_by` fields as readonly instead of excluded. This information might still be useful to view on the admin change pages.

I think this is a good idea. Approving, but go ahead and make this change before you merge. Thanks!

lucasmarchd01:
image
Now, `created_by` and `last_updated_by` fields are `readonly_fields`

jacobdgm (approved):

PR #1184, opened 2023-11-30
Add django-reversion

### LINKED ISSUE #430: Add Change History to all models
At the LinkedMusic project meeting, it was suggested that we should begin keeping a history of all changes to all of our models (rather than the current approach of simply noting when the most recent edit occurred and the user who made the edit). @ahankinson noted that there are simple, off-the-shelf ways to implement this (and suggested [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/), if I recall correctly).

jacobdgm:
This PR fixes #430 by installing `django-reversion`.

When this change makes its way to Staging and then to Production, we'll need to run both of the following commands:
- `python manage.py migrate`
  - for reasons I don't understand, migrations related to `django-reversion` are listed among the migrations when I run `showmigrations`, even without me having first run `makemigrations`. So, there are no migrations included in this PR. It seems to work fine anyways. 
- `python manage.py createinitialrevisions`

By adding django-revision to INSTALLED_APPS and to MIDDLEWARE, this ensures that all modifications made to objects via our views (and perhaps also via the admin interface, but not via the shell) are added as revisions. By setting `BaseModelAdmin` to inherit from `reversion.admin.VersionAdmin`, this adds a "History" button on all Change {Model} pages in the Admin area, where all modifications to the object in question are listed. It also adds a "Restore Deleted {Model}s" button allowing a user to view and to restore deleted objects.

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -61,6 +61,7 @@ INSTALLED_APPS = [
     "main_app",
     "articles",
     "django_quill",  # to provide rich-text field for articles
+    "reversion",  # django-reversion, for version history of objects in database
     "users",
 ]
 
@@ -73,6 +74,7 @@ MIDDLEWARE = [
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "django.contrib.flatpages.middleware.FlatpageFallbackMiddleware",
+    "reversion.middleware.RevisionMiddleware",
 ]
 
 ROOT_URLCONF = "cantusdb.urls"
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from reversion.admin import VersionAdmin
 from main_app.models import *
 from main_app.forms import (
     AdminCenturyForm,
@@ -22,7 +23,7 @@ EXCLUDE = (
 )
 
 
-class BaseModelAdmin(admin.ModelAdmin):
+class BaseModelAdmin(VersionAdmin):
     exclude = EXCLUDE
 
     # if an object is created in the admin interface, assign the user to the created_by field
--- a/django/cantusdb_project/requirements.txt
+++ b/django/cantusdb_project/requirements.txt
@@ -13,6 +13,7 @@ Django==4.2.3
 django-autocomplete-light==3.9.4
 django-extra-views==0.13.0
 django-quill-editor==0.1.40
+django-reversion==5.0.8
 django_debug_toolbar==3.8.1
 Faker==4.1.0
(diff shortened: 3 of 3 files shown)


ahankinson:
> for reasons I don't understand, migrations related to django-reversion are listed among the migrations when I run showmigrations, even without me having first run makemigrations

Third-party modules can come with their own migrations if they need to make changes to the database (Which reversion definitely does). 

https://github.com/etianen/django-reversion/tree/master/reversion/migrations

ahankinson (approved):

dchiller (approved):

jacobdgm:
Note that as I tried to run `createinitialrevisions` on staging, the command took too long and ssh quit on me:

```
root@da42ee733a02:/code/django/cantusdb_project# python manage.py createinitialrevisions
Creating revisions for century
- Created 54 / 54
Creating revisions for chant
client_loop: send disconnect: Broken pipe
client_loop: send disconnect: Broken pipe
```

I'll try running it again and see what happens. Just noting that we may need to babysit this and occasionally press a key or two while running this on Production. Not sure whether `caffeinate` would help. Maybe I should do some research into how long it takes for `ssh` to disconnect, and/or in what situations it does this.

jacobdgm:
update: I just tried running createinitialrevisions again on staging, and it seems the command ran to completion in the background in the first place: 

```
Creating revisions for century
- Created 0 / 0
Creating revisions for chant
- Created 0 / 0
Creating revisions for differentia
- Created 0 / 0
Creating revisions for feast
- Created 0 / 0
Creating revisions for genre
- Created 0 / 0
Creating revisions for notation
- Created 0 / 0
Creating revisions for office
- Created 0 / 0
Creating revisions for provenance
- Created 0 / 0
Creating revisions for segment
- Created 0 / 0
Creating revisions for sequence
- Created 0 / 0
Creating revisions for RISM siglum
- Created 0 / 0
Creating revisions for source
- Created 0 / 0
```

Going in the admin interface, I can view a version history for sources. For all the sources I've checked, this is just the "Initial version." I've tried deleting several things and then recovering them, and it works just fine. So it seems that everything is as it should be.

ahankinson:
I run commands like this in screen. It will keep the session even if disconnected

jacobdgm:
> I run commands like this in screen. It will keep the session even if disconnected

I was not aware of `screen` - very excited to have learned about this!

The bit of googling I've done suggests that one should `ssh` into the VM before running `screen`; if ssh disconnects, one can connect to the VM again and then use `screen -r` to resume the session that is still running on the VM (rather than running `screen` before `ssh`ing). Does this match how you use it?

ahankinson:
Yes. Also `screen -ls` will list all the open screens;  `Ctrl-A-D` will detach a running screen and put you back at your main shell, then you can reconnect when you want to rejoin the session.

ahankinson:
Also tmux is a newer, hipper alternative but I don't have it in my muscle memory yet. You might want to learn that if this is a new world.

Issue #1202, opened 2023-12-07
We should consider changing the way Chant List URIs are constructed


jacobdgm:
Currently, our Chant List page lives at `/chants/?source=<source id>`. This might be better at a URI constructed along these lines: `/source/<source id>/chants`, which makes it clearer that this list of chants is a thing that belongs to this particular source. If we made this change, we could set up redirects to ensure backwards compatibility.

One thing that this would allow: If an unauthenticated user lands on the Chant List page for an unpublished source and then tries logging in, we could bring them back to the same Chant List page, since there are no query parameters that would get lost in the process (see #1201). But I think this change would be worthwhile even if this related issue didn't exist - it would be nice to have cleaner, more consistent URIs for our pages.

PR #1204, opened 2023-12-07
Chant List page: Ensure 404 status is returned when `source` parameter is missing

### LINKED ISSUE #1201: We should return a 404 response when a user arrives at `/chants/` without query parameters
...rather than the 500 Server Error that currently occurs.

The Chant List page, with the path `/chants`, expects a query parameter, `?source=` in order to load the page. This all works fine unless a user visits a source and then presses the `Log in` button. The path of the page they're currently on is preserved as a query parameter (e.g., if you are on https://cantusdatabase.org/chants/?source=711225 and press "Log in", the URL of the next page is https://cantusdatabase.org/login/?next=/chants/. Note that the necessary `?source` parameter is not preserved. Upon finishing the login process, you land at https://cantusdatabase.org/chants/, which generates the 500 error.

It might make sense to rework the URL of the Chant List page to something that works better as a URI (`/source/<source id>/chants`?) and set up redirects for backwards compatibility. I'll open a separate issue for this, but we should not hit a 500 error when no query parameters are included.

jacobdgm:
This PR adds a check to ensure that we return a 404 status code when a user visits the Chant List page without specifying a source. It also catches the situation where the user specified a non-existent source. This prevents us from running into an uncaught exception and returning a 500 Server Error, as we currently do. In doing so, it fixes #1201.

This PR also includes a few new tests to ensure this check is working, and cleans up the import statements a bit at the top of `views/chants.py`.


lucasmarchd01 (approved):

dchiller (approved):

ahankinson:
See my comment on the issue. It should probably be a 400, not a 404.

jacobdgm:
> See my comment on the issue. It should probably be a 400, not a 404.

Good catch. Hopefully this error message will not be necessary for long, and we'll put together a better URI soon, but may as well get this right for the time being.

I'm at the end of my workday, but first thing tomorrow I'm going to turn off DEBUG and ensure the new 400 page works as expected, then mark this PR as ready-for-review.

annamorphism:
> > See my comment on the issue. It should probably be a 400, not a 404.
> 
> Good catch. Hopefully this error message will not be necessary for long, and we'll put together a better URI soon, but may as well get this right for the time being.

the title should also say 400 rather than 404, yes?

jacobdgm:
Hm... with my custom 400 page, I get shown 500 Server Errors at `http://localhost:8888/chants/?source=` and `http://localhost:8888/chants/`, but only when `DEBUG` is set to `False`.

When DEBUG is True and my 400 page is present, I get shown a yellow-and-grey traceback page (though no error code at the top, as I see when we hit a 404) when I try to visit the page in a browser.

When DEBUG is False and my 400 page is absent, `ChantListViewTest.test_missing_query_parameter` passes, but I get shown a 500 page when I visit the page in a browser.

In researching this, I've learned that while Django allows a developer to raise a `django.http.Http404` exception which gets caught by a handler somewhere and causes the 404 page to be displayed, Django doesn't supply a similar `Http400` exception. Other exceptions like `django.core.exceptions.BadRequest` are not caught in this same way.

For the time being, while I think it would be better to return a 400 response in this situation, it will be more trouble than it's worth to set up. Much better to show a custom 404 page than the 500 page Nginx gives us. **I'm going to reset things to the way they were when I initially opened this PR,** and open a low-priority issue suggesting we fix the response code at some point in the future.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -89,7 +85,7 @@ def parse_json_from_ci_api(path: str) -> Optional[list]:
     assert path[0] != "/"
     url: str = f"https://cantusindex.org/{path}"
     try:
-        response: Optional[Response] = requests.get(
+        response: Optional[Response] = get(
```
ahankinson, line 88: You should not de-namespace this -- it's important to know *what* get you're calling here, particularly in a request handler that, itself, defines it's own `get` method.

ahankinson:
I'm really not a fan of raising Http errors in a method that is, ostensibly, one that handles database lookups. The better way to do this is to validate that the request is well-formed before it even gets to do the database lookup. 

As it stands, if the request does not contain a `source` query handler, you're still querying for `None`. Better would be to check the query argument and raise the 404 if it's `None`. 

If you wanted to do it properly, you can override the `get` handler and check the incoming request there before even calling `get_queryset`. You don't have to override much logic -- check the request, if it has the necessary parameters then call `super().get` -- otherwise return a response object with a 400 status.

jacobdgm:
Current line of thought: table this PR, and work on implementing #1202. I think it will be simple to set this up properly once we don't need to deal with query parameters.

PR #1206, opened 2023-12-07
Hide created_by, last_updated_by, json_info from SourceAdmin

### LINKED ISSUE #1205: We should exclude created_by, json_info, etc. from SourceAdmin
Currently, on both Staging and Production, on the Change Source page, we can see "Created by", "Last updated by" and "Json info". These should all be hidden.

jacobdgm:
Fixes #1205.


jacobdgm:
gosh, the commit history is all messed up. let me close and re-create this.

PR #1207, opened 2023-12-07
admin: source change: Hide created_by, last_updated_by, json_info

### LINKED ISSUE #1205: We should exclude created_by, json_info, etc. from SourceAdmin

jacobdgm:
fixes #1205


lucasmarchd01:
see PR #1183. I forgot to link an issue to this PR, so we can go ahead with this one and I can modify the other PR to only fix the issue of being `read_only` instead of excluded

jacobdgm:
Oops, I forgot about this. Let's do this within your PR. Thanks for fixing this!

Issue #1209, opened 2023-12-12
We need a new API that displays concordances information for all chants in the database


jacobdgm:
In an email from Jan:

> The intensive process of getting all the data from CD via individual json-cid requests (59.000+) is running already on the 3rd day (and not finished yet) but this will not keep the Cantus data fresh in the Cantus Index API in the long term.
> 
> The solution would be to regularly create a large JSON file export of all the CD chants (with the same fields as in json-cid exports) and make it available as a file to download. An example of such json export is here: https://austriamanus.org/files/concordances-export.json
> This kind of data transfer works also with the MMMO database which has approximately half the amount of data compared to a CD. I believe it would also be the best solution for CD.

This will not be difficult. We can use the code in our `json-con` API, but return all chants rather than filtering them by Cantus ID.

What's a good path for this API to live at? `/json-concordances-export`?

jacobdgm:
I just tried setting this up along the lines of our existing json-cid API, the only substantial change being I didn't filter by Cantus ID. And go figure, `runserver` times out before our backend can finish processing the request.

A couple of possible solutions come to mind:
- we can occasionally generate the entire file and store it, along the lines of what Jan suggested (i.e., cache the file _in toto_)
- we can create a new `concordances_export` (or something) field on our BaseChant model which we use to store the dictionary objects needed for our `json-cid` and `json-concordances-export` APIs, and keep it updated with a hook in `signals.py` (i.e., cache the important information on individual objects in the database).
  - I'm not certain that this is the case, but I think this could be enough to calculate the entire response in one shot. Either way, it would make both APIs faster, even if we ended up needing to cache an entire file too.

Thoughts on both of these options?

jacobdgm:
> we can create a new `concordances_export` (or something) field on our BaseChant model which we use to store the dictionary objects needed for our `json-cid` and `json-concordances-export` APIs, and keep it updated with a hook in `signals.py` (i.e., cache the important information on individual objects in the database).

Hm... it occurs to me that the values in `["chantlink"]` and `["srclink"]` will need to be configured properly to use the correct domain, which is something Django gives to us in our views (`request.build_absolute_uri()`), but I don't think it will simply hand us in `signals.py`. So caching these values on chant objects in our database will maybe be a bit tricky.

ahankinson:
Have you benchmarked your sql queries to figure out why it’s slow?

ahankinson:
A few thoughts:

[...]

PR #1212, opened 2023-12-12
Write tests for future json-concordances-export API

jacobdgm:
In the spirit of test-driven development, I wrote a bunch of tests for the json-concordances-export API Jan asked for (see #1209). Upon trying to implement the API, however, I discovered it takes too long to process the request (see https://github.com/DDMAL/CantusDB/issues/1209#issuecomment-1852712180). I figured that we may as well add the tests to the codebase (with `@unittest.skip()` decorators) in anticipation of this API being implemented, even as we are still deciding how best to optimize the API itself.

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -16,14 +16,13 @@ from main_app.forms import (
 )
 
 # these fields should not be editable by all classes
-EXCLUDE = (
-    "json_info",
-)
+EXCLUDE = ("json_info",)
 
 READ_ONLY = (
     "created_by",
     "last_updated_by",
 )
+
 
 class BaseModelAdmin(VersionAdmin):
     exclude = EXCLUDE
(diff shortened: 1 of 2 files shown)


dchiller:
@jacobdgm Just checking....are these still relevant given discussion this morning and on #1209? In other words, should I still review?

jacobdgm:
> @jacobdgm Just checking....are these still relevant given discussion this morning and on #1209? In other words, should I still review?

I think all these tests will be relevant, but we may need to update the where we get the responses we're testing. I'm not sure how much of an impact different approaches to caching will have on this, so may as well hold off for now. I'll mark this as a draft and re-request your review as our picture of how this API will work comes into clearer focus.

jacobdgm:
This API has been implemented, and there are tests for the function that caches the concordances (nginx serves the file rather than django serving it as a view, as I initially assumed when I wrote this test). This PR is no longer necessary. Closing.

PR #1233, opened 2023-12-20
Add automatic certificate renewal with lego

dchiller:
Simplifies certificate management on the Cantus DB production server. Removes the use of `certbot` for certificate renewal and introduces `lego`. 

- Removes constantly-restarting `certbot` container and associate cron job.
- Intalls `lego` in the `nginx` container, removing need for certificate renewal process to touch multiple containers.
- Updates timing of certificate renewal in cron job (Thinking: we can check for renewal once-per-month and only renew if certificate expires within 45 days).

### DIFF
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -10,4 +10,4 @@
 20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
 30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
 40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
-50      4       *       *       0       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org -d www.cantusdatabase.org --renew-hook "nginx -s reload"
+50      4       *       *       0       /usr/local/bin/docker-compose run nginx lego -d cantusdatabase.org -d www.cantusdatabase.org --http --http.webroot /var/www/lego/ renew --days 30 --renew-hook "nginx -s reload" 
--- a/cron/cron.txt
+++ b/cron/cron.txt
@@ -10,4 +10,4 @@
 20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
 30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
 40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
-50      4       *       *       0       /usr/local/bin/docker-compose run nginx lego -d cantusdatabase.org -d www.cantusdatabase.org --http --http.webroot /var/www/lego/ renew --days 30 --renew-hook "nginx -s reload" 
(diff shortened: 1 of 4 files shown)


dchiller:
@ahankinson This introduces lego for certificate management and reduces the need for additional containers just for certificate management. What do you think?

nginx/Dockerfile
```diff
@@ -1,2 +1,8 @@
 FROM nginx:1.24
+ARG LEGOCHECKSUM="f5a978397802a2eb20771925ceb173dff88705b45fdbb2e68312269e205fa85d"
```
jacobdgm, line 2: Can we verify that it's fine that this is being committed to this, a public repository?

cron/cron.txt
```diff
@@ -10,4 +10,4 @@
 20      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/monthly.sh
 30      4       1       1       *       bash /home/ubuntu/code/CantusDB/cron/postgres/yearly.sh
 40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
-50      4       *       *       0       /usr/local/bin/docker-compose run --rm certbot certonly --webroot --webroot-path /var/www/certbot/ -d cantusdatabase.org -d www.cantusdatabase.org --renew-hook "nginx -s reload"
+50      4       7       *       *       /usr/local/bin/docker-compose run nginx lego -d cantusdatabase.org -d www.cantusdatabase.org --http --http.webroot /var/www/lego/ renew --days 45 --renew-hook "nginx -s reload"
```
jacobdgm: can we add `staging.cantusdatabase.org` to this line (which has now been set up - hooray!), and then test it on staging before we commit to running it on Production?

dchiller: For testing or to serve staging over https? In any case, I think it's a replace rather than add, since `cantusdatabase.org` won't validate on the staging server. But if we want to serve staging over https long-term, then we should maybe use an environment variable or something to switch between staging and production.

nginx/Dockerfile
dchiller: Yes, this is publicly available.

cron/cron.txt
ahankinson: With Ansible you could template this file and deploy it, with the appropriate host name, to the right server. I would recommend that approach.

jacobdgm (approved):
As far as I can tell, this is good to go. Anything left to do before we merge this?

dchiller:
If staging is set-up with a host name, then we could test once it makes it's way there. On production, we'll want to test and make sure to re-register the cron job. On both, we'll want to make sure to get rid of the existing certbot containers. But nothing standing in our way of merging the code, I don't think.

jacobdgm:
> If staging is set-up with a host name, then...

is http://staging.cantusdatabase.org sufficient as a host name?

ahankinson:
Don’t forget that the path to your certificates will change in your nginx conf…

dchiller:
> Don’t forget that the path to your certificates will change in your nginx conf…

You're right...that should be changed in the codebase. I'll take care of it.

jacobdgm:
> > Don’t forget that the path to your certificates will change in your nginx conf…
> 
> You're right...that should be changed in the codebase. I'll take care of it.

Going to mark this PR as a draft until this is complete, to prevent accidental merging. Thanks @ahankinson, @dchiller!

dchiller:
I just updated the `--path` option for the `lego renew` command in `cron.txt` to match roughly where we are currently storing certificates...however, I can't figure out what lego names the certificate files (and a way to pass a custom name looks like it is deprecated), nor can I verify if/what subdirectories are created as part of the process. 

I propose we merge now, and when we update staging, we get a certificate for staging, determine the filenames, and come back to make changes through a new PR at that point.

jacobdgm:
> I propose we merge now, and when we update staging, we get a certificate for staging, determine the filenames, and come back to make changes through a new PR at that point.

I'll put this change up on staging now.

dchiller:
> I'll put this change up on staging now.

Thanks! Happy to help troubleshoot if anything comes up.

PR #1235, opened 2023-12-21
Fix postgres pack-up cron scripts

### LINKED ISSUE #982: Cron not saving backups properly
I checked a little while ago, and I couldn't find the daily database backup on the Production machine which should have been created at 4am today.

@dchiller and I spent a while looking into this, and we think we have a solution. We'll try implementing it, and if it seems to work, we'll update the documentation. I'll also add more details to this issue about what the problem was.

dchiller:
I found a number of reports that confirmed this behaviour, with consensus being that it has to do with the piping of output when the cron job is running the `docker exec` function. 

This PR creates a separate script that runs the pg_dump command directly from within the docker container. This script is then called through the `docker exec` command in the cron job. The resulting backup file is copied to the expected directory.

Since each of the shell scripts (`daily.sh`, `weekly.sh`, etc) store some variables and then complete the exact same procedure, a more elegant solution might be to just have a single script that is called with arguments. I've gone for the 'minimal change' approach here, but am open to the argument that this should be adjusted in this PR.

This has been tested on the server. Fixes #982.


ahankinson:
Aren't these all just the same script?

a) why not just run them at different intervals?
b) why not just run a nightly, and then rotate? 7 nightlies, then keep one as your weekly, then keep one weekly as your monthly, then keep one monthly as your yearly. Or keep all 12 monthlies in perpetuity.

dchiller:
> Aren't these all just the same script?

Aha! I just edited the main description with a comment to this effect!

dchiller:
Ok, since this is your initial reaction too, I think it makes sense to change here.

jacobdgm:
this looks great - thanks for setting this up, @dchiller! Does the documentation need to be updated at all? I.e., I see that the Postgres `Dockerfile` has been updated - this only changes permissions for in the container, correct, and doesn't automatically cause the crontab to be updated on `git pull` (or something similar), does it?

jacobdgm:
also, this is currently configured to run on Production, yes? Do we want it to run on Staging too (not necessary, but I don't see any harm in it)? If so, we'll need documentation for updating `BACKUP_DIR` to work on staging.

dchiller:
> Does the documentation need to be updated at all?

The documentation looks good to me still. For example, when this is pulled to production, one will still manually update the crontab. 

> Do we want it to run on Staging too (not necessary, but I don't see any harm in it)? 

I agree there's no harm, but no one is ever adding data (besides developers making test sources, etc) to staging right? So I'm not sure when we'd ever want a back-up of staging vs of production.

PR #1236, opened 2023-12-27
Move source title to the left of source siglum

### LINKED ISSUE #1223: Source Detail: Siglum column should be to the right of Source column

lucasmarchd01:
Fixes #1223 
image


dchiller:
The header should also change right?

lucasmarchd01:
> The header should also change right?

Yes, thanks for catching that!

jacobdgm (approved):

PR #1237, opened 2023-12-29
Add css styling for autocomplete widgets

### LINKED ISSUE #1199: Chant Edit (and maybe also Create): Office autocomplete field is sometimes too wide
Screenshot 2023-12-07 at 11 33 33 AM

Steps to recreate: take a chant (do this in a test source, obviously) and choose an office with a long description (e.g. "X", as above). Save, and navigate back to the edit page. If the already-selected office is really long, the autocomplete field will be really wide. For a given page load, the width of the autocomplete field will stay the same no matter what you select, but if you select an office with a shorter description, save, and navigate back to the edit page, the width will be back to normal. Selecting an office with a moderate-length description (e.g., "E") leads to a widget that is somewhat wider (but not wide enough to fully accommodate the description, oddly)

@lucasmarchd01 and I briefly talked about this, and we think we have a solution. The library we're using for the autocomplete fields is rather finicky, so it will not be easy to change the dimensions of the field with styling, say. But we could truncate the description at 30 characters, say, ensuring that none of the options are too long.

lucasmarchd01:
Looking into #1199, I found that a general recommendation for the select2 autocomoplete widgets is to add css styling that fits the widget to 100% of the container width. This recommendation Fixes #1199.

The original idea was to truncate the labels in the widget. I tried this and found that it did in fact truncate the results in the widget dropdown but not the label that shows inside the widget if the field already has a selection.

image

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -127,6 +127,9 @@
         .guillemet::before {
             content: "» "
         }
+        .select2 {
+            width: 100%!important;
+          }
 
         /* this class is used for displaying results under the global search bar */
         .list-group-item-action:hover {


jacobdgm (approved):
As we discussed in person a second ago, ideally the Differentiae Database field would fit neatly among the others, but on the whole, this is good.

PR #1245, opened 2024-01-05
Change URI construction for Chant List and Chant Inventory

### LINKED ISSUE #1201: We should return a 404 response when a user arrives at `/chants/` without query parameters

### LINKED ISSUE #1202: We should consider changing the way Chant List URIs are constructed

### LINKED ISSUE #1210: On the Chant List page, we should return a 400 response when query parameters are missing instead of a 404 response
As pointed out by @ahankinson on #1201, the correct response code is 400 when a user visits `/chants` without specifying a `?source=`, and not 404 as is now the case.

The 404 is better than the 500 that we had previously raised, but ideally we would return a 400 response here.

lucasmarchd01:
This pull request modifies our url paths for the `chant-list` page from `/chants/?source=<source id` to `/source/<source id>/chants`. This makes it clear that the list of chants belongs to a specific source. A similar change is made for the `chant-index` page, which is now renamed to `chant-inventory`. Here is a summary of the changes:

- Change chant-list URI construction
- Change chant-index URI construction
- Modify templates to reflect changes in URI construction
- Update views and tests to reflect changes in URI construction

We should consider still setting up redirects for backwards compatibility.

Fixes #1202, Fixes #1201, Fixes #1210


ahankinson:
How many people's bookmarks will break with this change? Should the redirects be included in the change?

lucasmarchd01:
> How many people's bookmarks will break with this change? Should the redirects be included in the change?

I'm not sure how many people would bookmark the chant-list or chant-inventory pages, but I suspect there might be some. It might make more sense to fix it in this PR rather than making a new issue. (Changing to draft)

jacobdgm:
> How many people's bookmarks will break with this change? Should the redirects be included in the change?

+1 here - redirects should definitely be included as part of this PR.

Especially if this redirect is a function-based view like our other redirect views, it should be really easy to return a `400 - Bad Request` if the user doesn't include a `?source=` parameter (allowing us to address #1210 at the same time).

django/cantusdb_project/main_app/urls.py
```diff
@@ -72,7 +72,11 @@
     DifferentiaAutocomplete,
     ProvenanceAutocomplete,
     ProofreadByAutocomplete,
+    handle400,
```
ahankinson: I don't think you need these imports since you're not using these directly. You could change your code below to not express these as strings, but the way it is now they're unused.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -805,8 +806,14 @@ def redirect_node_url(request, pk: int) -> HttpResponse:
     raise Http404("No record found matching the /node/ query.")
 
 
+def handle400(request, exception):
+    print(f"Status Code: {exception.status_code}")
```
ahankinson: Errant print statement. This should be removed.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -805,8 +806,14 @@ def redirect_node_url(request, pk: int) -> HttpResponse:
...
+    return render(request, "400.html", status=400)
+
+
 def handle404(request, exception):
-    return render(request, "404.html")
+    print(f"Status Code: {exception.status_code}")
```
ahankinson: Also remove here.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -959,6 +966,36 @@ def redirect_documents(request) -> HttpResponse:
     return redirect(new_path)
 
 
+def redirect_chant_list(request) -> HttpResponse:
+    source_id: str = request.GET.get("source")
```
ahankinson:
```suggestion
    source_id: Optional[str] = request.GET.get("source")
```

django/cantusdb_project/main_app/views/views.py
```diff
@@ -959,6 +966,36 @@ def redirect_documents(request) -> HttpResponse:
...
+            "feast": feast_id,
+            "genre": genre_id,
+            "folio": folio,
+            "search_text": search_text,
+        }.items()
+        if value is not None
+    }
+
+    query_string: str = urlencode(params)
+    url: str = "{}{}".format(base_url, f"?{query_string}" if query_string else "")
```
ahankinson: It's a bit weird to mix `format()` calls with f-string calls.

django/cantusdb_project/main_app/views/views.py
```diff
@@ -959,6 +966,36 @@ def redirect_documents(request) -> HttpResponse:
...
+
+    base_url: str = reverse("chant-list", args=[source_id])
+
+    # optional search params
+    feast_id: str = request.GET.get("feast")
+    genre_id: str = request.GET.get("genre")
+    folio: str = request.GET.get("folio")
+    search_text: str = request.GET.get("search_text")
+
+    params: dict = {
```
ahankinson: You should move the dictionary construction outside of the comprehension. It's clearer, and actually takes fewer lines.

```
d = {
  "feast": feast_id,
  "genre": genre_id,
  "folio": folio,
  "search_text": search_text,
}
params = {k: v for k, v in d.items() if v is not None}

File: django/cantusdb_project/main_app/views/views.py [hunk 7]
```diff
@@ -959,6 +966,36 @@ def redirect_documents(request) -> HttpResponse:
...
+def redirect_chant_list(request) -> HttpResponse:
+    source_id: str = request.GET.get("source")
+    if source_id is None:
+        # source parameter must be provided
+        raise BadRequest("Source parameter must be provided")
+
+    base_url: str = reverse("chant-list", args=[source_id])
+
+    # optional search params
+    feast_id: str = request.GET.get("feast")
```
ahankinson: These should all be `Optional[str]`

```suggestion
    feast_id: Optional[str] = request.GET.get("feast")
```

A good rule is that if you need to check if it's `None` later, you must have marked it as Optional.

(Are you using any sort of type-checker, like mypy, when you develop? It helps catch these things. Also a good IDE will flag any unused imports, etc.)

ahankinson (changes requested):

lucasmarchd01:
Thanks @ahankinson and @jacobdgm for the helpful suggestions!

In this update, I removed the custom Django error handlers. After thorough testing, it was observed that these handlers were not being called, and Django is able to find and display error pages manually.

Additionally, our application doesn't have any specific requirements or special handling that necessitates custom error handlers. The removal of these custom handlers has no impact on the application's error handling capabilities, and Django is able to handle errors effectively without them.

I have also added tests for the chant-list and chant-inventory redirects.

django/cantusdb_project/main_app/urls.py
```diff
@@ -73,6 +73,8 @@
     ProvenanceAutocomplete,
     ProofreadByAutocomplete,
 )
+from django.conf.urls import handler400, handler404
```
ahankinson: Are these used?

django/cantusdb_project/main_app/templates/400.html
```diff
@@ -0,0 +1,39 @@
+{% load static %}
+<html lang="en">
```
ahankinson, line 2: Is this file still used too?

django/cantusdb_project/main_app/urls.py
ahankinson: I see you’ve removed them — great! If you’re not using a code editor that highlights these sort of syntax or style errors, you should. They are widely available and help catch these sorts of things before it needs “human linting”

django/cantusdb_project/main_app/views/views.py
```diff
@@ -959,6 +956,43 @@ def redirect_documents(request) -> HttpResponse:
...
+        "genre": genre_id,
+        "folio": folio,
+        "search_text": search_text,
+    }
+    params: dict = {k: v for k, v in d.items() if v is not None}
+
+    query_string: str = urlencode(params)
+    url: str = f"{base_url}?{query_string}" if query_string else base_url
+
+    return redirect(url)
```
dchiller: this issues a temporary redirect by default, but I think this is intended to be permanent?

ahankinson: Good catch

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2917,6 +2934,22 @@ def test_dd_column(self):
...
+        cantus_segment = make_fake_segment(id=4063)
+        source = make_fake_source(segment=cantus_segment)
+        source_id = source.id
+
+        url = reverse("redirect-chant-index")
+        response = self.client.get(f"{url}?source={source_id}")
+        self.assertRedirects(response, reverse("chant-inventory", args=[source_id]))
+
+    def test_redirect_without_source_parameter(self):
+        url = reverse("redirect-chant-index")
```
jacobdgm: The main place this needs to be changed is in `urls.py`, but this view should be `"redirect-chant-inventory", shouldn't it?

More generally, shouldn't this view now be called `SourceInventoryView`, and live in `views/source.py`? (Since/if this is the case, we maybe still want the redirect view to be "redirect-chant-index" or "-inventory", since we're redirecting from the old way the view was set up, with `/chant/` in the URL path...)

django/cantusdb_project/main_app/urls.py
```diff
@@ -293,6 +293,11 @@
         ChantInventoryView.as_view(),
         name="chant-inventory",
     ),
+    path(
+        "index/",
+        views.redirect_index_list,
```
jacobdgm: Not a fan of `redirect_index_list` as the name of the view....

jacobdgm (approved):
I think it's within the scope of this PR to rename the affected views to something that more accurately indicates what they do.

django/cantusdb_project/main_app/templates/400.html
lucasmarchd01: Our static file is used to load the background image displayed on the custom error pages.

django/cantusdb_project/main_app/views/views.py
lucasmarchd01: Yes thanks!

django/cantusdb_project/main_app/templates/400.html
dchiller: I am also a bit confused by this. It looks like `400.html` is introduced with this PR, but not referenced as far as I can tell anywhere else in the changes? Does django look for a `400.html` by default? 

Updated: It seems like django doesn't and that a view would need to be created and set as `handler400` in `urls.py`. https://docs.djangoproject.com/en/5.0/ref/urls/#module-django.conf.urls

Updated again: It does...thanks @lucasmarchd01

django/cantusdb_project/main_app/urls.py
```diff
@@ -491,5 +501,3 @@
         name="differentia-autocomplete",
     ),
 ]
-
```
dchiller, line 505: Also, with the removal of this, will the custom 404.html be used?

django/cantusdb_project/main_app/templates/400.html
lucasmarchd01: I spent a while verifying what is happening with this before removing the error handlers, and found that when Django encounters a 400 error (or 404, etc.) it will locate the 400.html page in the template directory by default. The error handlers are for the case where you might want to customize the error view. Django is able to serve the correct html files even if they are not referenced.

https://docs.djangoproject.com/en/5.0/ref/views/#the-404-page-not-found-view

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -5539,7 +5573,7 @@ def test_incipit_search(self):
     def test_cantus_id_search(self):
         chant_with_normal_cantus_id = make_fake_chant(cantus_id="012345")
         chant_with_numerals_in_incipit = make_fake_chant(
-            incipit="0 me! 0 my! This is unexpected!"
+            cantus_id="223455", incipit="0 me! 0 my! This is unexpected!"
```
jacobdgm: Is this Black compliant?

Also, this diff seems to be comparing to a not fully up-to-date `develop` branch... I don't know why this should be the case. Maybe GitHub will recalculate things when this PR stops being a draft?

dchiller: speaking of which...note that black's 2024 stable style guide was released in the last few days, so an upgrade might be in order.

lucasmarchd01: > Is this Black compliant?
> 
> Also, this diff seems to be comparing to a not fully up-to-date `develop` branch... I don't know why this should be the case. Maybe GitHub will recalculate things when this PR stops being a draft?

The potential conflict should be resolved now.

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -267,10 +267,10 @@ <h4>List of melodies</h4>
                                 </select>
 
                                 {% if previous_folio %}
-                                    <a href="{% url "chant-list" %}?source={{ source.id }}&folio={{ previous_folio }}">{{ previous_folio }} <</a>
+                                    <a href="{% url "chant-list" source.id %}?folio={{ previous_folio }}">{{ previous_folio }} <</a>
                                 {% endif %}
                                 {% if next_folio %}
-                                    &nbsp;<a href="{% url "chant-list" %}?source={{ source.id }}&folio={{ next_folio }}">> {{ next_folio }}</a>
+                                    &nbsp;<a href="{% url "chant-list" source.id %}?folio={{ next_folio }}">> {{ next_folio }}</a>
```
ahankinson, line 273: You can use keyword arguments in the `url` template tag so that you don't need to do string concatenation here...

Issue #564, opened 2023-02-21
Create JSON APIs for all models in the database


jacobdgm:
In doing some research for #563, I discovered a situation where Drupal and Django handle their databases differently, and this will cause a problem for our json-node API.

OldCantus appears to guarantee that all objects, whether they are sequences, chants, sources, etc., are all assigned a unique ID when they are created. In NewCantus, however, all sources have unique IDs relative to other sources, but there's nothing stopping a new source from being assigned same ID as an existing chant, say.

Currently, in our json_node_export view ([code](https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/main_app/views/views.py#L486)), we naively assume a given ID will match only a single object in the database - we check if it matches an existing chant, then sequence, then source. As soon as we find a match, we return information on it (it's worth noting that OldCantus's version of the API can also return information on indexers and news articles, which we don't currently do).

This will work fine up until people start adding new objects to the database in NewCantus. As people begin creating new sources, chants, etc., these ID duplications will become more and more common.

I can think of a few possible approaches to resolving this:
- before we make the big switch from OldCantus to New, we set up some validation step to ensure that IDs are unique among all objects
  - this would allow people to continue to interact with the API as they always have
  - this also feels like it will be quite messy/painful
- alternatively, we can change the URL of the API, to require whoever is sending the request to specify the type of object that should be returned
  - if we do this, we need to let everyone who currently uses the API know about the change
  - I might advocate for URLs to be patterned something along the lines of `/chant/123456/json/`, but there would be several reasonable ways to set this up

jacobdgm:
@dchiller and I just realized that CU needs to get at provenances' `name`s, so while resolving this issue, we can also set up a `/provenance/123456/json/` path to facilitate this (the current API doesn't support returning information on Provenances at all).

jacobdgm:
Our current solution for preventing /node/ URL collisions solves this issue.

I still like the idea of changing the paths of the URLs to function better as URIs. If we choose to do this, we can probably keep both URL paths available, or redirect requests to the new URL, similar to what we do for /node/ URLs currently. Marking this as an enhancement.

jacobdgm:
Increasing the priority to medium because of https://github.com/DDMAL/cantus/issues/808#issuecomment-1883989468.

jacobdgm:
[...]

PR #1251, opened 2024-01-11
Create provenance_json_export and notation_json_export APIs

jacobdgm:
This PR goes partway toward f\ixing #564. It includes two new APIs, for exporting information on CantusDB's Provenances and Notations.

Now that I've figured out how to set this up (deciding on how best to approach the tests was actually the main stumbling block), it will be easy to set up similar APIs for all our other objects, which will be useful for the general LinkedMusic project as well as (potentially) Cantus Ultimus (though I think the APIs for the two models in question are the ones most immediately needed by CU).

Sample output for `/notation/3870/json`:
```json
{
    "id": 3870,
    "date_created": "2020-07-09T21:21:57.919Z",
    "date_updated": "2020-07-09T21:21:57.921Z",
    "created_by_id": null,
    "last_updated_by_id": null,
    "name": "dtrns (???)"
}
```

Sample output for `/provenance/3622/json`:
```json
{
    "id": 3622,
    "date_created": "2020-07-09T21:22:28.612Z",
    "date_updated": "2020-07-09T21:22:28.613Z",
    "created_by_id": null,
    "last_updated_by_id": null,
    "name": "St-Lambrecht"
}
```

Figured I would get feedback on these two before implementing the rest. @lucasmarchd01 and @dchiller, I'd appreciate your thoughts!

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -744,6 +744,19 @@ def json_node_export(request, id: int) -> JsonResponse:
     return HttpResponseNotFound()
 
 
+def provenance_json_export(request, id: int) -> JsonResponse:
+    """
+    Return all fields of the provenance with the specified ID
+    """
+
+    provenance_qs: QuerySet[Provenance] = Provenance.objects.filter(id=id)
+    if len(provenance_qs) == 0:
+        return HttpResponseNotFound()
+
+    values: dict = dict(*provenance_qs.values())
+    return JsonResponse(values)
+
+
 def articles_list_export(request) -> HttpResponse:
     """Returns a list of URLs of all articles on the site
 
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -744,6 +744,19 @@ def json_node_export(request, id: int) -> JsonResponse:
     return HttpResponseNotFound()
 
 
+def notation_json_export(request, id: int) -> JsonResponse:
+    """
+    Return all fields of the notation with the specified ID
+    """
+
+    notation_qs: QuerySet[Notation] = Notation.objects.filter(id=id)
+    if len(notation_qs) == 0:
+        return HttpResponseNotFound()
+
+    values: dict = dict(*notation_qs.values())
+    return JsonResponse(values)
+
+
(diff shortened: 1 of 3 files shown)


dchiller (approved):

django/cantusdb_project/main_app/views/views.py
```diff
@@ -744,6 +744,32 @@ def json_node_export(request, id: int) -> JsonResponse:
     return HttpResponseNotFound()
 
 
+def notation_json_export(request, id: int) -> JsonResponse:
+    """
+    Return all fields of the notation with the specified ID
+    """
+
+    notation_qs: QuerySet[Notation] = Notation.objects.filter(id=id)
```
ahankinson, line 752: https://docs.djangoproject.com/en/5.0/topics/http/shortcuts/#get-object-or-404 ?

django/cantusdb_project/main_app/views/views.py
```diff
@@ -744,6 +744,32 @@ def json_node_export(request, id: int) -> JsonResponse:
...
+def provenance_json_export(request, id: int) -> JsonResponse:
+    """
+    Return all fields of the provenance with the specified ID
+    """
+
+    provenance_qs: QuerySet[Provenance] = Provenance.objects.filter(id=id)
+    if len(provenance_qs) == 0:
+        return HttpResponseNotFound()
+
+    values: dict = dict(*provenance_qs.values())
```
ahankinson, line 769: I feel like I’m missing something here. You’re doing a filter on an id, which likely means you want to do a get instead.

A QuerySet returns a list-like object, but you’re expanding it to a dict. I’m not sure how that’s possible but I guess it is?

Couldn’t everything here be replaced with `get_or_404(id=id).values()`, possibly with a simple cast to dict()?

jacobdgm: I followed the code already existing around https://github.com/DDMAL/CantusDB/blob/60606abb8a3f3a95837138fcb62c72c30dd098b3/django/cantusdb_project/main_app/views/views.py#L733 - While querysets can easily be converted into dictionaries via `.values()`, individual `Provenance` objects don't have a `.values()` method.

At the very least, I should have left a comment about why I was taking this approach. But even with a comment, this feels hacky - perhaps the correct approach here is to manually define a dictionary for each view.

PR #1254, opened 2024-01-12
Refactor provenance- and notation-json-export APIs

jacobdgm:
This PR amends the changes in #1251 based on comments on that PR after it was merged:

- `notation_` and `provenance_json_export` now use `get_object_or_404` instead of `queryset.filter(id=id)`
- the keys in the JsonResponse are now specified explicitly, rather than just getting all the fields for the model with `queryset.values()`. This change was necessitated by the use of `get_object_or_404`, but also has the effect of ensuring that these APIs will return data in the same format even if we make changes to the models in question.
- several keys were updated: `"created_by_id"` is now `"created_by"`, and `"last_updated_by_id"` is now `"last_updated_by"`.

Coming along for the ride is a commit updating the formatting of `views/chant.py` to comply with Black.

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -783,12 +783,25 @@ def provenance_json_export(request, id: int) -> JsonResponse:
     Return all fields of the provenance with the specified ID
     """
 
-    provenance_qs: QuerySet[Provenance] = Provenance.objects.filter(id=id)
-    if len(provenance_qs) == 0:
+    try:
+        provenance: Provenance = Provenance.objects.get(id=id)
+    except Provenance.DoesNotExist:
         return HttpResponseNotFound()
 
-    values: dict = dict(*provenance_qs.values())
-    return JsonResponse(values)
+    User = get_user_model()
+    created_by: Optional[User] = provenance.created_by
+    last_updated_by: Optional[User] = provenance.last_updated_by
+
+    data = {
+        "id": provenance.id,
+        "name": provenance.name,
+        "date_created": provenance.date_created,
+        "date_updated": provenance.date_updated,
+        "created_by": created_by.id if created_by else None,
+        "last_updated_by": last_updated_by.id if last_updated_by else None,
+    }
+
+    return JsonResponse(data)
 
 
 def articles_list_export(request) -> HttpResponse:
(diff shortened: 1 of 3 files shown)


jacobdgm:
(documentation on these APIs can be found at https://github.com/DDMAL/CantusDB/wiki/APIs#json-export-apis-for-various-models!)

django/cantusdb_project/main_app/views/views.py
```diff
@@ -746,28 +747,53 @@ def json_node_export(request, id: int) -> JsonResponse:
...
 def provenance_json_export(request, id: int) -> JsonResponse:
     """
-    Return all fields of the provenance with the specified ID
+    Return JsonResponse containing several key:value pairs
+    for the provenance with the specified ID
     """
 
-    provenance_qs: QuerySet[Provenance] = Provenance.objects.filter(id=id)
-    if len(provenance_qs) == 0:
+    try:
```
dchiller: this one doesn't use `get_object_or_404`?

jacobdgm: hm... something has gone wrong here. Let me investigate.

jacobdgm: I was so excited to update the docstring for the method, it seems, that I overlooked the raison d'être of the PR. Thanks for catching this.

dchiller (approved):

django/cantusdb_project/main_app/views/views.py
```diff
@@ -746,28 +747,50 @@ def json_node_export(request, id: int) -> JsonResponse:
...
 
-    notation_qs: QuerySet[Notation] = Notation.objects.filter(id=id)
-    if len(notation_qs) == 0:
-        return HttpResponseNotFound()
+    notation: Notation = get_object_or_404(Notation, id=id)
 
-    values: dict = dict(*notation_qs.values())
-    return JsonResponse(values)
+    User = get_user_model()
+    created_by: Optional[User] = notation.created_by
```
ahankinson, line 757: Sorry -- I can't resist. :)

You don't need to get the user model. If you want it only for the type annotation, it should be imported. But you only need the ID, which, as it happens, is actually the value of the field, and it is `None` without a value anyway. So this should be all you need:

`"created_by": notation.created_by_id`

(&tc.)

jacobdgm: This is a useful bit of light django magic! Thank you!

PR #1263, opened 2024-01-17
Fixes to certificate renewal

### LINKED ISSUE #1262: [Minor] fixes needed for certificate renewal process
These include:

- adding the path where certificates live on the server (`./certificates`) to `.gitignore`
- adjusting certificate volume paths in `docker-compose.yml` and certificate path configurations in the nginx configuration file
- verifying the certificate renewal command in `cron.txt`

dchiller:
This PR: 

- adds `./certificates` to `.gitignore` so that certificates are not accidentally committed to the repo (even though we would rarely open a PR from changes made on staging or production)
- modifies certificate paths in `docker-compose.yml` and the nginx config file
- uses `exec` rather than `run` in the certificate renewal command

Closes #1262.

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -16,8 +16,8 @@ server {
 
     listen 443 default_server http2 ssl;
 	
-    ssl_certificate /etc/nginx/ssl/live/certificates/staging.cantusdatabase.org.crt;
-    ssl_certificate_key /etc/nginx/ssl/live/certificates/staging.cantusdatabase.org.key;
+    ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
+    ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
 
     location / {
         proxy_pass http://django:8000;
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -16,8 +16,8 @@ server {
 
     listen 443 default_server http2 ssl;
 	
-    ssl_certificate /etc/nginx/ssl/live/cantusdatabase.org/fullchain.pem;
-    ssl_certificate_key /etc/nginx/ssl/live/cantusdatabase.org/privkey.pem;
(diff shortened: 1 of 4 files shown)


config/nginx/conf.d/cantusdb.conf
```diff
@@ -16,8 +16,8 @@ server {
 
     listen 443 default_server http2 ssl;
 	
-    ssl_certificate /etc/nginx/ssl/live/cantusdatabase.org/fullchain.pem;
-    ssl_certificate_key /etc/nginx/ssl/live/cantusdatabase.org/privkey.pem;
+    ssl_certificate /etc/nginx/ssl/live/certificates/staging.cantusdatabase.org.crt;
+    ssl_certificate_key /etc/nginx/ssl/live/certificates/staging.cantusdatabase.org.key;
```
jacobdgm: We should have these set up to work on Production - if a developer misses a step and thus takes down a server, we'd rather that server be Staging than Production.

jacobdgm: Note that Production has two addresses - `cantusdatabase.org` and `www.cantusdatabase.org`. Do we know how this will work - are there going to be two separate certificates and two separate keys in two separate directories? Are we going to be able to configure nginx to handle this situation?

jacobdgm (changes requested):

config/nginx/conf.d/cantusdb.conf
dchiller: Good point...and good question. Let me look into it.

dchiller: Ok, based on what I am seeing a few places (incl. https://stackoverflow.com/questions/50496214/how-can-i-generate-a-letsencrypt-cert-for-multiple-domain-variants and lego's issue 228) it looks to me like we should expect certificates on production to be stored at `/etc/nginx/ssl/live/certificates/cantusdatabase.org`...I'll make that change here.

dchiller:
Once this is good to merge, let's make sure to update the wiki to reflect that this change (the adjustment to the `ssl_certificate` and `ssl_certificate_key` parameters in `cantusdb.conf`) will need to be made when deploying to Staging.

jacobdgm (approved):

PR #1267, opened 2024-01-18
Create webroot path during `nginx` container build

dchiller:
`config/nginx/conf.d/cantusdb.conf` configures `/var/www/lego` as the location of the challenge for certificate creation and renwal. This update to the nginx `Dockerfile` ensures that this path exists.


jacobdgm (approved):
having tested this change on staging, we're confident this works.

PR #1276, opened 2024-01-23
Create save_concordances_to_file command

jacobdgm:
This PR adds a new management command, `save_concordances_to_file`, working towards implementing #1209.

The command compiles a list of all the unique Cantus IDs in the database, then for each ID, it returns an array of JSON objects, each representing a chant matching that ID. All published chants in the database should thus be returned, grouped by Cantus ID.

I chose to put the list of Cantus IDs with their chants in a nested JSON object at `["data"]`, and I included the date the file was last updated in another nested object at `["metadata"]`. Open to suggestions as to where/whether to include the date last updated, and also to additional metadata that could be useful to developers.

I didn't take the time to transplant the logic into a view so I could profile it with DDT, but the command runs to completion in about 1.5 minutes on my local machine - every time I ran it, it took about 75s to compute the concordances, and about 20s to write the results to file. I'm hoping that, by using `.prefetch_related()` and `.values()`, I've avoided any unnecessary database queries (one query to establish the list of Cantus IDs, and one query per Cantus ID); everything should already be a `dict` by the time it reaches `[build_cid_dictionary(chant) for chant in chants_values]`. I'm open, of course, to suggestions as to how to further optimize this.

Here's a substantially-trimmed example of the file created at `cantusdb_project/api_cache/all_concordances.json`:
```
{
    "data": {
        ...
        "006014c": [
            ...
            {
                "siglum": "F-R 248",
                "srclink": "https://cantusdatabase.org/source/123662/",
                "chantlink": "https://cantusdatabase.org/chant/442642/",
                "folio": "079v",
                "sequence": 1,
                "incipit": "Jussu Herodis amputatum est",
                "feast": "Decoll. Jo. Bapt.",
                "genre": "V",
                "office": "M",
                "position": "01",
                "cantus_id": "006014c",
                "image": null,
                "mode": "1",
                "full_text": null,
                "melody": null,
                "db": "CD"
            },
            {
                "siglum": "F-Pn Lat. 12044",
                "srclink": "https://cantusdatabase.org/source/123628/",
                "chantlink": "https://cantusdatabase.org/chant/402573/",
                "folio": "180r",
                "sequence": 10,
                "incipit": "Jussu Herodis amputatum est caput ",
                "feast": "Decoll. Jo. Bapt.",
                "genre": "V",
                "office": "M",
                "position": "01",
                "cantus_id": "006014c",
                "image": "http://gallica.bnf.fr/ark:/12148/btv1b6000531z/f363.image",
                "mode": "1",
                "full_text": "Jussu Herodis amputatum est caput Joannis in carcere quo audito discipuli ejus",
                "melody": "1---h--hghg-gf---g--gh--g---g--g--gh--g---g---g--g---g--hF--gh---gH---h7--hijh--h---gf---gh--hj--h---h--hghg--gf--fgh---hgfe-gh--gf---3---d--fgh--g---3",
                "db": "CD"
            },
            ...
        ],
        ...
    },
    "metadata": {
        "last_updated": "2024-01-23T19:03:09.528799"
    }
}
```

And STDOUT looks something like this:
```
root@acb3b5443859:/code/django/cantusdb_project# python manage.py save_concordances_to_file
Running save_concordances_to_file at 2024-01-23T19:59:50.398818.
  save_concordances_to_file: Computing concordances information for 30589 Cantus IDs
  save_concordances_to_file: attempting to make directory at api_cache to hold cache: successfully created directory at api_cache.
  save_concordances_to_file: Writing concordances information to api_cache/all_concordances.json at 2024-01-23T20:00:42.436162.
save_concordances_to_file: Concordances successfully written to api_cache/all_concordances.json.
```

Things that remain to be done toward resolving #1209 after this PR is reviewed and merged:
- [x] Configure Nginx to find and serve the file at `cantusdb_project/api_cache/all_concordances.json` (completed with #1280)
- [x] Add a new cron job to run `save_concordances_to_file` periodically (completed with #1280)
- [x] Add documentation to https://github.com/DDMAL/CantusDB/wiki/APIs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,7 @@ django/cantusdb_project/Drupal_scripts
 
 # in case you're working on importing data from a database dump
 *.sql
+
+# cached files for APIs
+api_cache/
+
--- a/django/cantusdb_project/main_app/management/commands/save_concordances_to_file.py
+++ b/django/cantusdb_project/main_app/management/commands/save_concordances_to_file.py
@@ -2,6 +2,7 @@ import json
 import os
 from sys import stdout
 from datetime import datetime
+from collections import defaultdict
 from django.db.models.query import QuerySet
 from django.core.management.base import BaseCommand
 from main_app.models import Chant
@@ -10,103 +11,41 @@ from main_app.models import Chant
 class Command(BaseCommand):
     def handle(self, *args, **kwargs) -> None:
         CACHE_DIR: str = "api_cache"
-        FILEPATH: str = "api_cache/all_concordances.json"
+        FILEPATH: str = "api_cache/concordances.json"
         start_time: str = datetime.now().isoformat()
         stdout.write("Running save_concordances_to_file " f"at {start_time}.\n")
-        concordance_info: dict = get_concordance_info_for_all_cantus_ids()
+        concordances: dict = get_concordances()
         write_time: str = datetime.now().isoformat()
         metadata: dict = {
             "last_updated": write_time,
         }
         data_and_metadata: dict = {
-            "data": concordance_info,
+            "data": concordances,
             "metadata": metadata,
         }
-        stdout.write(
-            "  save_concordances_to_file: attempting to make directory "
-            f"at {CACHE_DIR} to hold cache: "
-        )
+        stdout.write("Attempting to make directory " f"at {CACHE_DIR} to hold cache: ")
         try:
             os.mkdir(CACHE_DIR)
             stdout.write(f"successfully created directory at {CACHE_DIR}.\n")
         except FileExistsError:
             stdout.write(f"directory at {CACHE_DIR} already exists.\n")
-        stdout.write(
-            "  save_concordances_to_file: Writing concordances "
-            f"information to {FILEPATH} "
-            f"at {write_time}.\n"
-        )
+        stdout.write(f"Writing concordances to {FILEPATH} " f"at {write_time}.\n")
         with open(FILEPATH, "w") as json_file:
             json.dump(data_and_metadata, json_file)
-        stdout.write(
-            "save_concordances_to_file: Concordances successfully written to "
-            f"{FILEPATH}.\n\n"
-        )
+        stdout.write(f"Concordances successfully written to {FILEPATH}.\n\n")
 
 
-def get_concordance_info_for_all_cantus_ids() -> dict:
-    """
-    Create a dictionary with information on all published chants
-    in the database, grouped by Cantus ID
+def get_concordances() -> dict:
+    DOMAIN: str = "https://cantusdatabase.org"
 
-    Returns:
-        dict: a dictionary, with each key being one Cantus ID, and each
-        value being a list of chants in the database matching that ID
-    """
-    cantus_ids = list(get_set_of_cantus_ids())
-    cantus_id_count = len(cantus_ids)
-    stdout.write(
-        f"  save_concordances_to_file: Computing concordances "
-        f"information for {cantus_id_count} Cantus IDs.\n"
-    )
-    dictionary: dict = {
-        cantus_id: get_concordance_info_for_cantus_id(cantus_id)
-        for cantus_id in cantus_ids
-    }
-
-    return dictionary
-
-
-def get_set_of_cantus_ids() -> set[str]:
-    """
-    Create a set of all Cantus IDs in the database.
-
-    Returns only those Cantus IDs associated with published chants,
-    and ignoring those from the Bower Database.
-
-    Returns [set[str]]: a set of strings, each string being
-        a single Cantus ID
-    """
+    stdout.write("Querying database for published chants\n")
     published_chants: QuerySet[Chant] = Chant.objects.filter(source__published=True)
-    published_cantus_ids: QuerySet[dict] = published_chants.values("cantus_id")
-    unique_cantus_ids: set[str] = set(c["cantus_id"] for c in published_cantus_ids)
-
-    return unique_cantus_ids
-
-
-def get_concordance_info_for_cantus_id(cantus_id: str) -> list[dict]:
-    """
-    Given a Cantus ID, return a list of dictionareis containing
-    information on all published chants in the database with that Cantus ID
-
-    Args:
-    - cantus_id [str]: the Cantus ID
-
-    Returns [list[dict]]: a list of dictionaries, with each dictionary representing
-    one chant
-    """
-    published_chants: QuerySet[Chant] = Chant.objects.filter(source__published=True)
-    chants_matching_cantus_id: QuerySet[Chant] = published_chants.filter(
-        cantus_id=cantus_id
-    ).prefetch_related(
-        # .prefetch_related() is necessary in order to be able to follow
-        # source__siglum, feast__name, etc. when using .values(), below
+    values: QuerySet[dict] = published_chants.select_related(
         "source",
         "feast",
         "genre",
         "office",
-    )
-    chants_values: QuerySet[dict] = chants_matching_cantus_id.values(
+    ).values(
         "id",
         "source_id",
         "source__siglum",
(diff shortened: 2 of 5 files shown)


django/cantusdb_project/main_app/management/commands/save_concordances_to_file.py
```diff
@@ -0,0 +1,153 @@
...
+    Create a set of all Cantus IDs in the database.
+
+    Returns only those Cantus IDs associated with published chants,
+    and ignoring those from the Bower Database.
+
+    Returns [set[str]]: a set of strings, each string being
+        a single Cantus ID
+    """
+    published_chants: QuerySet[Chant] = Chant.objects.filter(source__published=True)
+    published_cantus_ids: QuerySet[dict] = published_chants.values("cantus_id")
```
dchiller: A case for `QuerySet.distinct`?

dchiller:
With the caveats that this does what it says and will run once per day in 75 seconds so I think this approach is ok, something feels off about querying the database for each Cantus ID when what this is doing seems to be essentially a grouping operation...

ahankinson:
There's a whole lotta loopin' goin' on. 

Hints:

- Maybe `values_list` is useful?
- Don't overuse `values`. You can access the model objects directly. You don't alway have to convert the model data to a dict to use it. 
- If you get rid of the `values` calls you can use `select_related` which is much preferable to `prefetch_related`. 
- Distinct is an excellent suggestion. 
- [`cantus_id__in`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#in) might be something to test?

dchiller:
I sort of think in sql when doing a query like this, and I can write one that does this with a single query, which yeah, makes me think there must be some way to avoid that loop.

ahankinson:
In order to not just be a "reply guy" I took on the task of putting my proverbial money where my mouth is and see whether my hints are worth anything.

Here's what I came up with:

```python
def get_concordances() -> dict:
    DOMAIN: str = "https://cantusdatabase.org"
    print("Getting concordances")
    qs: QuerySet[Chant] = Chant.objects.filter(source__published=True)
    qsr = qs.select_related(
        "source",
        "feast",
        "genre",
        "office"
    ).values(
        "id",
        "source_id",
        "source__siglum",
        "folio",
        "c_sequence",
        "incipit",
        "feast__name",
        "genre__name",
        "office__name",
        "position",
        "cantus_id",
        "image_link",
        "mode",
        "manuscript_full_text_std_spelling",
        "volpiano",
    )
    # collections.defaultdict negates the need for doing a distinct() call. Since
    # dictionary keys are necessarily unique, and we simply need to add a list of
    # chants to that cantus id, then the default dict automatically creates a list()
    # for new key insertions. This means we can guarantee that a list is always present
    # in the dict, without first checking to see if the key is present.
    res = collections.defaultdict(list)
    print("Processing chants")
    for chant in qsr:
        res[chant["cantus_id"]].append({
                "siglum": chant["source__siglum"],
                "srclink": f"{DOMAIN}/source/{chant['source_id']}/",
                "chantlink": f"{DOMAIN}/chant/{chant['id']}/",
                "folio": chant['folio'],
                "sequence": chant['c_sequence'],
                "incipit": chant['incipit'],
                "feast": chant["feast__name"],
                "genre": chant["genre__name"],
                "office": chant["office__name"],
                "position": chant["position"],
                "cantus_id": chant["cantus_id"],
                "image": chant["image_link"],
                "mode": chant["mode"],
                "full_text": chant["manuscript_full_text_std_spelling"],
                "melody": chant["volpiano"],
                "db": "CD",
            }
        )

    print(f"Found {len(res.keys())} chants")
    # a defaultdict needs to be cast to a dict()
    return dict(res)
```

The results are (slightly different from the use of `print()`):

```
Running save_concordances_to_file at 2024-01-25T09:07:30.345187.
Getting concordances
Processing chants
Found 30586 chants
Done collecting concordance info
  save_concordances_to_file: attempting to make directory at api_cache to hold cache: directory at api_cache already exists.
  save_concordances_to_file: Writing concordances information to api_cache/all_concordances2.json at 2024-01-25T09:07:31.852346.
save_concordances_to_file: Concordances successfully written to api_cache/all_concordances2.json.
```

So this takes ~2s. The size of the output was the same (~191MB). 

To run this you just have to change L18 to:

```
concordance_info: dict = get_concordances()
```

So:
 - The `values()` call is useful -- it was taking much longer without it. 
 - `select_related()` works with the queryset and with `values()`
 - The output seems to be the same, roughly 191MB for both the old and new outputs (on my machine, using a dump from a couple weeks ago). But of course verification on your part is necessary.

Anyway, I hope this is helpful. It's not meant to step on toes, @jacobdgm, but really as a real-world example of how you can optimize the database lookups quite a bit. While it probably doesn't matter much in this specific instance, I think these patterns can help in other areas where you need to provide similar dumps.

jacobdgm:
> Anyway, I hope this is helpful. It's not meant to step on toes, @jacobdgm, but really as a real-world example of how you can optimize the database lookups quite a bit. While it probably doesn't matter much in this specific instance, I think these patterns can help in other areas where you need to provide similar dumps.

This is excellent - thanks @ahankinson (and @dchiller)!

The thought had occurred to me that it should in principle be possible to do this in a single query (as initially noted by @dchiller, and then implemented by @ahankinson). One main takeaway for me from all this is that there are situations where comprehensions can run slower than `for` loops, especially in cases where the comprehension necessitates separate database queries.

Again, thanks to both of you for these suggestions. I'll be back soon with an update to this PR following this approach.

ahankinson:
> One main takeaway for me from all this is that for loops can be more efficient than comprehensions, especially in cases where the comprehension leads to many separate database queries.

Hm. That's not really a takeaway that I would agree with. Comprehensions are more efficient than for-loops, but they also hide the fact in your code that you're looping. You were iterating three times more than you needed to -- that's the source of the inefficiency here. Trying to speed this up by using a list comprehension is like trying to use a slightly bigger spoon to drain the ocean.

(You were also triggering a query for every Cantus ID, which was actually the biggest cause of slowdown, but I suspect you know this.)

jacobdgm:
I just pushed a refactored version of this command using @ahankinson's algorithm.

I also added tests.

On my local machine, it takes ~2 seconds to process all the chants and another ~20 to write them to file.

ahankinson:
I really can't help myself. I like these sorts of challenges! 

I tested it with `ujson` instead of the built-in `json`. On my machine the built-in `json` module took 5s to write the file; `ujson` took < 1s. 

I know I'm going a bit crazy with the optimizations here, and there's no obligation to do this. But I thought I would add that as a point of reference. (Also, you must not have an SSD if it takes 20s to write the file...)

```
python manage.py update_cached_concordances
Running update_cached_concordances at 2024-01-25T16:22:07.203500.
Querying database for published chants
Processing chants
All chants processed - found 30586 Cantus IDs
Attempting to make directory at api_cache to hold cache: directory at api_cache already exists.
Writing concordances to api_cache/concordances.json at 2024-01-25T16:22:09.281572.
Concordances successfully written to api_cache/concordances.json at 2024-01-25T16:22:10.048619.
```

.gitignore
```diff
@@ -120,3 +120,7 @@ django/cantusdb_project/Drupal_scripts
 
 # in case you're working on importing data from a database dump
 *.sql
+
+# cached files for APIs
+api_cache/
```
dchiller, line 125: I don't understand why we would expect this directory to exist? Won't this directory just inside the container?

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -0,0 +1,100 @@
...
+from collections import defaultdict
+from django.db.models.query import QuerySet
+from django.core.management.base import BaseCommand
+from main_app.models import Chant
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs) -> None:
+        CACHE_DIR: str = "api_cache"
+        FILEPATH: str = "api_cache/concordances.json"
```
dchiller: Unless I'm not understanding the use of these, it would make sense to have:

`FILEPATH: str = f"{CACHE_DIR}/concordances.json"`

otherwise these could diverge and I don't think you want them to?

jacobdgm:
> I know I'm going a bit crazy with the optimizations here, and there's no obligation to do this. But I thought I would add that as a point of reference. (Also, you must not have an SSD if it takes 20s to write the file...)

I like optimizing things too... :)

The machine I'm using certainly has an SSD. But trying it now, ujson cuts write time by a factor of ~10. Especially if we're going to cache the results of more of our APIs eventually, I think it's worth adding a new dependency for this. Thanks for this suggestion - I hadn't been aware of this possibility!

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -0,0 +1,100 @@
...
+            f"Concordances successfully written to {FILEPATH} at {end_time}.\n\n"
+        )
+
+
+def get_concordances() -> dict:
+    DOMAIN: str = "https://cantusdatabase.org"
+
+    stdout.write("Querying database for published chants\n")
+    published_chants: QuerySet[Chant] = Chant.objects.filter(source__published=True)
+    values: QuerySet[dict] = published_chants.select_related(
```
dchiller, line 46: Not critical, but I'm confused by this type annotation... a `QuerySet` of dicts?

dchiller (approved):

.gitignore
jacobdgm: I don't completely understand the second part of your question, but I'll say something that will hopefully address it: In the documentation, there are instructions to include a `devcontainer.json` file, used for an extension that causes files inside and outside of the container to be synchronized, allowing VS Code to run inside the container. A developer using this extension who runs this command will have a directory in this location, and we don't want it or its contents to be committed.

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: hm... once we've used `.values()`, the items within the queryset are `dicts` instead of `Chant` objects. I almost considered casting `values` here to a list, so I could annotate it `list[dict]`, but that seemed unnecessary.

I'm gathering you think it would be clearer to annotate it as `QuerySet[Chant]` again - is this a change you'd like me to make?

jacobdgm: Going to merge this so that I can clean up something I noticed in `requirements.txt` without creating merge conflicts. But I'd be happy to make this change in another small PR if you think it would improve the function.

.gitignore
dchiller: Hmmmm...ok. I'm not familiar with this extension, so that must by why.

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
dchiller: If your type checker fine with QuerySet[dict] that must be it...it just seems odd. Iterating through the QuerySet gives you dicts, but I think it would still be a QuerySet of a particular model

ahankinson: > Returns a QuerySet that returns dictionaries, rather than model instances, when used as an iterable.

https://docs.djangoproject.com/en/5.0/ref/models/querysets/#values

The type annotation is correct, I think.

ahankinson:
ujson is a (mostly) drop-in replacement for the built-in `json` module, so you can use it in other places as well. 

You can also, if you're feeling 👿 , do `import ujson as json` and not change any code...

PR #1278, opened 2024-01-25
Clean up test_views.py

### LINKED ISSUE #1244: Intermittent failure of AjaxSearchViewTest
```
======================================================================
FAIL: test_cantus_id_search (main_app.tests.test_views.AjaxSearchBarTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/django/cantusdb_project/main_app/tests/test_views.py", line 5563, in test_cantus_id_search
    self.assertEqual(len(non_matching_chants), 0)
AssertionError: 1 != 0

----------------------------------------------------------------------
```

One of the chants created in this test has a randomly chosen Cantus ID - occasionally, searching for `1`, as we do in the test, will find this chant.

jacobdgm:
This PR makes two improvements to our `test_views` test suite:
- We had two similar `test_chant_with_volpiano_with_no_incipit`s defined within `ChantDetailViewTest`. The second of these has been removed
- `AjaxSearchBarTest.test_cantus_id_search` has been refactored to be deterministic (we had been creating two chants, and only the first had a `cantus_id` specified. When the other chant's `cantus_id` began with a `1` (P=0.1), the test would fail. Now, both chants' Cantus IDs are specified). This fixes #1244.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -3282,22 +3282,6 @@ class SourceEditChantsViewTest(TestCase):
         chant.refresh_from_db()
         self.assertIs(chant.manuscript_full_text_std_proofread, True)
 
-    def test_chant_with_volpiano_with_no_incipit(self):
-        # in the past, a Chant Proofread page will error rather than loading properly when the chant has volpiano but no fulltext/incipit
-        source = make_fake_source()
-        chant = make_fake_chant(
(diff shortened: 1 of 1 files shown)


lucasmarchd01:
#1244 has a pending draft PR in #1245, but this should be a quicker and better way of getting the issue fixed!

lucasmarchd01 (approved):

dchiller (approved):

PR #1280, opened 2024-01-26
Configure nginx to serve cached concordances

### LINKED ISSUE #1209: We need a new API that displays concordances information for all chants in the database

jacobdgm:
This PR updates our nginx configuration to serve cached concordances created by our `update_cached_concordances` command. It fixes #1209. 

- A new docker volume, `api_cache`, has been created. `update_cached_concordances` now writes its output to this volume, and the nginx container has access to this volume too, in order to be able to serve the concordances
- nginx has been configured to return the cached concordances when it receives a request at `/concordances`
- `cron.txt` has been updated to run `update_cached_concordances` once per day

The cron changes haven't been tested - we can do this once these changes have made their way to the staging server.

After making the last commit implicated in this PR, I
- followed the steps in https://github.com/DDMAL/CantusDB/wiki/Getting-Started-with-Development, and rebuilt and restarted the docker containers
- navigated to localhost/concordances in my browser, and successfully downloaded a bunch of concordances in .json format that had been created yesterday
- ran `python manage.py update_cached_concordances`
- navigated to localhost/concordances again, downloaded a new batch of concordances, and saw that the timestamp had been updated
- in short, this is behaving exactly as expected and as hoped-for!

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -1,37 +1,6 @@
 server {
+
     listen 80;
-
-    server_tokens off;
-
-    location ^~ /.well-known/acme-challenge/ {
-        root /var/www/lego;
-    }
-
-    location / {
-        return 301 https://$host$request_uri;
-    }
-}
-
-server {
-    # Redirect all https traffic for mass.cantusdatabase.org to cantusdatabase.org
-    listen 443 ssl;
-
-    server_name mass.cantusdatabase.org;
-
-    ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
-    ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
-
-    location / {
-        return 301 https://cantusdatabase.org$request_uri;
-    }
-}
-
-server {
-
-    listen 443 default_server http2 ssl;
-	
-    ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
-    ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
 
     location / {
         proxy_pass http://django:8000;
@@ -49,6 +18,7 @@ server {
 
     location /concordances {
         alias /resources/api_cache/concordances.json;
+        expires 6h;
     }
  
     location = /style.css {
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -49,7 +49,7 @@ server {
 
     location /concordances {
         alias /resources/api_cache/concordances.json;
-        expires 6h;
+        expires modified +24h;
     }
  
     location = /style.css {
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -46,7 +46,11 @@ server {
     location /media {
         alias /resources/media/;
     }
-    
+
+    location /concordances {
+        alias /resources/api_cache/concordances.json;
+    }
+ 
     location = /style.css {
         root /;
     }
--- a/config/nginx/conf.d/cantusdb.conf
+++ b/config/nginx/conf.d/cantusdb.conf
@@ -1,6 +1,37 @@
 server {
-
     listen 80;
+
+    server_tokens off;
+
+    location ^~ /.well-known/acme-challenge/ {
+        root /var/www/lego;
+    }
+
+    location / {
+        return 301 https://$host$request_uri;
+    }
+}
+
+server {
+    # Redirect all https traffic for mass.cantusdatabase.org to cantusdatabase.org
+    listen 443 ssl;
+
+    server_name mass.cantusdatabase.org;
+
+    ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
+    ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
+
+    location / {
+        return 301 https://cantusdatabase.org$request_uri;
+    }
+}
+
+server {
+
+    listen 443 default_server http2 ssl;
+	
+    ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
+    ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
 
     location / {
         proxy_pass http://django:8000;
--- a/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
+++ b/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
@@ -10,7 +10,7 @@ from main_app.models import Chant
 
 class Command(BaseCommand):
     def handle(self, *args, **kwargs) -> None:
-        CACHE_DIR: str = "api_cache"
+        CACHE_DIR: str = "/resources/api_cache"
         FILEPATH: str = f"{CACHE_DIR}/concordances.json"
         start_time: str = datetime.now().isoformat()
         stdout.write(f"Running update_cached_concordances at {start_time}.\n")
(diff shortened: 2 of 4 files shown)


jacobdgm:
requesting review from @dchiller to make sure I haven't missed anything (we worked on this together yesterday), and from @ahankinson and @lucasmarchd01 to get the review of at least one person who didn't actively implement this change.

jacobdgm:
(oops - just realized I didn't commit the changes to cron.txt! Note, @dchiller, that I added only the first half of the command we wrote yesterday - we no longer need to copy files anywhere because we set up the volume that both django and nginx can access).

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -10,7 +10,7 @@
 
 class Command(BaseCommand):
     def handle(self, *args, **kwargs) -> None:
-        CACHE_DIR: str = "api_cache"
+        CACHE_DIR: str = "/resources/api_cache"
```
ahankinson: You should probably configure this by command-line parameter. Running this locally on my machine will put this in the root of my system, which is not good.

config/nginx/conf.d/cantusdb.conf
```diff
@@ -46,7 +46,11 @@ server {
     location /media {
         alias /resources/media/;
     }
-    
+
+    location /concordances {
```
ahankinson, line 50: What’s the actual URL that this will be served with?

ahankinson: Never mind, I see it in the PR

ahankinson:
- do you have gz or brotli compression enabled for JSON responses?
- Any value in having an expires header for client-side caches?

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: If this is saving to your root directory, I'm guessing you're not running the project in Docker?

In my local setup, this saves to `/resources/api_cache` within the django container, right where it needs to be to get picked up by the `api_cache` volume specified in our `docker-compose.yml` file.

Unless I'm missing something - considering this command has been designed to save the concordances to a Docker volume, and all our documentation makes it clear that the application is meant to be run in Docker, I feel like configuring this via the command line would just add complexity.

Would adding a comment here - something along the lines of "this is the path for the `api_cache` volume configured in `docker-compose.yml`" - help?

jacobdgm:
> do you have gz or brotli compression enabled for JSON responses?

Not to my knowledge! Let me look into this.

> Any value in having an expires header for client-side caches?

Probably - let me look into this!

Considering we know when the command will begin to run each day, as specified in `cron.txt`, but we don't know exactly how long it will take to finish, this strikes me as a difficult thing to get exactly right - I can imagine a situation where a client requests the concordances at just the wrong second, and ends up caching out-of-date-by-a-day concordances for a full day. So maybe it would be better to configure this to say the response expires after several hours? Or perhaps someone has already solved this, or there's a simple solution that hasn't occurred to me. @ahankinson, do you have a sense of how best to approach this?

jacobdgm:
I just pushed a new commit configuring Nginx to add an "Expires" header to the response headers. I set it to expire after 6h, but this will be an easy thing to change/update if we decide a different expiration policy would be better. Response headers now look something like

```
HTTP/1.1 200 OK
Server: nginx/1.24.0
Date: Tue, 30 Jan 2024 15:32:30 GMT
Content-Type: application/octet-stream
Content-Length: 192128923
Last-Modified: Fri, 26 Jan 2024 20:42:45 GMT
Connection: keep-alive
ETag: "65b41945-b73a79b"
Expires: Tue, 30 Jan 2024 21:32:30 GMT
Cache-Control: max-age=21600
Accept-Ranges: bytes
```

jacobdgm:
I spent some time looking into compressing the response; it looks like we could configure it similarly, but there are lots of associated options (see https://docs.nginx.com/nginx/admin-guide/web-server/compression/; perhaps there's a more concise guide for this somewhere?). I created a new issue (#1281) for this; suggesting we merge this PR as-is, and implement compression in a separate PR after we've had a bit more time to figure out the best approach.

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
ahankinson: Adding an output directory configuration is pretty standard. You can also set a default value if you want. But the option to change the location, should it be useful, is really handy.

(I don’t run docker. I’ve deployed a number of Django sites to production, and they didn’t need Docker either. I actually think this application doesn’t need docker, and that you would get better performance and have a much simpler deploy process if you got rid of it. 

But it works for now, and there are other more important things to do. )

ahankinson:
The think about Expires headers is that your clients need to support client-side caching. I imagine that most of the clients are scripts, and they generally don’t cache responses.

The only clients that really benefit are browsers, so I was more asking whether they would make up any of the traffic to this data.

Of course, having it doesn’t break anything, and it’s good for other types of resources so it’s good if you add it to the response for them.

config/nginx/conf.d/cantusdb.conf
```diff
@@ -46,7 +46,12 @@ server {
...
+        alias /resources/api_cache/concordances.json;
+        expires 6h;
```
dchiller: What is the logic behind this choice of expiration time? What about `modified 24h` given that theoretically there should be a new one every 24 hours

jacobdgm: If we had chosen a specific time of day, I was worried about our cache being updated just before or just after that time, and a client caching an out-of-date response at just the wrong time - 6h felt like a good balance between it being a decent chunk of time while excluding the possibility of a client's cache becoming stale by significantly more than 24h.

But I think `expires modified +24h;` is much better. Thanks for unearthing this!

jacobdgm: response headers now look like:
```
HTTP/1.1 200 OK
Server: nginx/1.24.0
Date: Tue, 30 Jan 2024 21:00:10 GMT
Content-Type: application/octet-stream
Content-Length: 192128923
Last-Modified: Tue, 30 Jan 2024 20:46:59 GMT
Connection: keep-alive
ETag: "65b96043-b73a79b"
Expires: Wed, 31 Jan 2024 20:46:59 GMT
Cache-Control: max-age=85609
Accept-Ranges: bytes
```

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: I wasn't fully absorbing what you meant by "by command-line parameter"... Just pushed a commit that allows you to run `python manage.py update_cached_concordances -d (or --directory) "/a/filepath/of/my/choosing"`. Thanks for this suggestion!

dchiller (approved):

ahankinson:
FYI: https://docs.python.org/3/library/argparse.html#default

This allows you to assume a value if none is supplied, which would also tidy up the `if` check you currently have.

jacobdgm:
> FYI: https://docs.python.org/3/library/argparse.html#default
> 
> This allows you to assume a value if none is supplied, which would also tidy up the `if` check you currently have.

This is a good suggestion - thank you! I just opened #1285.

PR #1286, opened 2024-02-02
Create duplicate configuration files for Staging and Development environments

jacobdgm:
This PR creates modified duplicates of various configuration files (i.e., `config/nginx/conf.d/cantusdb.conf`, `cron/cron.txt` and `docker-compose.yml`) for use in Staging and local development environments.

The following files have been created:

- For Staging:
  - `config/nginx/conf.d/cantusdb-staging.conf`, a modified duplicate of `config/nginx/conf.d/cantusdb.conf`
  - `cron/cron-staging.txt`, cf. `cron/cron.txt`
  - (`docker-compose.yml` requires no changes)
- For local development:
  - `config/nginx/conf.d/cantusdb-development.conf`, cf. `config/nginx/conf.d/cantusdb.conf`
  - `docker-compose-development.yml`, cf. `compose-development.yml`
  - (`cron/cron.txt` needs not be run in local development, and so a modified duplicate has not been provided)

Once this PR has been merged, during deployment, setting up configuration will include the following steps:
- On production:
  - in `cron/cron.txt`, replace the dummy email address in the certificate renewal job, and add the jobs to the running crontab
- On Staging:
  - cat the contents of `config/nginx/conf.d/cantusdb-staging.conf` to `config/nginx/conf.d/cantusdb.conf`
  - cat the contents of `cron/cron-staging.txt` to `cron/cron.txt`
  - in `cron/cron.txt`, replace the email address in the certificate renewal job, and add the jobs to the running crontab
- In local development
  - cat the contents of `config/nginx/conf.d/cantusdb-development.conf`, to `config/nginx/conf.d/cantusdb.conf`
  - cat the contents of `docker-compose-development.yml` to `docker-compose.yml`


docker-compose-development.yml
```diff
@@ -0,0 +1,51 @@
...
+      context: ./nginx
+    ports:
+      - 80:80
+      # - 443:443
+    volumes:
+      - ./config/nginx/conf.d:/etc/nginx/conf.d
+      - static_volume:/resources/static
+      - media_volume:/resources/media
+      - api_cache_volume:/resources/api_cache
+      # - ./certificates:/etc/nginx/ssl/live
```
dchiller: I think this could stay.

docker-compose-development.yml
```diff
@@ -0,0 +1,51 @@
...
+    restart: always
+    depends_on:
+      - postgres
+
+  nginx:
+    build:
+      context: ./nginx
+    ports:
+      - 80:80
+      # - 443:443
```
dchiller: Perhaps this could be an environment variable.

cron/cron-staging.txt
```diff
@@ -0,0 +1,12 @@
...
+
+# min   hour    day     month   weekday command
+0       4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/postgres/db_backup.sh
+10      4       *       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh update_cached_concordances
+40      4       1       *       *       bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_next_chant_fields; bash /home/ubuntu/code/CantusDB/cron/management/manage.sh populate_is_last_chant_in_feast
+# 50      4       *       *       7       /usr/local/bin/docker-compose -f /home/ubuntu/code/CantusDB/docker-compose.yml exec -T nginx lego --path /etc/nginx/ssl/live -d cantusdatabase.org -d www.cantusdatabase.org -d mass.cantusdatabase.org -m updateme@example.com --http --http.webroot /var/www/lego/ renew --days 45 --renew-hook "nginx -s reload" 
+50      4       *       *       7       /usr/local/bin/docker-compose -f /home/ubuntu/code/CantusDB/docker-compose.yml exec -T nginx lego --path /etc/nginx/ssl/live -d staging.cantusdatabase.org -d staging-alias.cantusdatabase.org -m updateme@example.com --http --http.webroot /var/www/lego/ renew --days 45 --renew-hook "nginx -s reload"
```
dchiller, line 12: Could this run a single command that depends on a variable? (eg. If staging, then x; if production, then y)

config/nginx/conf.d/cantusdb-development.conf
```diff
@@ -0,0 +1,80 @@
...
+# 
+#     ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
+#     ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
+# 
+#     location / {
+#         return 301 https://cantusdatabase.org$request_uri;
+#     }
+# }
+ 
+server {
```
dchiller: Looking at this now I'm realizing we don't have a `server_name` in this block?

dchiller: I noticed the other day that we aren't doing host checking at the Django level either, but I thought it was ok because we were doing it here, but it looks like we aren't. We should do both: https://docs.djangoproject.com/en/5.0/topics/security/#host-headers-virtual-hosting

jacobdgm:
I discovered that having multiple files with `.conf` extensions in `/config/nginx/conf.d/` causes Nginx to try to read all of them. Renaming `cantusdb-staging.conf` to `cantusdb.conf.staging`, and likewise for development, fixes the problem. I also made a similar substitution for our `docker-compose.yml` file.

jacobdgm:
@lucasmarchd01 noted that, once we've made local modifications to `cantusdb.conf` and to `docker-compose.yml`, it would be nice if those files could be ignored by git.

I think we could accomplish this by adding those two files to our `.gitignore`, and renaming the current files to `cantusdb.conf.production` and `docker-compose.yml.production`. This would mean an extra couple of steps while deploying to production, but would make the Staging and Production processes more similar. Thoughts on implementing this?

config/nginx/conf.d/cantusdb-development.conf
jacobdgm: If I understand, this is less relevant to of this file in particular, and more to the project in general - correct?

I think this is beyond the scope of this PR. Let me open a couple of new issues. Thanks for catching this!

cron/cron-staging.txt
jacobdgm: i.e., we create a new file, `update-certificates.sh` say, which takes an `--environment` argument say, with possible values `staging` and `production`, which would be called here?

docker-compose-development.yml
jacobdgm: Tried uncommenting this, and things still run fine. Good!

docker-compose-development.yml
jacobdgm: Are you suggesting that we include this line (or don't include it) based on an environment variable, or are you saying we can include a list of port mappings in an environment variable?

I've taken a look at CU's `docker-compose.yaml`, and done some cursory googling, and can't see how we can include/exclude a line based on a boolean environment variable. Do you have something specific in mind?

jacobdgm: A good argument for figuring out how to do this via environment variable is that it would allow us to have just a single `docker-compose.yml` file, yes?

dchiller: I'm saying would it work to have a variable in `dev_env` that is `PORT` and then have this be:

```yaml
ports:
     - ${PORT}:${PORT}
```

so that this can be set appropriately in the dev_env to whatever environment you're in (`443` on a server, `80` for development)

dchiller: > A good argument for figuring out how to do this via environment variable is that it would allow us to have just a single `docker-compose.yml` file, yes?

Yes

cron/cron-staging.txt
dchiller: Yes, either with a keyword or required argument. This could also be done with the email.

config/nginx/conf.d/cantusdb-development.conf
dchiller: I guess it depends if we want to validate the Host with nginx at all. If so, then that will be different for staging and production. I'm not yet seeing clear answers on the google.

jacobdgm: Just opened #1288 and #1289. I'm not sure I fully understand the ramifications of this, so feel free to weigh in on these issues with more information and any corrections as necessary.

docker-compose-development.yml
jacobdgm: > I'm saying would it work to have a variable in `dev_env` that is `PORT` and then have this be:
> 
> ```yaml
> ports:
>      - ${PORT}:${PORT}
> ```
> 
> so that this can be set appropriately in the dev_env to whatever environment you're in (`443` on a server, `80` for development)

Ah, I understand! And this would be specified in a new line in our `dev_env` file, correct?

jacobdgm: I've tried a few different ways to specify this environment variable in `dev_env`, but I keep getting an error when I try to `docker-compose build`:

```
$ docker-compose build
WARN[0000] The "PORT" variable is not set. Defaulting to a blank string. 
WARN[0000] The "PORT" variable is not set. Defaulting to a blank string. 
1 error(s) decoding:

* error decoding 'Ports': No port specified: :<empty>
```

(concluding that the variables in `dev_env` only show up once the container is built. So we would need to set an environment variable _outside_ the container before trying to build...?)

dchiller: Env variables can be passed from a file to Docker Compose using the `env_file` parameter: https://docs.docker.com/compose/environment-variables/set-environment-variables/

jacobdgm: yes - our docker-compose.yml already includes this line in the django container, and I added it to our nginx container too while testing this. Let me try again...

jacobdgm: looking at your message again and the documentation, the env_file is a parameter to the `docker-compose build` command, and not to be included in `docker-compose.yml` itself. Got it.

jacobdgm: I wonder whether this would actually make the configuration process simpler.

(Perhaps this is standard procedure and I've just not become accustomed to it having spent a bunch of time going through CantusDB's current deployment workflow as my only interaction with docker-compose...)

ahankinson:
Why not template these files with Ansible for the different envs?

dchiller:
> Why not template these files with Ansible for the different envs?

So include the relevant files in the ansible repo rather than here?

docker-compose-development.yml
dchiller: > yes - our docker-compose.yml already includes this line in the django container, and I added it to our nginx container too while testing this. Let me try again...

Did this not work?

ahankinson:
Yes, that’s generally how I do it. You can include some examples, but you generally include your config in Ansible then template the fields into place depending on the environment. It means there is one place for your entire app configuration.

cron/cron-staging.txt
ahankinson: Seriously, use Ansible. 

Deploy the staging cron file to staging. Deploy the production cron file to production. Template the paths/domains you need.

docker-compose.yml.development
```diff
@@ -0,0 +1,51 @@
...
+    restart: always
+    depends_on:
+      - postgres
+
+  nginx:
+    build:
+      context: ./nginx
+    ports:
+      - 80:80
+      # - 443:443
```
dchiller: Now that I think about it, I'm not even sure why we need to change this for development at all?

docker-compose.yml.development
```diff
@@ -0,0 +1,51 @@
...
+    build:
+      context: ./django
+    volumes:
+      - ./:/code/
+      - static_volume:/resources/static
+      - media_volume:/resources/media
+      - api_cache_volume:/resources/api_cache
+    env_file: ./config/envs/dev_env
+    ports:
+      - 3000:3000
```
dchiller: I was curious too why this port mapping was here and if my code working was successful, it's related to VSCode....in which case it should only be for development, I think

dchiller (commented):
Ok, I think besides my most recent port-related questions, I am good to approve this. And then I think it can be ansible-ized for the new servers.

docker-compose-development.yml
jacobdgm: Correct - including `env_file: ./config/envs/dev_env` in the `nginx` block did not resolve the error that `- ${PORT}:${PORT}` was causing.

As far as I can tell, specifying an `env_file` means that the variables defined in that file will be available inside the container when it is running (allowing django, for example, to [get the value of `CANTUSDB_DEBUG` in `settings.py`](https://github.com/DDMAL/CantusDB/blob/13e91741d06999a80879c745a8e75319460d7fdd/django/cantusdb_project/cantusdb/settings.py#L39)). But it doesn't make the variables available to docker when it's evaluating `docker-compose.yml` itself.

Doing some digging, it appears that we can specify all the necessary environment variables in the command every time we call `docker-compose` (i.e., `docker-compose up -d` becomes `PORT=80 docker-compose up -d`). But if we do this, I'm worried about it cascading into a bunch of changes in several other files, especially those relating to our cron setup.

(reference: https://serversforhackers.com/c/div-variables-in-docker-compose . Part of this page suggests that maybe the error I'm hitting could be resolved by putting `dev_env` directly in the root of the repository, next to `docker-compose.yml`?)

dchiller: > Part of this page suggests that maybe the error I'm hitting could be resolved by putting `dev_env` directly in the root of the repository

This is what CU does and it works, but it's annoying that you can't specify another path... I guess it was worth a try? But I agree that at this point it's more trouble than it's worth (and adding things to the command, I agree, seems like it would be potentially troublesome)

jacobdgm:
Okay.

I'm interested in learning Ansible, but I've been resisting diving into learning it because 1) I didn't want to add another tool/dependency to the project when the possibility exists to just tweak what we currently have, and 2) It wasn't clear to me that time spent learning to use Ansible and implementing it here would result in a net savings of developer time in the long run.

Based on your feedback, @ahankinson, I'm getting the sense that learning and using Ansible is really worth doing here. I'm going to suggest that we make a couple of last tweaks to this PR and then merge it. I've opened #1296 - I'll spend a little time every day learning about Ansible in general and exploring how DDMAL currently uses it in our other repos, and hopefully open a PR that addresses #1296 sometime in the next few weeks.

docker-compose.yml.development
jacobdgm: I'll need to check, but I'm pretty sure that if we keep `443:443` here, Nginx will go looking for certificates, fail to find them, and error out.

docker-compose.yml.development
jacobdgm: good catch! I'll make this change.

dchiller:
@jacobdgm @lucasmarchd01 I just added you to the recently-created DACT CantusDB ansible repo.

docker-compose.yml.development
dchiller: I think nginx will only look for it if there is `ssl` in the `listen` line...but yes, good to check

jacobdgm:
I was getting tired of making changes to my `docker-compose.yml` to make it run locally, accidentally committing those changes, and then having to fix my mistakes in follow-on commits. So I went ahead and created a `docker-compose.yml.production` and a `cantusdb.conf.production`, and added `docker-compose.yml` and `cantusdb.conf.production` to `.gitignore`.

docker-compose.yml.development
jacobdgm: it works with 443 present, and I changed this back (i.e. uncommented in the file).

dchiller (approved):

ahankinson:
If you have more than one server (production and staging) then Ansible is absolutely necessary. It does much more than Docker, and in much more understandable ways. If you have it set up you can point it to a completely bare bones server and have your application up and running in seconds. Under the hood all it’s doing is running shell commands over ssh, but it makes you spell out all the steps needed to get your application running, including things like installing docker.

I like it because there is no real magic. To me it obviates the need for something like docker since I can simply point it at a fresh machine and have a new install of my app in a few minutes, without needing an extra layer of containers and overlays and images blah blah blah. You get fewer layers of abstraction with the same ease of deployment. And you have all your steps written out in the roles so if you forget what you’ve done, you can just refer to the Ansible repo to see.

I’ve given you and Lucas access to my RISM Online Ansible setup so you can see how it works. With this I can (and have) deploy a new RISM Online stack from scratch.

PR #1293, opened 2024-02-02
Change how postgres finds postgres.conf, which contains modified configurations

### LINKED ISSUE #1283: Postgres configurations are not active
On the servers, values of the postgres configs are different than what's in the postgres conf... (Showing production here)

image

image

`/var/lib/postgresql/data/postgres.conf` looks nothing like what's committed to the repo.

I wonder whether these databases were ever updated with the new configs//these looks like they are being overwritten every time?

dchiller:
Previously, the postgres configuration file was added to the /var/lib/postgresql/data directory which was overwritten by the already-existing named data volume (`postgres_data`), causing the modified configurations not to be picked up. When initializing a database, the placement of the configuration file in /var/lib/postgresql/data caused an error that the "PGDATA was not empty."

This commit fixes this issue by placing the configuration file in a different location (/etc/postgresql) and then running the Dockerfile entrypoint with the path to the config file.

Closes #1283

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,7 +39,7 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
     restart: always
-    command: -c 'config_file=/etc/postgresql/postgresql.conf'
+    command: postgres -c 'config_file=/etc/postgresql/postgresql.conf'
 
 volumes:
   postgres_data:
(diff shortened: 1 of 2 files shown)


jacobdgm (approved):
This looks reasonable. Thanks for looking into this!

jacobdgm:
Will this require running any commands on the staging/production servers when this change makes its way there, or do we just need to docker-compose build, down and up?

dchiller:
> Will this require running any commands on the staging/production servers when this change makes its way there, or do we just need to docker-compose build, down and up?

Just build and restart containers, yes. We should definitely verify when we deploy to both, but that was all that was required on my local version.

PR #1299, opened 2024-02-06
Make separate production and staging versions of cron.txt

### LINKED ISSUE #1269: It would be a good idea to save separate configuration files for staging vs. production
This will simplify the deployment of changes to staging in particular.

Plan of action: for every file in https://github.com/DDMAL/CantusDB/wiki/Getting-Started-with-Development#docker-set-up (i.e., `docker-compose.yml` and `config/nginx/conf.d/cantusdb.conf`)
I should also review `cron.txt`, but I'm pretty sure this won't need any changes now that #1273 is complete.

Our deployment instructions can then be updated to be things like "run `cat docker-compose.yml.staging > docker-compose.yml`" rather than a series of "on the line that begins {thus}, replace {this} with {that}"s

jacobdgm:
This PR:
- renames `cron-staging.txt` -> `cron.txt.staging`, and renames `cron.txt` -> `cron.txt.production`, thus bringing our cron configuration into alignment with our docker (in `docker-compose.yml`) and nginx (in `conf.d/cantusdb.conf`) configuration.
- Adds `cron/cron.txt` to `.gitignore`

This PR fully resolves #1269.


dchiller (approved):
Looks good. We can reduce a deployment step...these don't need to be `cat`-ed anywhere before they are added to crontab, but ok as is.

PR #1300, opened 2024-02-06
Add postgres command with path to config file to docker-compose staging

dchiller:
Simultaneous work on #1286 and #1293 meant the staging and production versions of `docker-compose.yml` had different commands. This fixes the command in the staging version with an argument pointing to the location of the `postgres.conf` file in the `postgres` container.

### DIFF
--- a/docker-compose.yml.staging
+++ b/docker-compose.yml.staging
@@ -37,6 +37,7 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
     restart: always
+    command: postgres -c 'config_file=/etc/postgresql/postgresql.conf'
 
 volumes:
   postgres_data:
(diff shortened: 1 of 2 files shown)


docker-compose.yml.staging
```diff
@@ -37,6 +37,7 @@ services:
     volumes:
       - postgres_data:/var/lib/postgresql/data/
     restart: always
+    command: postgres -c 'config_file=/etc/postgresql/postgresql.conf'
```
jacobdgm, line 40: Please add this line to `docker-compose.yml.development` too!

jacobdgm (changes requested):

docker-compose.yml.staging
dchiller: do we want that? Weren't these determined for the servers only?

jacobdgm: I think we want the configuration to be as similar as possible between develop, staging and production, don't we?

jacobdgm: I can see where you're coming from, though.

jacobdgm (approved):

PR #1308, opened 2024-02-08
Ensure CSRF_TRUSTED_ORIGINS and ALLOWED_HOSTS are specified

### LINKED ISSUE #1288: We should make sure our `ALLOWED_HOSTS` is complete
As noted by @dchiller in https://github.com/DDMAL/CantusDB/pull/1286#discussion_r1476453258

> If I understand correctly, our staging/production sites are configured to receive traffic from `www.cantusdatabase.org`, `mass.cantusdatabase.org`, `staging.cantusdatabase.org`, etc., and we open ourself up to security issues if these hosts are not all included in `ALLOWED_HOSTS`.

jacobdgm:
This PR specifies a `CSRF_TRUSTED_ORIGINS` and `ALLOWED_HOSTS` in `settings.py`, and thus fixes #1288.

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -200,7 +200,21 @@ INTERNAL_IPS = [
     "127.0.0.1",
 ]
 
-CSRF_TRUSTED_ORIGINS = ["https://cantusdatabase.org", "https://www.cantusdatabase.org"]
+CSRF_TRUSTED_ORIGINS = [
+    "https://cantusdatabase.org",
+    "https://www.cantusdatabase.org",
+    "https://staging.cantusdatabase.org",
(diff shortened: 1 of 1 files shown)


dchiller:
I think all of these should depend on the deployment environment. So `staging.cantusdatabase.org` should not be allowed on production, etc.

jacobdgm:
ah, and I see on line 42 that we use `os.getenv` to set ALLOWED_HOSTS. If we merged this change, it would overwrite that. Let me see what I can do to properly address this.

jacobdgm:
I made some changes. Please also refer to the "CantusDB Resources" page on the internal wiki to see the accompanying changes to the `dev_env` file.

dchiller (approved):

PR #1315, opened 2024-02-15
Specify server names in nginx configuration

### LINKED ISSUE #1289: We should ensure our `conf.d/cantusdb.conf.*` files include all relevant `server_name`s
As noted by @dchiller in https://github.com/DDMAL/CantusDB/pull/1286#discussion_r1476453258

jacobdgm:
This PR specifies all server names in our staging and production nginx configurations. It fixes #1289.

This is a change I can't really test locally, and different changes have been applied to the production configuration than to the staging configuration, so some real scrutiny of these changes (especially to `cantusdb.conf.production`) would be appreciated.

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf.production
+++ b/config/nginx/conf.d/cantusdb.conf.production
@@ -1,5 +1,8 @@
 server {
+    # Redirect all http traffic to corresponding https page
     listen 80;
+
+    server_name cantusdatabase.org www.cantusdatabase.org mass.cantusdatabase.org;
 
     server_tokens off;
 
@@ -27,8 +30,9 @@ server {
 }
 
 server {
-
     listen 443 default_server http2 ssl;
+
+    server_name cantusdatabase.org www.cantusdatabase.org;
 	
     ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
     ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
--- a/config/nginx/conf.d/cantusdb.conf.production
+++ b/config/nginx/conf.d/cantusdb.conf.production
@@ -16,10 +16,11 @@ server {
 }
 
 server {
-    # Redirect all https traffic for mass.cantusdatabase.org to cantusdatabase.org
+    # Redirect all https traffic for www.cantusdatabase.org and mass.cantusdatabase.org
+    # to cantusdatabase.org
     listen 443 ssl;
 
-    server_name mass.cantusdatabase.org;
+    server_name www.cantusdatabase.org mass.cantusdatabase.org;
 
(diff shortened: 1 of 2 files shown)


dchiller (approved):

config/nginx/conf.d/cantusdb.conf.production
```diff
@@ -27,8 +30,9 @@ server {
 }
 
 server {
-
     listen 443 default_server http2 ssl;
+
+    server_name cantusdatabase.org www.cantusdatabase.org;
```
ahankinson: As I mentioned to @dchiller, you don't really want to do this, since it will consider `cantusdatabase.org` and `www.cantusdatabase.org` to both be valid addresses. You should set up a different server and redirect any traffic coming from `www.cantusdatabase.org` to the canonical `cantusdatabase.org` instead.

dchiller: You did say this...and it is so in the ansible set-up...my bad for not catching here!

dchiller (approved):

config/nginx/conf.d/cantusdb.conf.production
```diff
@@ -1,6 +1,9 @@
 server {
+    # Redirect all http traffic to corresponding https page
```
ahankinson, line 2: Returning  301 $host will return a redirect to the https version of the hostnames, which will then do a 301 to the canonical. You should just do cantusdatabase.org instead.

dchiller: In other words, we want to listen on 80 for all allowed hostnames and redirect all to `cantusdatabase.org` on 443? (Also listening as we do on 443 for those hostnames as we do here in the other block)

ahankinson: Just change

`return 301 https://$host$request_uri;`

to

`return 301 https://cantusdatabase.org$request_uri;`

dchiller: great, thanks!

config/nginx/conf.d/cantusdb.conf.staging
```diff
@@ -1,6 +1,9 @@
 server {
+    # redirect all http traffic to corresponding https page
     listen 80;
 
+    server_name staging.cantusdatabase.org staging-alias.cantusdatabase.org;
```
ahankinson, line 5: I think you also needed to change it on staging...

config/nginx/conf.d/cantusdb.conf.production
jacobdgm: I can see why it was initially set up as it is - the first redirect is like a function that does just one thing (http -> https) and the second does just one other thing (variant subdomain -> canonical). But it does create extra network traffic, so I see the value in setting it up as you suggested, @ahankinson. Thanks for this suggestion - I just opened #1322 implementing it.

PR #1318, opened 2024-02-16
Create hook in signals.py to automatically update chants' and sequences' incipits

### LINKED ISSUE #1187: We should create a hook that updates a chant/sequence's `incipit` field when the chant/sequence is saved
We don't keep chants' `incipit`s up-to-date, yet `ajax_search_bar` searches the `incipit` field. So, as users add chants, they will not appear in search results while the user is typing in the search bar.

In general, we need to do something with incipits: either update them based on some rule in `signals.py`, or calculate them on-the-fly. See #361, and perhaps one or two other issues. Recent times we've met with her, Debra has mentioned wanting to have a consistent way to handle incipits.

jacobdgm:
This PR adds several functions in `signals.py` that automatically keep chants' and sequences' incipits up-to-date. In doing so, it will resolve #1187.

While working in `signals.py`, return annotations were added to all functions. Also, several functions that were declared within other functions have be un-nested, which should make the code simpler/clearer to read.

In the admin area, it sets the `incipit` field for Chants and Sequences to be read-only.

It also updates a great many tests in the test suite, as we had been using incipits in many places to, e.g., test that the expected chant was being returned among a set of results.

### DIFF
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -21,6 +21,8 @@ EXCLUDE = ("json_info",)
 READ_ONLY = (
     "created_by",
     "last_updated_by",
+    "date_created",
+    "date_updated",
 )
 
 
@@ -61,11 +63,7 @@ class ChantAdmin(BaseModelAdmin):
         "id",
     )
 
-    readonly_fields = READ_ONLY + (
-        "date_created",
-        "date_updated",
-        "incipit",
-    )
+    readonly_fields = READ_ONLY + ("incipit",)
 
     list_filter = (
         "genre",
--- a/django/cantusdb_project/main_app/admin.py
+++ b/django/cantusdb_project/main_app/admin.py
@@ -64,6 +64,7 @@ class ChantAdmin(BaseModelAdmin):
     readonly_fields = READ_ONLY + (
         "date_created",
         "date_updated",
+        "incipit",
     )
 
     list_filter = (
@@ -170,6 +171,7 @@ class SequenceAdmin(BaseModelAdmin):
         "source",
         "feast",
     )
+    readonly_fields = READ_ONLY + ("incipit",)
     ordering = ("source__siglum",)
     form = AdminSequenceForm
 
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -121,7 +121,7 @@ def update_volpiano_fields(instance) -> None:
     )
 
 
-def generate_volpiano_notes(volpiano) -> None:
+def generate_volpiano_notes(volpiano) -> str:
     """
     Populate the ``volpiano_notes`` field of the ``Chant`` model
 
@@ -136,7 +136,7 @@ def generate_volpiano_notes(volpiano) -> None:
     # unwanted_chars are non-note chars, including the clefs, barlines, and accidentals etc.
     # the `searchMelody.js` on old cantus makes no reference to the b-flat accidentals ("y", "i", "z")
     # so put them in unwanted chars for now
-    unwanted_chars = [
+    unwanted_chars: list[str] = [
         "-",
         "1",
         "2",
@@ -153,9 +153,9 @@ def generate_volpiano_notes(volpiano) -> None:
         "z",
     ]
     # convert all charactors to lower-case, upper-case letters stand for liquescent of the same pitch
-    volpiano_lower = volpiano.lower()
+    volpiano_lower: str = volpiano.lower()
     # `)` stands for the lowest `g` note liquescent in volpiano, its 'lower case' is `9`
-    volpiano_notes = volpiano_lower.replace(")", "9")
+    volpiano_notes: str = volpiano_lower.replace(")", "9")
     # remove none-note charactors
     for unwanted_char in unwanted_chars:
         volpiano_notes = volpiano_notes.replace(unwanted_char, "")
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -164,7 +164,7 @@ def generate_volpiano_notes(volpiano) -> str:
     return volpiano_notes
 
 
-def generate_volpiano_intervals(volpiano_notes) -> None:
+def generate_volpiano_intervals(volpiano_notes) -> str:
     """
     Populate the ``volpiano_intervals`` field of the ``Chant`` model
 
@@ -178,24 +178,23 @@ def generate_volpiano_intervals(volpiano_notes) -> None:
         str: A str of digits, recording the intervals between adjacent notes
     """
     # replace '9' (the note G) with the char corresponding to (ASCII(a) - 1), because 'a' denotes the note A
-    volpiano_notes = volpiano_notes.replace("9", chr(ord("a") - 1))
+    volpiano_notes: str = volpiano_notes.replace("9", chr(ord("a") - 1))
     # we model the interval between notes using the difference between the ASCII codes of corresponding letters
     # the letter for the note B is "j" (106), note A is "h" (104), the letter "i" (105) is skipped
     # move all notes above A down by one letter
-    volpiano_notes = list(volpiano_notes)
-    for j, note in enumerate(volpiano_notes):
+    notes_list: list = list(volpiano_notes)
+    for j, note in enumerate(notes_list):
         if ord(note) >= 106:
-            volpiano_notes[j] = chr(ord(note) - 1)
+            notes_list[j] = chr(ord(note) - 1)
 
     # `intervals` records the difference between two adjacent notes.
     # Note that intervals are encoded by counting the number of scale
     # steps between adjacent notes: an ascending second is thus encoded
     # as "1"; a descending third is encoded "-2", and so on.
-    intervals = []
+    intervals: list[int] = []
     for j in range(1, len(volpiano_notes)):
         intervals.append(ord(volpiano_notes[j]) - ord(volpiano_notes[j - 1]))
-    # convert `intervals` to str
-    volpiano_intervals = "".join([str(interval) for interval in intervals])
+    volpiano_intervals: str = "".join([str(interval) for interval in intervals])
     return volpiano_intervals
 
 
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -18,7 +18,7 @@ from main_app.models import Source
 
 
 @receiver(post_save, sender=Chant)
-def on_chant_save(instance, **kwargs):
+def on_chant_save(instance, **kwargs) -> None:
     update_source_chant_count(instance)
     update_source_melody_count(instance)
 
@@ -28,28 +28,28 @@ def on_chant_save(instance, **kwargs):
 
 
 @receiver(post_delete, sender=Chant)
-def on_chant_delete(instance, **kwargs):
+def on_chant_delete(instance, **kwargs) -> None:
     update_source_chant_count(instance)
     update_source_melody_count(instance)
 
 
 @receiver(post_save, sender=Sequence)
-def on_sequence_save(instance, **kwargs):
+def on_sequence_save(instance, **kwargs) -> None:
     update_source_chant_count(instance)
     update_sequence_incipit_field(instance)
 
 
 @receiver(post_delete, sender=Sequence)
-def on_sequence_delete(instance, **kwargs):
+def on_sequence_delete(instance, **kwargs) -> None:
     update_source_chant_count(instance)
 
 
 @receiver(post_save, sender=Feast)
-def on_feast_save(instance, **kwargs):
+def on_feast_save(instance, **kwargs) -> None:
     update_prefix_field(instance)
 
 
-def update_chant_search_vector(instance):
+def update_chant_search_vector(instance) -> None:
     """When saving an instance of Chant, update its search vector field.
 
     Called in on_chant_save()
@@ -67,7 +67,7 @@ def update_chant_search_vector(instance):
     )
 
 
-def update_source_chant_count(instance):
+def update_source_chant_count(instance) -> None:
     """When saving or deleting a Chant or Sequence, update its Source's number_of_chants field
 
     Called in on_chant_save(), on_chant_delete(), on_sequence_save() and on_sequence_delete()
@@ -83,7 +83,7 @@ def update_source_chant_count(instance):
         source.save()
 
 
-def update_source_melody_count(instance):
+def update_source_melody_count(instance) -> None:
     """When saving or deleting a Chant, update its Source's number_of_melodies field
 
     Called in on_chant_save() and on_chant_delete()
@@ -103,13 +103,13 @@ def update_source_melody_count(instance):
         source.save()
 
 
-def update_volpiano_fields(instance):
+def update_volpiano_fields(instance) -> None:
     """When saving a Chant, make sure the chant's volpiano_notes and volpiano_intervals are up-to-date
 
     Called in on_chant_save()
     """
 
-    def generate_volpiano_notes(volpiano):
+    def generate_volpiano_notes(volpiano) -> None:
         """
         Populate the ``volpiano_notes`` field of the ``Chant`` model
 
@@ -151,7 +151,7 @@ def update_volpiano_fields(instance):
         volpiano_notes = re.sub(r"(.)\1+", r"\1", volpiano_notes)
         return volpiano_notes
 
-    def generate_volpiano_intervals(volpiano_notes):
+    def generate_volpiano_intervals(volpiano_notes) -> None:
         """
         Populate the ``volpiano_intervals`` field of the ``Chant`` model
 
@@ -197,7 +197,7 @@ def update_volpiano_fields(instance):
     )
 
 
-def update_prefix_field(instance):
+def update_prefix_field(instance) -> None:
     pk = instance.pk
 
     if instance.feast_code:
@@ -207,7 +207,7 @@ def update_prefix_field(instance):
         instance.__class__.objects.filter(pk=pk).update(prefix="")
 
 
-def update_chant_incipit_field(chant: Chant):
+def update_chant_incipit_field(chant: Chant) -> None:
     """Update the incipit field of the specified Chant to be the first
     several words of the chant's standardized-spelling fulltext
 
@@ -223,7 +223,7 @@ def update_chant_incipit_field(chant: Chant):
         Chant.objects.filter(id=chant.id).update(incipit=new_incipit)
 
 
-def update_sequence_incipit_field(sequence: Sequence):
+def update_sequence_incipit_field(sequence: Sequence) -> None:
     """Update the incipit field of the specified Sequence to be the first
     several words of the sequence's standardized-spelling fulltext
 
--- a/django/cantusdb_project/main_app/signals.py
+++ b/django/cantusdb_project/main_app/signals.py
@@ -192,8 +192,8 @@ def generate_volpiano_intervals(volpiano_notes) -> str:
     # steps between adjacent notes: an ascending second is thus encoded
     # as "1"; a descending third is encoded "-2", and so on.
     intervals: list[int] = []
-    for j in range(1, len(volpiano_notes)):
-        intervals.append(ord(volpiano_notes[j]) - ord(volpiano_notes[j - 1]))
+    for j in range(1, len(notes_list)):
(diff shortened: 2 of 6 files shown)


jacobdgm:
(So it turns out we had been doing quite a lot of setting/reading test chants' incipits in our test suite. I'm not finished working through the necessary changes, but we're down from ~40 failures/errors to 12 errors. Hopeful I'll be able to polish them off in short order early next week!)

django/cantusdb_project/main_app/admin.py
```diff
@@ -170,6 +171,7 @@ def get_source_siglum(self, obj):
         "source",
         "feast",
     )
+    readonly_fields = READ_ONLY + ("incipit",)
```
dchiller, line 172: This isn't exactly related, but I am having trouble following the readonly_fields assignment in this file. 

- Why does a `sequence` and a `chant` have different `readonly_fields`? 
- Most of the Admin classes use the `readonly_fields` of the `BaseModelAdmin` but it looks like `date_created` and `date_updated` are not marked as `readonly_fields` there and yet they don't show up in the admin view? 

I'll continue to poke around...but let me know if I am missing something obvious...

django/cantusdb_project/main_app/signals.py
```diff
@@ -201,3 +205,64 @@ def update_prefix_field(instance):
         instance.__class__.objects.filter(pk=pk).update(prefix=prefix)
     else:  # feast_code is None, ""
         instance.__class__.objects.filter(pk=pk).update(prefix="")
+
+
+def update_chant_incipit_field(chant: Chant):
```
dchiller: Type annotation for return? Functions with no explicit `return` return `None` I believe

django/cantusdb_project/main_app/signals.py
```diff
@@ -21,6 +23,7 @@ def on_chant_save(instance, **kwargs):
     update_source_melody_count(instance)
 
     update_chant_search_vector(instance)
+    update_chant_incipit_field(instance)
```
dchiller, line 26: What happens if this fails? Do we need to worry about this?

I think in the case of `update_source_chant_count` this matters less (you'd be off by one until the next chant in the source is saved), but in a case like the incipit, that won't update again until the specific chant is saved again.

django/cantusdb_project/main_app/signals.py
```diff
@@ -201,3 +205,64 @@ def update_prefix_field(instance):
...
+
+    Args:
+        sequence (Sequence): The sequence from the database whose `incipit`
+        field is to be updated
+    """
+    fulltext: Optional[str] = ""
+    std_fulltext: Optional[str] = sequence.manuscript_full_text_std_spelling
+    ms_fulltext: Optional[str] = sequence.manuscript_full_text
+    title: Optional[str] = sequence.title
+    if title:  # most (all?) sequences in the database have no
```
dchiller: Just checked...

All sequences have empty `manuscript_full_text` and `manuscript_full_text_std_spelling` fields

django/cantusdb_project/main_app/signals.py
```diff
@@ -201,3 +205,64 @@ def update_prefix_field(instance):
...
+    Args:
+        chant (Chant): The chant from the database whose `incipit` field
+        is to be updated
+    """
+    fulltext: Optional[str] = chant.manuscript_full_text_std_spelling
+    if fulltext:  # many chants in the database have only an incipit -
+        # we should only update the incipit if the chant has a fulltext,
+        # just in case a chant manages to get saved without a fulltext somehow
+        new_incipit: str = generate_incipit(fulltext)
+        Chant.objects.filter(id=chant.id).update(incipit=new_incipit)
```
dchiller, line 224: So, if a chant doesn't have full text, how do I edit an incipit after this PR since the incipit field is now read-only in the Admin view and not a field in the Chant Edit form?

django/cantusdb_project/main_app/admin.py
jacobdgm: You're right, I don't think there's a good reason for these to have different values for `readonly_fields` for Sequences and Chants. Let me update `READ_ONLY` to include `date_created` and `date_updated`, to bring the two ModelAdmins (and all the others, for that matter) into alignment. Thanks for catching this!

django/cantusdb_project/main_app/signals.py
jacobdgm: thanks for catching this!

django/cantusdb_project/main_app/signals.py
jacobdgm: I don't know what Debra et al. have in mind for touching up the Bower database, but I don't find it hard to imagine a situation where someone adds fulltexts for some of our sequences. Or was this a comment on the comment itself?

I'm never quite sure how to handle comments in the code about the state of the live database at particular moments in time. Would something like `# as of late February 2024, all sequences in the database have no fulltext, etc.` be better here, you think? At the very least, it will remain accurate into the future...

django/cantusdb_project/main_app/signals.py
jacobdgm: I just tested on staging - in the admin area, you cannot save a chant without specifying a fulltext. It has never been possible in NewCantus to update a chant's incipit without suppling a fulltext, as far as I know.

django/cantusdb_project/main_app/signals.py
jacobdgm: I'm not sure I can envision a situation where this fails.

What does "fail" mean here? If a developer ends up creating a bug accidentally, wouldn't the user be shown a 500 page, and hopefully alert Debra to the fact that a chant they tried to save led to an error? Are you worried about the case where the hook somehow fails silently?

django/cantusdb_project/main_app/admin.py
dchiller: Do you know why, though, on current production in many places `date_created` and `date_updated` don't show up in the edit form? For example: https://cantusdatabase.org/admin/main_app/century/3861/change/

My current guess is that the `Meta` class and the customizations (in the Century form, the `name` field, for example) are not working well together. See 

https://github.com/DDMAL/CantusDB/blob/07be678c5e18934bf958d9f3aae728ec54e3e264/django/cantusdb_project/main_app/forms.py#L488-L493

dchiller: > Let me update `READ_ONLY` to include `date_created` and `date_updated`, to bring the two ModelAdmins (and all the others, for that matter) into alignment.

What you've done looks better, I just think there's something happening here that's unexpected that might need a fix.

jacobdgm: Got it. In principle, we could open an issue saying "there's a bug: some fields that we expect to be showing up in the admin area, like `date_created` and `date_updated`, are not actually showing up". This PR would fix the lack of `date_created`/`_updated`, but since we haven't figured out why they were missing in the first place, it could be that there are other fields that are missing that we're not aware of.

Am I understanding what you're getting at?

django/cantusdb_project/main_app/signals.py
dchiller: So, first there is the thing about the comment...we do know that no sequences have full text so I think we can say that here. 

Second, I don't know that it makes sense to use these other text fields for the incipit when we don't currently make use of those text fields. Yes, someday those may be used, but since we currently don't, it seems like it is more likely to cause issues (someone accidentally types something into one of these fields and it messes up the incipit) than not.

django/cantusdb_project/main_app/signals.py
dchiller: > Are you worried about the case where the hook somehow fails silently?

Yes, exactly. 

So, I just pulled your branch and ran a test locally. If the signal fails, I get an error, but the change to the full text still saves, so at least I got an error (I know something went wrong) but now things are out of sync.

My worry is not so much that it would be horrible for something to fail (it might and someone could always make an edit again) but if things get out of sync.

django/cantusdb_project/main_app/admin.py
jacobdgm: fwiw, with this PR as it currently stands, for all the models (other than Chant, Sequence and Source, which have tons of fields and I haven't checked them thoroughly), all the fields that we expect to appear (i.e., those defined in the model itself, plus the four defined in BaseModel) all appear as expected in the Admin area.

django/cantusdb_project/main_app/signals.py
dchiller: Primary question here is perhaps why use a signal and why not just do this on save?

Secondary question is maybe this is a place to work with django's db transaction stuff (although I don't know how that works with signals) (https://docs.djangoproject.com/en/5.0/topics/db/transactions/)

django/cantusdb_project/main_app/signals.py
jacobdgm: Would it be better to get rid of the various `if`s here and just set incipit to be equal to `title`?

django/cantusdb_project/main_app/signals.py
dchiller: Cool! I just looked and there are only four chants with empty manuscript_full_text_std_spelling in Cantus DB, so yes, not something that happens.

django/cantusdb_project/main_app/signals.py
jacobdgm: hm... I think I just put this in `signals.py` because similar functionality has gone here up until now, and it hadn't really occurred to me to try to put this in the model's `.save()` method.

Perhaps we merge this PR [nearly] as-is (I still want to update a couple of things, like how we handle sequences' incipits [this has now been updated]), and then open an issue to figure out whether overriding models' `.save()` methods would be more robust than the current approach with signals. As you noted, this affects other things like sources' `number_of_chants`, many of which are beyond the scope of this PR.

django/cantusdb_project/main_app/admin.py
dchiller: > fwiw, with this PR as it currently stands, for all the models (other than Chant, Sequence and Source, which have tons of fields and I haven't checked them thoroughly), all the fields that we expect to appear (i.e., those defined in the model itself, plus the four defined in BaseModel) all appear as expected in the Admin area.

Yes, it looks good. I guess I don't know why these fields weren't showing up before, but they seem to be with this PR (I also just looked through things locally), so I think we can resolve this.

django/cantusdb_project/main_app/signals.py
jacobdgm: I just pushed commits updating the comment, and always setting the sequence's incipit to its title

dchiller: I think that would be good. Or, as I read it again now, if you wanted to write this forward-looking functionality, maybe just change the `fulltext` variable name to something like `text_to_gen_incipit` or some such. I think setting `fulltext` multiple times to things that are not always "full texts" adds to the confusion.

jacobdgm: or `text_for_generating_incipit`, since characters are cheap (though I'm sure @JoyfulGen would appreciate being immortalized in the CantusDB codebase in this way :P)

Moot point now though - since the things in the `title` field are already incipits, I've simply gone with `incipit`.

django/cantusdb_project/main_app/tests/make_fakes.py
```diff
@@ -310,22 +303,20 @@ def make_fake_segment(name: str = None, id: int = None) -> Segment:
...
         source = make_fake_source(segment_name="Bower Sequence Database")
     if title is None:
         title = faker.sentence()
-    if incipit is None:
-        incipit = title[:INCIPIT_LENGTH]
     if cantus_id is None:
         cantus_id = make_random_string(6, "0123456789")
     sequence = Sequence.objects.create(
         title=title,
+        # most sequences in the database have no standard-spelling fulltext,
```
dchiller: They do? I don't think this materially impacts anything, but I find none that have it...

django/cantusdb_project/main_app/signals.py
annamorphism: > Cool! I just looked and there are only four chants with empty manuscript_full_text_std_spelling in Cantus DB, so yes, not something that happens.

wait, really? I thought there were quite a few chants without a full text (but with an incipit). But the scenario you propose is basically moot: if you want to update an incipit, it's presumably because you know what the text should be, which means you are in a position to give it a full text (or at least to put something in that field), which is desirable anyway. So requiring a full text to change the incipit seems fine...

dchiller: Whelp! looks like I fell into the character-based field null trap (https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.null) 

There are many chants with null `manuscript_full_text_std_spelling`, but only four with empty string.

So...my bad....nvm.

django/cantusdb_project/main_app/tests/make_fakes.py
jacobdgm: You're right. I made an assumption; thanks for actually verifying this.

django/cantusdb_project/main_app/signals.py
```diff
@@ -193,11 +121,139 @@ def generate_volpiano_intervals(volpiano_notes):
     )
 
 
-def update_prefix_field(instance):
+def generate_volpiano_notes(volpiano) -> None:
```
dchiller: Some discrepancies between type annotations, docstring, and return statement...

django/cantusdb_project/main_app/signals.py
```diff
@@ -193,11 +121,139 @@ def generate_volpiano_intervals(volpiano_notes):
...
+    volpiano_notes = volpiano_lower.replace(")", "9")
+    # remove none-note charactors
+    for unwanted_char in unwanted_chars:
+        volpiano_notes = volpiano_notes.replace(unwanted_char, "")
+    # remove duplicate consecutive chars
+    volpiano_notes = re.sub(r"(.)\1+", r"\1", volpiano_notes)
+    return volpiano_notes
+
+
+def generate_volpiano_intervals(volpiano_notes) -> None:
```
dchiller: Same here...

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2370,6 +2488,9 @@ def test_column_header_links(self):
             feast=feast,
             position=position,
         )
+        # incipit is automatically calculated from fulltext
+        # on chant save, so we need to set the incipit this way.
+        Chant.objects.filter(id=chant.id).update(incipit=incipit)
```
dchiller: I am confused by what's going on here. The comment is correct, but it seems to me that then your test shouldn't work around it (by bypassing the signal just implemented)?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2122,8 +2219,16 @@ def test_order_by_genre(self):
...
         source = make_fake_source()
-        chant_1 = make_fake_chant(incipit="isaac", cantus_id="121393", source=source)
-        chant_2 = make_fake_chant(incipit="baal", cantus_id="196418", source=source)
+        chant_1 = make_fake_chant(
+            manuscript_full_text_std_spelling="isaac", cantus_id="121393", source=source
+        )
+        chant_2 = make_fake_chant(
+            manuscript_full_text_std_spelling="baal", cantus_id="196418", source=source
+        )
+        chant_1.refresh_from_db()
```
dchiller: Not sure this is worth it at this point, but do you always need to `.refresh_from_db()` once you've created the chant. Why not return the refreshed version from `make_fake_chant`? Then you don't need this and the accompanying comment everywhere...

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1966,30 +2019,50 @@ def test_keyword_search_searching_all_fields(self):
...
+        # we make the request for the page.
+
         chant_incipit = make_fake_chant(
             source=source,
-            incipit=includes_search_term,  # <==
             manuscript_full_text=doesnt_include_search_term,
             manuscript_full_text_std_spelling=doesnt_include_search_term,
         )
+        Chant.objects.filter(id=chant_incipit.id).update(
+            incipit=includes_search_term  # <==
```
dchiller: What does this comment mean?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1966,30 +2019,50 @@ def test_keyword_search_searching_all_fields(self):
...
+        )
+
         chant_ms_spelling = make_fake_chant(
             source=source,
-            incipit=doesnt_include_search_term,
             manuscript_full_text=includes_search_term,  # <==
```
dchiller: Oh, I guess it means that this is what we will be searching for in this chant?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -1966,30 +2019,50 @@ def test_keyword_search_searching_all_fields(self):
         includes_search_term = "brevity is the soul of wit"
         doesnt_include_search_term = "longevity is the soul of wit"
         source = make_fake_source()
+
+        # For all the chants below: incipit is automatically calculated
```
dchiller: I don't understand for this test why we want to set the incipit to be something different than it would be as it is with the signal? It seems like this is testing a situation (where the incipit is created through means other than the signal) that should never happen!

Just looking through these:

For `chant_incipit`, I have a chant whose incipit does include the search term but neither of the full texts do...if this is happening, it's bad!

`chant_ms_spelling` and `chant_std_spelling` could be real objects in the database (the term is somewhere in the either of the full texts but not in the first 5 words), but in that case, why not make the test chants fit that real case rather than what you do here? (e.g. put the search term as word # 6).

dchiller: After looking through the code of the relevant view, I now see what this is testing... but to my mind, the fact that you have to skirt around the function this PR introduces to make this test work points to the fact that the view needs to change (yay using tests to identify these kinds of things!). Don't search `incipit` anymore in this context because it provides no information not in these other fulltext fields.

django/cantusdb_project/main_app/signals.py
jacobdgm: Thank you!

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: I have a hard time visually parsing series of items whose relevant difference is `includes_search_term` vs. `doesnt_include_search_term`. This was my effort to flag the relevant items that should be caught by the view.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: this is a good point. thanks for flagging this - I'll update the view!

jacobdgm: on second thought, this might be an item for a separate PR.

I haven't started trying to code this, but one thing that occurs to me is that there are several chants, inherited from OldCantus, for which we have only an incipit and no fulltext. So this will require a bit of thought.

(if we end up still searching the incipit field and the fulltext fields, we should at least have a test to check the incipit-yes—fulltext-no case, and not one that tests an impossible situation as we currently have)

dchiller: I see...yes, it seems like it would make sense to keep the view as is for those legacy cases where there is just an incipit, but agree that a change in the test so that the test chant "looks like" that legacy case would make the purpose behind it much clearer.

jacobdgm:
Going to mark this as a draft for now, so as to ensure no-one accidentally merges it in the next few days so we can keep the pipeline clear for #1323. Once we've updated the concordances API and I've implemented changes based on a few of @dchiller's suggestions, I'll reopen this and request an updated review.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: Took me a while to figure out what had been going on here - this was an adaptation of the previous method of identifying a matching search term. I've updated it to just use the beginning of the STD Spelling fulltext.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: This is a good observation/suggestion. I've implemented this!

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: So, I've not touched the corresponding view - we need to search through the incipit field as well as through the fulltext fields - but I've updated the test to clarify the situation in which we might expect to find a match in the `incipit` field but not in the other fields.

Thank you for spotting this!

dchiller (approved):

PR #1320, opened 2024-02-16
home page: change hamburger menu

### LINKED ISSUE #1319: We should make changes on the hamburger menu to decrease its width
On the hamburger menu, we should consider making various changes to decrease the overall width

- "Home Page" under "Home" is redundant, since there is no dropdown we should remove "Home Page"
- "Docs, Instructions, Tools" is taking up too much space. Instead it should just say "Instructions"

lucasmarchd01:
Fixes #1319 

Update: with screenshot!
Screenshot 2024-02-19 at 9 11 49 AM

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -168,9 +168,7 @@
 
                         <!-- HOME -->
                         <li class="nav-item">
-                            <a class="nav-link font-weight-bold" href="/">HOME<br style:"line-height: 50pt;"/><span
-                                    class="small">Home
-                                    Page</span></a>
+                            <a class="nav-link font-weight-bold" href="/">HOME</a>
                         </li>
 
                         <!-- ABOUT -->
@@ -253,7 +251,7 @@
                         <li class="nav-item dropdown">
                             <a class="nav-link dropdown-toggle font-weight-bold" href="#" id="navbarDropdownMenuLink"
                                 role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                                RESOURCES<br><span class="small">Docs, Instructions, Tools</span>
+                                RESOURCES<br><span class="small">Instructions</span>
                             </a>
                             <div class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
                                 <a class="dropdown-item" href="/tutorial/">Tutorial for Cantus Contributors</a>
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -175,7 +175,7 @@
                         <li class="nav-item dropdown text-info">
                             <a class="nav-link dropdown-toggle font-weight-bold" href="#" id="navbarDropdownMenuLink"
                                 role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                                ABOUT<br><span class="small">Information</span>
+                                ABOUT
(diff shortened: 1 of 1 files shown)


jacobdgm (changes requested):
The code looks fine - can you update the description with before/after screenshots so @fujinaga can confirm it matches his vision?

fujinaga:
@lucasmarchd01 Do put up the screenshots.

jacobdgm:
Thanks for the update with the screenshot! "Home" looks pretty lonely now, almost as though it was supposed to have a thing beneath it which was missed/forgotten. @fujinaga, what do you think?

One possibility could be to remove "Home" entirely - if people want to navigate to the homepage, they can click on the logo that's already present in the upper left. I checked a few big websites to see how they do it:
- Facebook has its "f" logo in the upper left corner, with no mouse-over text
- Google has its "Google" logo in the upper left corner, with "Go to Google Home" on mouse-over
- YouTube has its "YouTube" logo in the upper left corner, with "YouTube Home" on mouse-over

Currently, we have no `"title"` attribute on our upper-left logo. We could add it, saying "Cantus Database Home", and remove the "Home" from the title bar?

dchiller:
I recognize that these comments are out of scope of the original issue, but I will take the opportunity to add my thoughts about the current menu!

I feel that all of the menu subtitles are at best unnecessary and at worst confusing:

1. "About - Information": "Information" here feels totally meaningless. The whole site has information! If it said just "About" or if we thought it needed more detail, "About Cantus Database", then I think a user would know what they could get there...but "information"?
2. "Sources - List of Inventories": This is the one where I think the subtitle has the greatest claim to usefulness, but I'm not sure that "Source Inventories" or "Browse Sources" or frankly just "Sources" doesn't basically do the trick (that is, if we really don't want inconsistencies with having vs not having menu subtitles).
3. "Chants - Search":  "Chant Search"...All of the sub-pages here are various types of chant search, so why not just say that?
4. "Feasts - Lists of Feasts": Redundant.
5. "Terms - Genres and Fields": Why is "genres" singled out? "Fields" feels a little opaque? "Glossary" seems like it communicates in a single word much more of what is here.
6. "Resources - Instructions": Kind of like "About - Information", "Instructions" doesn't really add anything imho. "Instructions" for what? "Resources" is a catch-all and "Instructions" is only relevant to some of the resources there and even then doesn't really tell you anything. One possibility here is just have this be "More" (like Facebook has on the left-side menu) and then the catch-all-ness is really communicated.


I think this could be reduced to a menu with no subtitles and the look would be less cluttered and the clarity would in most cases be improved!

annamorphism:
> I feel that all of the menu subtitles are at best unnecessary and at worst confusing:


I agree completely! I think they are there to trigger the little drop-down arrows, but if that were rethought the whole thing would be both prettier AND clearer (not always the case!) and the hamburger problems would basically go away: win-win-win.

fujinaga:
@lucasmarchd01 can you make a mockup of the menu without the subtitles? It would be easier for people (mostly Debra) to evaluate different looks.

lucasmarchd01:
Screen Shot 2024-02-21 at 7 14 49 PM

I agree with what has been said, I think this is much better!

jacobdgm (changes requested):
I like the look of this much better.

If you can add `title="Cantus Database Home"` to the logo, I think this PR will be ready to go (or at least, ready to be put up on staging).

PR #1321, opened 2024-02-17
fix invalid markup

igneus:
(I tried correcting to `<br style="line-height: 50pt;">`, but it distorts the layout, suggesting that the malformed inline style is outdated and deleting it is the right thing to do.)

### DIFF
--- a/django/cantusdb_project/templates/base.html
+++ b/django/cantusdb_project/templates/base.html
@@ -168,7 +168,7 @@
 
                         <!-- HOME -->
                         <li class="nav-item">
-                            <a class="nav-link font-weight-bold" href="/">HOME<br style:"line-height: 50pt;"/><span
+                            <a class="nav-link font-weight-bold" href="/">HOME<br><span
                                     class="small">Home
                                     Page</span></a>
                         </li>


jacobdgm:
Thanks for suggesting this, @igneus. Alas, we're about to remove the implicated `<br>` entirely in #1320.

igneus:
A month later there's still broken markup on the cantusdatabase.org homepage (and throughout the site). Isn't it time to reconsider deploying a hotfix, even if overhaul of the layout is planned?

jacobdgm:
We appreciate you caring about this @igneus - again, thanks for noticing this and suggesting this change. The main reason I closed this PR, if I recall my motivations at the time, was to avoid merge conflicts with #1320.

In the end, #1338 superseded #1320, but the former was merged earlier today and is currently up for testing on https://staging.cantusdatabase.org. I'm optimistic that we can have these changes up on our Production server by Friday at the latest - the offending `<br>` should be gone by the end of the week.

Again, thanks for taking the time to put this together and to follow up on it, and I'm sorry it didn't make sense to merge the specific changes you proposed as part of this PR. We appreciate your patience as a fix for this bug you found makes its way through testing!

jacobdgm:
The new header has been deployed! `line-height: 50pt;` is no more!

Issue #1323, opened 2024-02-22
We need to update our concordances export API for new specification


jacobdgm:
In an email from Jan, who's working to use our [concordances API](https://github.com/DDMAL/CantusDB/wiki/APIs#concordances) on Cantus Index:

> There is a small issue with parsing your JSON: 
> You have a 'data' element and a Cantus ID element ("004650" in the following snippet) which is not necessary and not compatible with the current CI parser. 
> ```
> ["data":{"004650":[{"siglum":"A-KR VI\/258","srclink":"https:\/\/[cantusdatabase.org](http://cantusdatabase.org/)\/source\/123660\/","chantlink":"https:\/\/[cantusdatabase.org](http://cantusdatabase.org/)\/chant\/307346\/","folio":"060r","sequence":11,"incipit":"Rex autem David cooperto","feast":"De Regum","genre":"A","office":"H","position":"9","cantus_id":"004650","image":null,"mode":"?","full_text":null,"melody":null,"db":"CD"},{"siglum":"SI-Lna 19 (olim 18)","srclink":"https:\/\/[cantusdatabase.org](http://cantusdatabase.org/)\/source\/123659\/","chantlink":"https:\/\/[cantusdatabase.org](http://cantusdatabase.org/)\/chant\/356859\/","folio":"021v","sequence":2,"incipit":"Rex autem David cooperto","feast":"De Regum","genre":"A","office":"H","position":"10","cantus_id":"004650","image":null,"mode":"8","full_text":null,"melody":null,"db":"CD"},
>```
> 
> Is it possible to remove these elements and keep only what follows?
> ```
[...]

PR #1327, opened 2024-02-22
Display useful placeholders for when user full_name is not set

### LINKED ISSUE #1310: In source description sidebar, we should display a useful placeholder when a user's full_name is not set
On our User Detail page, we display "User" when a user's `full_name` is not set.

We should consider doing something similar in the sidebar that can be found on the Source/Chant Detail page as well. In the screenshot below, Inventoried by is set to user 251687, whose email is set but who has no full_name.

Screenshot 2024-02-13 at 9 16 04 AM

jacobdgm:
This PR touches several templates, displaying `User <id>` in places where we used to just display `None` or `User`. The implicated templates are:
- `chant_detail.html`
- `indexer_list.html`
- `source_detail.html`
- `user_detail.html`
- `user_list.html`

#1310 specifically noted when this appeared in the source description sidebar, which appeared on both the Chant Detail page and the Source Detail page. This PR thus fixes #1310

One place I didn't attempt to make any changes is the Content Overview page. In case of no `full_name`, I believe this will end up displaying `User Object`. I don't understand the logic on this page, and didn't feel it was worth trying to unpick it since only a few highly privileged users have access to this page.


dchiller (approved):
In this case, or in a case like it, I think it might be better to put this in the `User` model once rather than have this `if/then` logic in each template. That way, if you ever wanted to change this logic (let's say, display the user institution rather than the id, or even something more complicated) you wouldn't have to find every template where this logic is used, but you could just change the logic on the model. 

I'm approving because this is already done and I don't know that this is a situation where we'd expect much further development along these lines, so I don't think changing it now will pay too many dividends. But just to say that that would be, I think, the way to go in future cases.

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -444,17 +448,25 @@ <h4>List of melodies</h4>
...
+                                            <br>
                                         {% endfor %}
                                     </ul>
                                 {% endif %}
 
                                 {{ source.indexing_notes|default_if_none:"" }}
                                 <br>
                                 {% if source.created_by %}
-                                    Contributor: <a href={% url 'user-detail' source.created_by.id %}><b>{{ source.created_by.full_name }}</b></a>
+                                    {% if source.created_by.full_name %}
```
ahankinson: Just a flag to remind you to make sure this model traversal (`a.b.c`) is optimized. This sort of code is exactly the kind that will introduce lots of SQL queries if you're not careful.

jacobdgm:
Going to mark this as a draft for now, as we want to keep the pipeline clear for #1323 and I don't want to accidentally merge this.

Before merging eventually, I'd like to wrap the changed places in the code in `{% with %}` blocks - I don't think it will matter much in this case - for all the sources I'm aware of, at most small handful of people edit/proofread/index each of them, but this does feel like best practice and I'd like to build some muscle memory for it. Thanks for this suggestion/reminder, @ahankinson!

jacobdgm:
I ran into difficulties when I attempted to push, so I attempted to pull and rebase; thus a bunch of extra commits got added to this PR - hopefully this has not created problems.

I added a `{% with creator=source.created_by %}`  to the Source Detail page and an analogous `with` to the Chant Detail page; they don't seem to decrease the number of SQL queries generated, but on the Chant Detail page at least, the page loads a few milliseconds faster. As I said, I don't think it makes much of a difference here, but this is a good thing for me/us to keep an eye out for elsewhere.

dchiller:
My thoughts are the same as above (ie. approved). I'm not sure what the `with` block was supposed to do? 

Although changes already introduced to `develop` are repeated here?

jacobdgm:
> I'm not sure what the `with` block was supposed to do?

see https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#with:

> [`with`] is useful when accessing an “expensive” method (e.g., one that hits the database) multiple times.

This change was added in response to @ahankinson's comment,

> Just a flag to remind you to make sure this model traversal (`a.b.c`) is optimized. This sort of code is exactly the kind that will introduce lots of SQL queries if you're not careful.

---

> Although changes already introduced to `develop` are repeated here?

> I ran into difficulties when I attempted to push, so I attempted to pull and rebase; thus a bunch of extra commits got added to this PR - hopefully this has not created problems.

ahankinson:
`with` blocks only assign variables, they don’t change the underlying SQL. You need to add ` select _related ` to the underlying query set to join those tables.

I don’t know if it makes a difference here. I was just flagging it as the kind of code that usually causes slow page loads if you’re not paying attention.

PR #1333, opened 2024-02-27
Update update_cached_concordances command to comply with new spec

### LINKED ISSUE #1323: We need to update our concordances export API for new specification

jacobdgm:
This PR updates our `update_cached_concordances` command to output a file matching the updated specification (see #1323). In doing so, it fixes #1323.

The tests have also been updated to reflect the new spec.

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
+++ b/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
@@ -6,9 +6,11 @@ from django.db.models.query import QuerySet
 from django.core.management.base import BaseCommand
 from main_app.models import Chant
 
-
 # Usage: `python manage.py update_cached_concordances`
 # or `python manage.py update_cached_concordances -d "/path/to/directory/in/which/to/save/concordances"`
+
+# Used for creating URIs for chant and source detail pages
+CANTUSDB_DOMAIN: str = "https://cantusdatabase.org"
 
 
 class Command(BaseCommand):
@@ -97,11 +99,10 @@ def make_chant_dict(chant: dict) -> dict:
         dict: A dictionary representing a chant, with several values standardized
             (e.g., `siglum`) and several values calculated (e.g., `srclink`)
     """
-    DOMAIN: str = "https://cantusdatabase.org"
     source_id: int = chant["source_id"]
-    source_uri: str = f"{DOMAIN}/source/{source_id}/"
+    source_uri: str = f"{CANTUSDB_DOMAIN}/source/{source_id}/"
     chant_id: int = chant["id"]
-    chant_uri: str = f"{DOMAIN}/chant/{chant_id}/"
+    chant_uri: str = f"{CANTUSDB_DOMAIN}/chant/{chant_id}/"
     processed_chant: dict = {
         "siglum": chant["source__siglum"],
         "srclink": source_uri,
--- a/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
+++ b/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
@@ -122,36 +122,3 @@ def make_chant_dict(chant: dict) -> dict:
     }
 
     return processed_chant
-
-    # stdout.write("Processing chants\n")
-    # concordances: defaultdict = defaultdict(list)
-    # for chant in values:
-    #     source_id: int = chant["source_id"]
-    #     source_absolute_url: str = f"{DOMAIN}/source/{source_id}/"
-    #     chant_id: int = chant["id"]
-    #     chant_absolute_url: str = f"{DOMAIN}/chant/{chant_id}/"
-
-    #     concordances[chant["cantus_id"]].append(
-    #         {
-    #             "siglum": chant["source__siglum"],
-    #             "srclink": source_absolute_url,
-    #             "chantlink": chant_absolute_url,
-    #             "folio": chant["folio"],
-    #             "sequence": chant["c_sequence"],
-    #             "incipit": chant["incipit"],
-    #             "feast": chant["feast__name"],
-    #             "genre": chant["genre__name"],
-    #             "office": chant["office__name"],
-    #             "position": chant["position"],
-    #             "cantus_id": chant["cantus_id"],
-    #             "image": chant["image_link"],
-    #             "mode": chant["mode"],
-    #             "full_text": chant["manuscript_full_text_std_spelling"],
-    #             "melody": chant["volpiano"],
-    #             "db": "CD",
-    #         }
-    #     )
-
-    # stdout.write(f"All chants processed - found {len(concordances)} Cantus IDs\n")
-
-    # return dict(concordances)
--- a/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
+++ b/django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
@@ -45,13 +45,13 @@ class Command(BaseCommand):
         )
 
 
-def get_concordances() -> dict:
-    """Fetch all published chants in the database, group them by Cantus ID, and return
-    a dictionary containing information on each of these chants.
+def get_concordances() -> list[dict]:
+    """Fetch all published chants in the database, and return a list
+    of dictionaries containing information on these chants.
 
     Returns:
-        dict: A dictionary where each key is a Cantus ID and each value is a list all
-          published chants in the database with that Cantus ID.
+        list[dict]: A list of dictionaries, each representing a single chant
+            from the database
     """
 
     stdout.write("Querying database for published chants\n")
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -89,35 +81,79 @@ def get_concordances() -> dict:
...
+        "image": chant["image_link"],
+        "mode": chant["mode"],
+        "full_text": chant["manuscript_full_text_std_spelling"],
+        "melody": chant["volpiano"],
+        "db": "CD",
+    }
+
+    return processed_chant
+
+    # stdout.write("Processing chants\n")
```
dchiller: Did you mean to keep these commented lines?

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -62,7 +55,6 @@ def get_concordances() -> dict:
         dict: A dictionary where each key is a Cantus ID and each value is a list all
```
dchiller: The doctring for this function is out of date since there is now no more grouping....

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -89,35 +81,79 @@ def get_concordances() -> dict:
...
+    Args:
+        chant (dict): A dictionary representing a chant from the database,
+            from a QuerySet.values() call, with several non-standardized
+            keys from database traversals (e.g., `source__siglum`)
+
+    Returns:
+        dict: A dictionary representing a chant, with several values standardized
+            (e.g., `siglum`) and several values calculated (e.g., `srclink`)
+    """
+    DOMAIN: str = "https://cantusdatabase.org"
```
ahankinson: This is a constant and doesn’t need to be redeclared every time the function is called.

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -56,57 +51,33 @@ def test_concordances_structure(self):
...
 
-        concordances: dict = update_cached_concordances.get_concordances()
-        concordances_for_single_id: list = concordances["123456"]
+        concordances: list = update_cached_concordances.get_concordances()
         self.assertEqual(len(concordances), 1)
 
-        single_concordance: dict = concordances_for_single_id[0]
-        expected_incipit: str = published_chant.incipit
+        single_concordance: dict = concordances[0]
+        # since the fulltext, specified above, contains <= 5 words,
```
dchiller: I would remove this comment and either adjust the test fully in anticipation of #1318 (remove reference to incipit altogether) or keep the creation of fake chants with incipits for this PR and adjust in #1318. The comment is not accurate as of this PR (incipits on fake chants are limited by characters, not words) and in the event that the incipit length is changed in the signal that will soon be introduced, this test could begin to fail.

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
```diff
@@ -89,35 +81,79 @@ def get_concordances() -> dict:
...
+        "mode": chant["mode"],
+        "full_text": chant["manuscript_full_text_std_spelling"],
+        "melody": chant["volpiano"],
+        "db": "CD",
+    }
+
+    return processed_chant
+
+    # stdout.write("Processing chants\n")
+    # concordances: defaultdict = defaultdict(list)
```
dchiller: Is `defaultdict` now an unused import?

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: oof - I'll remove this. thanks for flagging this.

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: good call - thank you!

django/cantusdb_project/main_app/management/commands/update_cached_concordances.py
jacobdgm: you're right - thanks!

django/cantusdb_project/main_app/tests/test_functions.py
jacobdgm: ah - when I was writing this, I was under the impression that our dictionary contained only an `"incipit"` key and nothing relating to fulltext. But I see we do have a `"full_text"` key after all. Thanks for prompting me to think a bit more about this - I'll update it!

jacobdgm:
I spent a few minutes looking into whether we could rename `"feast__name"` -> `"feast"` etc. within `get_concordances` when calculating `values` by using `.annotate()` , without having to have a separate `make_chant_dict` function that gets called in the list comprehension. But initial results suggest it will be tricky to create `"chantlink"` and `"srclink"` on the database level. Considering the function already takes but a couple of seconds to run, I'm thinking it's not going to be worth the trouble to apply this particular optimization. Or do any of you tagged as reviewers think this is worth looking into?

dchiller:
Were you using concat for `srclink` and `chantlink`? https://docs.djangoproject.com/en/5.0/ref/models/database-functions/#concat I think that would be the way to do what you are referring to.

I don't feel like this is necessary unless you want to.

dchiller (approved):

jacobdgm:
> Were you using concat for `srclink` and `chantlink`? https://docs.djangoproject.com/en/5.0/ref/models/database-functions/#concat I think that would be the way to do what you are referring to.
> 
> I don't feel like this is necessary unless you want to.

This is it, yes!

Maybe I'll get around to refactoring this once we've finished rolling it out as it is. For now, though, I think we're fine as-is.

PR #1337, opened 2024-02-29
Streamline header (subtle option)

jacobdgm:
This PR streamlines the header that appears on each page. In specific, it:
- shortens the subtitle under RESOURCES
- removes the subtitle under HOME
- adds a `title` attribute to the title so that "Cantus Database Home" is displayed on mouseover

Before:
Screenshot 2024-02-29 at 4 44 52 PM

After:
Screenshot 2024-02-29 at 4 46 00 PM


jacobdgm:
Note that this version does not allow us to trigger the expanded-collapsed breakpoint from `xl` to `lg`, as is the case with #1338.

jacobdgm:
Current plan is to merge this PR and send it via Staging to Production quickly (it should not ruffle feathers, we expect), and then put #1338 up on Staging to show to Debra.

django/cantusdb_project/templates/base.html
```diff
@@ -167,8 +167,7 @@
         <div class="container p-0">
             <nav class="navbar navbar-expand-xl navbar-light bg-white border-top col-12">
                 <a class="navbar-brand" href="/">
-                    <img src="{% static "CantusLogoSmall.gif" %}"
-                        alt="Cantus Database: A Database for Latin Ecclesiastical Chant" loading="lazy">
+                    <img src="{% static "CantusLogoSmall.gif" %}" title="Cantus Database Home" alt="Cantus Database Logo" loading="lazy">
```
dchiller, line 170: Better alt text :) good!

dchiller (approved):

jacobdgm:
Obviated by the merging of #1338.

PR #1338, opened 2024-02-29
Streamline header (drastic option)

### LINKED ISSUE #1319: We should make changes on the hamburger menu to decrease its width

jacobdgm:
This PR streamlines the header that appears on each page. In specific, it:
- removes all subtitles
- removes the "HOME" navbar item
- adds a `title` attribute to the title so that "Cantus Database Home" is displayed on mouseover
- changes the bootstrap breakpoint for displaying the hamburger menu from `xl` to `lg` (i.e. fixes #1319)
- Adds a link to the homepage to the "ABOUT" navbar item
- Adds query parameters to the two links to the homepage, allowing us to track how often users click on each

Before:
Screenshot 2024-02-29 at 4 44 52 PM

After:
Screenshot 2024-02-29 at 4 47 04 PM


jacobdgm:
Current plan: once #1338 is deployed to Production, we'll merge this and put it on Staging in order to show it to Debra.

dchiller (approved):
Looks good to me...approving for whenever we're ready for this merge.

jacobdgm:
We showed this to Debra, and both she and Jennifer like it!

One thing she suggested is that we add a "Home" as the first item in the "About" dropdown. My initial thoughts are that this isn't necessary, and perhaps a bit confusing. I feel like "Home" is not really an "About" thing - if I were to see this on another site, I might get frustrated that the label didn't match the contents of the tin, so to speak.

Surveying a handful of big sites, I see that:
- Google.com has a logo in the upper left, that says something other than "Google Home" on cursor hover, and there isn't really another way to get to the homepage.
- Facebook.com has a logo in the upper left, that says nothing on cursor hover. There's also a little house-looking icon in the navbar that takes you home.
- Youtube.com has a logo in the upper left, that says "YouTube Home" on cursor hover. There's also a "Home" in the nav sidebar
- Amazon.ca has a logo in the upper left, that says nothing on cursor hover, and there is no other way to get back to the homepage.

So it seems sites are divided as to whether to include another way to return to the homepage, but for those which add this additional way home, they don't put it under "About" (i.e., they do it as we've been doing it up until now, with a logo plus an immediately apparent "Home").

jacobdgm:
One thought: we can add "Home" under "About" as Debra suggested, but make the logo link to `/?ref=navbar_logo` and the dropdown one link to `/?ref=navbar_about`.

After a week or so, we can check the logs and observe the ratio of `navbar_logo`s to `navbar_about`s in the Nginx traffic. If the ratio is higher than, say, 20:1, we can confidently tell Debra that people are finding their way home using the logo without difficulty.

jacobdgm:
I posted a little survey to facebook, dropping people at a source detail page and asking them to get back home. The three people who have responded so far have said it took them ~1s. But one commented "I wouldn't say that it's intuitive, more a trained response". Another said that if they "were looking for a text-based link on the site that visually tells me that I'm going home I couldn't find one. I went looking in the footer for a sitemap style collection of links, or a HOME link in the top navigation menu. If I were a person less familiar with the conventions of the internet I might have had a harder time."

The latter of these people also commented "And a small note on accessibility: The alt text on the logo saying where the link goes is good, but the text in the image is small enough that it's hard to read for well-sighted people, and doesn't exist at all for people who rely on the alt text of images." They conflated alt text and title text, but this raises a thing that could be worth discussing - should the alt text read "CANTUS: A Database for Latin Ecclesiastical Chant"?

I don't really know how big a problem this actually presents for people "less familiar with the conventions of the internet". I think that gathering data via query parameters could shed some light on the issue, so I'll implement it.

dchiller:
FWIW, I share the concern that having a "Home" link under the "About" menu is confusing...if there is a worry that people won't easily be able to make their way to the home page because clicking on the header logo is unfamiliar or confusing, I think we should just keep an additional "Home" button in the navbar. To me, it would be *more* unintuitive to know that I am supposed to go to the "About" menu to get home as it would be to know to click the header logo.

dchiller:
> _more_ unintuitive

some might also know this as *less* intuitive :no_mouth:

dchiller:
> They conflated alt text and title text, but this raises a thing that could be worth discussing - should the alt text read "CANTUS: A Database for Latin Ecclesiastical Chant"?

So the idea is keep the title text as is ("Cantus Database Home") but potentially change the alt text from "Cantus Database Logo" to "CANTUS: A Database for Latin Ecclesiastical Chant"?

Just from some quick googling I found this page (https://accessibility.huit.harvard.edu/describe-content-images) with some details if you scroll down on how to treat "Icons" and "Images as Links" and this page (https://help.siteimprove.com/support/solutions/articles/80000863904-accessibility-image-alt-text-best-practices) scroll down to "Functional Images"...

from those examples (which seem probably reputable) it seems like the alt text convention would be "Cantus Database Home" (i.e. the current title text).

I really don't know the proper way to do this (again, this was just a few minutes of googling) so I don't feel attached.

jacobdgm:
> from those examples (which seem probably reputable) it seems like the alt text convention would be "Cantus Database Home" (i.e. the current title text).
> 
> I really don't know the proper way to do this (again, this was just a few minutes of googling) so I don't feel attached.

Looking at https://accessibility.huit.harvard.edu/technique-use-real-text, it seems that "CANTUS: A Database for..." is the right thing to do.

But looking at https://help.siteimprove.com/support/solutions/articles/80000863904-accessibility-image-alt-text-best-practices#Functional-images-(Linked-image), it seems that "Cantus Database Home" is the preferred option.

In terms of accessibility, our job would be much easier if the logo simply read "CANTUS" with no text beneath, alas.

jacobdgm:
Following https://www.w3.org/WAI/tutorials/images/decision-tree/:

Does the image contain text? - yes
… and the text has a specific function, for example is an icon. - yes
-> Use the alt attribute to communicate the function of the image. See [Functional Images](https://www.w3.org/WAI/tutorials/images/functional/).

So W3 seems to think the preferred approach is "Cantus Database home"

dchiller:
> Looking at https://accessibility.huit.harvard.edu/technique-use-real-text, it seems that "CANTUS: A Database for..." is the right thing to do.

I think "a functional" is more specific and relevant to this case than "an image with text in it" so I think the latest commit is better.

jacobdgm:
How do you feel about the `?ref=navbar_about` aspect of this? Should we try this empirical approach, or just tell Debra we think the site would be better without that item in the About menu?

dchiller:
> How do you feel about the `?ref=navbar_about` aspect of this? Should we try this empirical approach, or just tell Debra we think the site would be better without that item in the About menu?

I think the empirical approach in combination with what you already sent via email makes sense!

dchiller (approved):

jacobdgm:
> After a week or so, we can check the logs and observe the ratio of `navbar_logo`s to `navbar_about`s in the Nginx traffic. If the ratio is higher than, say, 20:1, we can confidently tell Debra that people are finding their way home using the logo without difficulty.

So, I just checked, and we have some results from our empirical approach in traffic from Mar 22 up to a few minutes ago:
- Running `docker logs d459947b2481 2>&1 | grep "GET /?ref=" | grep -v "lychee" | grep -v "bot" | grep -v "crawl" | grep -v "spider"` (I.e. search STDOUT for the logs for the nginx container for all instances of `GET /?ref=`, filtering out anything containing `lychee`, `bot`, `crawl` or `spider`), we get **185 matching lines**.
- Searching this output, there are
  - **85** instances of `GET /?ref=navbar_about`
  - **100** instances of `GET /?ref=navbar_logo`.
- So, slightly more people are going to the homepage via the logo, but really it's about 50-50.

While I still think having a link to the homepage under "About" is inelegant, it seems Debra and Jennifer were founded in their concern about having the logo be the only way to the homepage - about half of website users are looking within "About" and clicking "Home" instead of just clicking on the logo.

So if we're serious about being guided by the data we've collected, I guess we open another PR to remove the `?ref` QPs, but leave "Home" within "About"?

dchiller:
> So if we're serious about being guided by the data we've collected, I guess we open another PR to remove the `?ref` QPs, but leave "Home" within "About"?

Fascinating... as mentioned, not what I would have expected, but what a great test you came up with :) I think this makes sense.

PR #1340, opened 2024-03-01
Add `default_server` to nginx development configuration

### LINKED ISSUE #1331: We should specify a default server in our nginx.conf
> On the other hand, this made me realize that since we don't have a default server block in our nginx config, requests to the IP are getting through to django even though we've specified server_name's in all blocks in `nginx.conf`.

_Originally posted by @dchiller in https://github.com/DDMAL/CantusDB/issues/1326#issuecomment-1960422968_

jacobdgm:
This PR removes existing `default_server`s statements from `cantusdb.conf.staging` and `cantusdb.conf.production`, introducing new server blocks with `default_server` to handle and deny requests not matching a proper hostname. it fixes #1331

This PR also removes commented-out lines from `cantusdb.conf.development` and `cantusdb.conf.staging` - I had initially thought these would be helpful, but `cantusdb.conf.production` has been updated and this change has not been added to the comments in the two other files, and I was finding that the extra commented lines make things difficult to read anyways.

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -3,7 +3,7 @@
 # contents of this file.
  
 server {
-    listen 80 default_server;
+    listen 80;
 
     location / {
         proxy_pass http://django:8000;
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -3,11 +3,7 @@
 # contents of this file.
  
 server {
-    # listen 443 default_server http2 ssl;
     listen 80 default_server;
-	
-    # ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
-    # ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
 
     location / {
         proxy_pass http://django:8000;
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -1,38 +1,10 @@
 # This file is configured for deployment of the CantusDB project in local development.
 # When building the project locally, replace the contents of cantusdb.conf with the
 # contents of this file.
-
-# server {
-#     listen 80;
-# 
-#     server_tokens off;
-# 
-#     location ^~ /.well-known/acme-challenge/ {
-#         root /var/www/lego;
-#     }
-# 
-#     location / {
-#         return 301 https://$host$request_uri;
-#     }
-# }
- 
-# server {
-#     # Redirect all https traffic for mass.cantusdatabase.org to cantusdatabase.org
-#     listen 443 ssl;
-# 
-#     server_name mass.cantusdatabase.org;
-# 
-#     ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
-#     ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
-# 
-#     location / {
-#         return 301 https://cantusdatabase.org$request_uri;
-#     }
-# }
  
 server {
     # listen 443 default_server http2 ssl;
-    listen 80;
+    listen 80 default_server;
 	
     # ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
     # ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
--- a/config/nginx/conf.d/cantusdb.conf.production
+++ b/config/nginx/conf.d/cantusdb.conf.production
@@ -79,3 +79,18 @@ server {
         root /;
     }
 }
+
+# Default server to block requests to IP address and non-valid hostnames
+server {
+    access_log off; 
+    server_name _;
+
+    listen 80 default_server;
+    listen 443 default_server;
+    ssl_reject_handshake on;
+
+    # Close connection immediately for requests that don't match a defined
+    # server name
+    return 444;
+}
+
--- a/config/nginx/conf.d/cantusdb.conf.production
+++ b/config/nginx/conf.d/cantusdb.conf.production
@@ -31,10 +31,10 @@ server {
 }
 
 server {
-    listen 443 default_server http2 ssl;
+    listen 443 http2 ssl;
 
     server_name cantusdatabase.org;
-	
+
     ssl_certificate /etc/nginx/ssl/live/certificates/cantusdatabase.org.crt;
     ssl_certificate_key /etc/nginx/ssl/live/certificates/cantusdatabase.org.key;
 
@@ -56,7 +56,7 @@ server {
         alias /resources/api_cache/concordances.json;
         expires modified +24h;
     }
- 
+
     location = /style.css {
         root /;
     }
--- a/config/nginx/conf.d/cantusdb.conf.staging
+++ b/config/nginx/conf.d/cantusdb.conf.staging
@@ -32,7 +32,7 @@ server {
 }
 
 server {
-    listen 443 default_server http2 ssl;
+    listen 443 http2 ssl;
 
     server_name staging.cantusdatabase.org;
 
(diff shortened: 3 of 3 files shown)


dchiller:
This is not really what I meant in the comment that sparked this issue...I should probably have been more specific in the issue. 

(I approve getting rid of the commented lines, so if you'd still like to do that, all good!)

- Adding `default_server` where you have in the development `nginx.conf` file makes the default explicit, but that server was already the default one by virtue of being the first (and only, for that matter) server block.
- What prompted the issue was noticing that a request with an incorrect host (or with no host) to the production server on port 80 is getting a 301 redirect to `https://cantusdatabase.org` and a request with an incorrect host (or with no host) to the production server on port 443 is getting passed to the `django` container (gunicorn). And mutatis mutandis with staging. So this doesn't affect that (as you say, it doesn't touch the actual configurations on staging or production).

dchiller:
I guess the upshot is that I approve if we de-link this from the issue b/c it otherwise looks fine.

jacobdgm:
so I'm confused - we do have a default server specified in `cantusdb.conf.production`: https://github.com/DDMAL/CantusDB/blob/259a1af7508d91303a3d6c1dcf2ca4485a6fcf24/config/nginx/conf.d/cantusdb.conf.production#L34
It's also in the analogous place in `cantusdb.conf.staging`. `.development` was the only nginx conf file I could find where we weren't specifying a `default_server`, hence this PR.

You've referred several times to `nginx.conf` - is this shorthand for `cantusdb.conf`, or am I missing something?

dchiller:
> You've referred several times to `nginx.conf` - is this shorthand for `cantusdb.conf`, or am I missing something?

Yes, sorry...I mean the nginx config file, `cantusdb.conf`.

dchiller:
> we do have a default server specified in `cantusdb.conf.production`

Sorry, I think my language in the original comment was probably misleading. It's not so much that we need to specify *a* default server (as you note, we have either done that explicitly in both our `.production` and `.staging` configurations in the case of our `443` port, or, in the case of our `80` port, we only have the single server block so it is the default by, well, default. It's that we need to specify a *better* default server block (we need to make a new server block that we will set as default) so that the case that I mention in the original comment and above (requests with bad hosts being accepted) don't happen.

Something like: 

```
server {
    listen 80 default_server;
    return 400;
}
```

Since we are moving away from the multiple `cantusdb.conf` files anyway with the ansiblized deployment, I think we can just merge the PR as is for now, unless you want to add to this PR.

ahankinson:
Thanks to this issue I've had a look at our own configuration for RISM Online. Here's what I've come up with:

```
    # A default server block to weed out all the direct
    # IP address and non-valid hostnames.
    server {
        access_log off;
        server_name _;

        listen 80 default_server;
        listen [::]:80 default_server;

        listen 443 default_server;
        listen [::]:443 default_server;
        ssl_reject_handshake on;

        # Any request that does not match a defined server name
        # will have its connection closed immediately.
        return 444;
    }
```

A bit more context: You can set a `Host` header manually, while still targeting the IP address of your website. With this you can do "host name spoofing" where you try to bypass name-based virtual hosts. Sometimes, if the webserver is not configured correctly, you could have a "private" admin tool (phpmyadmin) being the default virtual host if no actual hostname matches, just by virtue of it being the first configuration loaded. 

By setting an unnamed server name (`server_name _`) you're catching all requests for virtual hosts other than the ones configured. This then returns a `444` response, which immediately closes the connection. 

See: https://nginx.org/en/docs/http/request_processing.html#how_to_prevent_undefined_server_names

A few extra bits: Since you don't want to do anything with the request, there's no need to do any SSL stuff, so you can also have it listen to 443, and just immediately reject any SSL attempts, instead of configuring a certificate to only then reject it.

Also, there's not really a need to log any of it (at least, it's not particularly interesting to me) so I just turned off access log entries for this whole server block.

jacobdgm:
to summarize, just to make sure I'm understanding correctly (this summary assumes we're talking about the Production server):
- We have a `default_server` specified, but it currently handles any traffic sent to domains other than `cantusdatabase.org`, `www.cantusdatabase.org` and `mass.cantusdatabase.org`.
- However, that server block should really only handle traffic to `cantusdatabase.org`.
- We should remove all the instances of `default_server` that are currently in `cantusdb.conf`, and create a new server block to serve as our `default_server` which will handle all requests to non-valid hostnames
  - This new block will listen on both port 80 and port 443, and will reject any requests that come to it (either via a 444 response or via `ssl_reject_handshake`)

Question: I see, @ahankinson, that you've included both `listen 80` and `listen [::]:80`. A little bit of research suggests that the latter of these allows our server to handle IPv6 traffic in addition to IPv4 traffic. Should we be adding an analogous line with `[::]:` elsewhere in our configuration wherever we have a `listen`?

dchiller:
The `default_server` designation indicates which server block handles an incoming request when the incoming request does not match any `server_name`. The nginx documentation has a useful [page](https://nginx.org/en/docs/http/request_processing.html) that gives more details. If no server block has `default_server`, the first block in the configuration file (for the given port) is the server that handles the request. 

> * We have a `default_server` specified, but it currently handles any traffic sent to domains other than `cantusdatabase.org`, `www.cantusdatabase.org` and `mass.cantusdatabase.org`.

Technically, in our current configuration, there are two server blocks that can be thought of as a "default": the first `server` (the only one listening on port `80`) and the last `server` (the one listening on port `443` with the `default_server` designation). Any request coming in on port `80` will currently be handled by the first server (because it is the first server block in our configuration listening on port `80` and there is no other server block listening on port `80` that has been designated as the default with `default_server`). Any request coming in on `443` that is not to `www.cantusdatabase.org` or `mass.cantusdatabase.org` will be handled by the `server` listening to `443` with the `default_server` designation. The default server, whether designated by `default_server` or by its position in the configuration file, will handle all requests on the given port that don't match a `server_name`. Sorry if this is nit-picking, but I figure it may be helpful. 

> * However, that server block should really only handle traffic to `cantusdatabase.org`.

Correct, with the addition that is should only handle ssl traffic to `cantusdatabase.org`. We are saying we only want to really return `200` responses to `https` requests to `cantusdatabase.org`. If you request on port `80` and/or with a host alias (`mass.` or `www.`) we will redirect you, but anything else you come with, we are not going to respond.

ahankinson:
I don't know if this is helpful? I've pasted two commands and their results below.

The first is a cURL request to the cantus database server, using the IP address, but spoofing the `Host` header to something nonsensical. You can see that the server actually responds with a 301 to https://cantusdatabase.org. 

However, this isn't defined behaviour. The fact that it caught this isn't really by design -- it just caught the port 80 handler for http://cantusdatabase.org/ and returned that. It definitely **didn't** catch it by matching the `server_name`.  If your order of hostname config file parsing changes, it could actually land on another host handler that might not be as useful. 

By setting a "catch-all" default server, you can actually route all non-matching hostnames to a dedicated "go away" handler. I've demonstrated it with the same request to the RISM Online server, and you can see that it simply closes the connection. 

Cantus Database
```
❯ curl -vvv -H "Host: smooflydoo.org" http://206.12.93.196/
*   Trying 206.12.93.196:80...
* Connected to 206.12.93.196 (206.12.93.196) port 80
> GET / HTTP/1.1
> Host: smooflydoo.org
> User-Agent: curl/8.4.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Server: nginx
< Date: Tue, 05 Mar 2024 15:37:08 GMT
< Content-Type: text/html
< Content-Length: 162
< Connection: keep-alive
< Location: https://cantusdatabase.org/
<
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
* Connection #0 to host 206.12.93.196 left intact
```

RISM Online

```
❯ curl -vvv -H "Host: smooflydoo.org" http://195.15.241.20/
*   Trying 195.15.241.20:80...
* Connected to 195.15.241.20 (195.15.241.20) port 80
> GET / HTTP/1.1
> Host: smooflydoo.org
> User-Agent: curl/8.4.0
> Accept: */*
>
* Empty reply from server
* Closing connection
curl: (52) Empty reply from server
```

ahankinson:
PS: If you look at your log files you'll also now see a request for the site "smooflydoo.org".

ahankinson:
> should we be adding an analogous line with [::]: elsewhere in our configuration wherever we have a listen?

No, because your server doesn't route IPv6. If you check the instances in the Arbutus OpenStack you'll see the difference between the servers that have IPv6, and those that don't. 

So you can remove that bit from your config.

dchiller (approved):

PR #1332, opened 2024-02-23
Change header from bootstrap "container-fluid" to "container"

jacobdgm:
This PR updates the header on all pages to be the same width as the main content panel + any sidebars beneath it. This brings the CantusDB logo back from way off on the left, where it used to be. It fixes #1330

This PR also updates many of the templates. Some of these changes were made necessary by then `"container-fluid"`-to-`"container"` change on `base.html`, but others smooth out the formatting of the sidebar (mostly, adding a margin to the bottom of the search bar, which had previously been smooshed up against the sidebar card beneath it). On the chants-by-cantus-ID page, a div with `table-responsive` was necessary to prevent the table from spilling out of the main panel.

In the header, we added a `<br>` to "Docs, Instructions, Tools", thus preventing the header from spilling beyond its white background when between `lg` and `xl` breakpoints (this change also holds when the hamburger menu is expanded).

While working through the templates, I also renamed several of the templates, simplifying and standardizing them (`source_create_form.html`->`source_create.html`, `source_confirm_delete.html`->`source_delete.html`, `chant_confirm_delete.html`->`chant_delete.html`). `urls.py`, several of the views, and `test_views.py` were updated to accommodate these changes.

Much credit goes to @yinanazhou, whose help was invaluable in tracking down which bits of bootstrap needed to be tweaked and in identifying when a solution that went beyond bootstrap was necessary.

### Source Create page
(these screenshots representative of pages that have no cards in the sidebar)
#### before
Screenshot 2024-02-23 at 3 43 30 PM
#### after
Screenshot 2024-02-23 at 3 44 38 PM

### My Sources page
(these screenshots representative of pages that have only a single card in the sidebar)
#### before
Screenshot 2024-02-23 at 3 22 58 PM
#### after
Screenshot 2024-02-23 at 3 22 00 PM

### Chant Detail page
(these screenshots representative of other pages that have two cards in the sidebar)
#### before
Screenshot 2024-02-23 at 3 37 33 PM
#### after
Screenshot 2024-02-23 at 3 38 42 PM

### Source Detail page
(these screenshots representative of other pages that have three cards in the sidebar)
#### before
Screenshot 2024-02-23 at 3 46 46 PM
#### after
Screenshot 2024-02-23 at 3 47 21 PM

(these screenshots representative of the other pages at narrow breakpoints)
#### before, narrow breakpoint
Screenshot 2024-02-23 at 3 33 46 PM
#### after, narrow breakpoint
[...]

PR #1343, opened 2024-03-01
Fix layout for flatpages, articles templates

### LINKED ISSUE #1342: Flatpages template and articles template don't look nice with the new header
further to #1332 - try interacting with them on Staging between the medium and large breakpoints

jacobdgm:
Fixes #1342

- Article List and Article Detail page both have one card in the sidebar; I applied the same changes as in `user_source_list.html` in #1332.
- Flatpages `default.html` has three cards in sidebar; I applied the same changes as in `source_detail.html`.


dchiller (approved):
Looks good to me -- only suggestion is to add a link to the PR where the other changes were made since you reference them ("the same changes as")

PR #1344, opened 2024-03-05
Add link to new "Basic Principles" flat page in "About" menu

dchiller:
Debra created a new flatpage at https://cantusdatabase.org/about/principles_of_indexing/. This PR adds the flat page to the "About" dropdown.

**Screenshot with new menu item in "About" dropdown**

image


jacobdgm (approved):
Screenshots would have been useful here, but this change looks fine. Thanks!

jacobdgm:
thinking a tiny bit more, should the URL for this flatpage not be `principles-of-indexing`, with dashes rather than underscores? We'll have to make this change in the database as well, but this would bring the URL's formatting in line with `content-statistics` and `id-numbers` and so on.

dchiller:
> thinking a tiny bit more, should the URL for this flatpage not be `principles-of-indexing`, with dashes rather than underscores?

Good point...I'll make the change and alert Debra...

Screenshots incoming.

PR #1356, opened 2024-03-12
Sequence Detail: Wrap concordances table in table-responsive div

### LINKED ISSUE #1355: Sequence Detail Page: Concordances table lacking `class="table-responsive"` div
Screenshot 2024-03-12 at 10 31 57 AM

jacobdgm:
Fixes #1355.


dchiller (approved):
Would suggest an "after" screenshot, but looks good!

jacobdgm:
> Would suggest an "after" screenshot, but looks good!

I know, I know. It was a lower-priority issue that I was doing my best to fix in under 5 minutes. Thanks for the review :)

PR #1362, opened 2024-03-14
Fix "Browse Chants" page "Source" drop-down bug

### LINKED ISSUE #717: When selecting a folio of a manuscript, "Source" name is briefly wrong
Steps:
1. Select a source, either from the "Browse sources" dropdown menu on the home page or the "Cantus Sources" page;
2. Select any folio from the dropdown menu;
3. When the page loads, the "Source" will briefly be A-Gu 1584, before switching to the source it actually is.
4. This will keep happening every time you select a new folio. It will also happen if you use the little folio navigation arrow thingies on the right of the dropdown menu.

It would seem that regardless of the source, A-Gu 1584 shows up first. The longest it took to change was 3 seconds.

Demo:

https://github.com/DDMAL/CantusDB/assets/83373378/9c1eb272-5a3f-43ad-8bf8-c7fd85a7abd6

### LINKED ISSUE #1336: Browse Chants page: we must fix "Source" dropdown behavior
Browse chants has a dropdown of shelfmarks that always shows A-Gu 29; seems like this should show the current source...
e.g. https://cantusdatabase.org/source/714555/chants/

image

lucasmarchd01:
This PR addresses the bug that was found on the chant-list page.

When the user would select a different source from the drop-down menu, instead of being redirected to 
`source/<int:source_id>/chants/` they get sent to `/source/<int:source_id>/chants/?folio=&feast=&search_text=&genre=&source=<int:new_source_id>`

There was some discussion in #1336 about whether this functionality is useful to users on the site. Perhaps the best option is to remove it altogether, but in the meantime it should work as expected for users that might be browsing this way.

The code modified includes the chant_list.js file which handles the filters. The page will now redirect correctly and the source parameter is no longer used in the URL. Additionally, the current source that the user is browsing is correctly displayed as the selected option in the drop-down menu.

Example after selecting a new source from the drop-down menu:

Screenshot 2024-03-14 at 5 38 39 PM

Screenshot 2024-03-14 at 5 38 26 PM

Fixes #1336, Fixes #717


django/cantusdb_project/static/js/chant_list.js
```diff
@@ -41,14 +41,15 @@ window.addEventListener("load", function () {
...
-    // if the source is reset, reset every selector
+    // if the source is reset, redirect to source/<int:source_id>/chants/
     function setSource() {
-        url.searchParams.set('folio', '');
-        url.searchParams.set('feast', '');
-        url.searchParams.set('search_text', '');
-        url.searchParams.set('genre', '');
         const source = sourceFilter.options[sourceFilter.selectedIndex].value;
-        url.searchParams.set('source', source);
+        url.pathname = "/source/" + source + "/chants/";
```
jacobdgm: Can this be changed to `const sourceId = ...`?

jacobdgm (approved):
I suggested one small tweak to a variable name, but on the whole, this looks very good. Thanks!

PR #1364, opened 2024-03-15
Deploy drastic header streamlining to Staging Server

jacobdgm:
This PR is parallel to #1338; merging that same branch into a separate branch so that we don't mess up the staging branch itself in the situation where we need to revert these changes.


jacobdgm:
Commit 5e54e918afff6aa35de9f961ba1cf29c6dbfe1a6 is present in the `staging-header-drastic` branch, but looking at this PR's diff, the changes in that PR are not present. I can't figure out why this should be the case.

I don't think we'll make much headway with Debra if she's not able to access the new flatpage she created last week.

jacobdgm:
> Commit [5e54e91](https://github.com/DDMAL/CantusDB/commit/5e54e918afff6aa35de9f961ba1cf29c6dbfe1a6) is present in the `staging-header-drastic` branch, but looking at this PR's diff, the changes in that PR are not present. I can't figure out why this should be the case.
> 
> I don't think we'll make much headway with Debra if she's not able to access the new flatpage she created last week.

Ah - pulling the recent changes from `develop` seems to have fixed it.

There are a bunch of commits now that run ahead of the current state of the `staging` branch, but that's fine - nothing that goes on in `staging-header-drastic` will have a direct effect on the `staging` branch. Going to go ahead and merge and then deploy to staging.

PR #1377, opened 2024-03-19
Hide unpublished sources in Source dropdown for anonymous users

### LINKED ISSUE #1376: Browse Chants page - unpublished sources appear in list of sources
https://staging.cantusdatabase.org/source/702611/chants/
Screenshot 2024-03-19 at 3 01 48 PM

jacobdgm:
The main purpose of this PR is to fix #1376. In order to do so, in the view responsible for the Browse Chants page, we now add a check whether the requesting user is authenticated. If they aren't, we hide any unpublished sources when querying for the list of sources.

As I was working on this, I made several other changes:
- Added type annotations to the responsible view
- Since it's really clear that this is a Source-related view more than a Chant-related view, I moved the view from `views.chant` to `views.source`
- It has taken me a while to fully internalize that the Chant List page is the one that lists all the chants in a source, rather than the one that allows you to search through all the chants in all the sources. Since the relevant template has "Browse Chants" as its largest header, I went ahead and renamed this view from `ChantListView` to `SourceBrowseChantsView`, which should make the connection a bit more apparent and head off potential misunderstandings.
  - Along the same lines, in `urls.py`, I renamed `"chant-list"` to `"browse-chants"`
  - and `ChantListViewTest` is now `SourceBrowseChantsTest`

These last few changes affected many files. You might want to step through this PR commit-by-commit in reviewing it, @dchiller and/or @lucasmarchd01


django/cantusdb_project/main_app/urls.py
```diff
@@ -283,8 +283,8 @@
     ),
     path(
         "source/<int:source_id>/chants/",
-        ChantListView.as_view(),
-        name="chant-list",
+        SourceBrowseChantsView.as_view(),
+        name="browse-chants",
```
lucasmarchd01, line 287: in `urls.py` we also have the `redirect-chant-list` path that could be renamed to `redirect-browse-chants`

django/cantusdb_project/main_app/templates/chant_list.html
lucasmarchd01, line 1: Maybe we should also reconsider naming this file to something like `browse_chants.html` or `chant_browse.html`.

lucasmarchd01 (changes requested):

lucasmarchd01:
Good catch, thank you for the changes!

django/cantusdb_project/main_app/urls.py
jacobdgm: you're right! I should be able to fix this first thing on Thursday. Thanks!

django/cantusdb_project/main_app/templates/chant_list.html
jacobdgm: it should either start with "source" - `source_browse_chants.html`, or just be `browse_chants.html`. Thanks for catching this!

django/cantusdb_project/main_app/urls.py
jacobdgm: ...actually, isn't the old path, the one we're redirecting from, "chant-list"? if that's the case, I think `redirect-chant-list` remains the more accurate name.

But I should check in on this to make sure we don't have any out-of-date comment, at least. Again, thanks - this is a worthwhile thought!

lucasmarchd01: The old path is redirecting from `/chants/?source=<int:source_id>`. Anything with `/chant-list` brings a 404 so I don't believe this was a path that we set up a redirect for or that ever existed.

dchiller (commented):
You've got some unused imports now in views.chant:

`django.db.models.Value`
`main_app.models.Segment`

django/cantusdb_project/main_app/urls.py
jacobdgm: I renamed `redirect_chant_list` to `redirect_chants` - it's not super descriptive, but the path was literally `/chants/`, so I feel it's accurate at least.

django/cantusdb_project/main_app/urls.py
```diff
@@ -190,8 +190,8 @@
     ),
     path(
         "chants/",
-        views.redirect_chant_list,
-        name="redirect-chant-list",
+        views.redirect_chants,
+        name="redirect-chants",
```
dchiller, line 194: Do these tests still pass?

It looks like we use `reverse("redirect-chant-list")`...

https://github.com/DDMAL/CantusDB/blob/fe15edc7106369eef622fe7df5e585a95bf6cfb7/django/cantusdb_project/main_app/tests/test_views.py#L710-L727

dchiller: A quick text search of the rest of the repo looks like we don't use the old name anywhere else...does that make sense to you?

jacobdgm: Thanks for catching this!

dchiller (approved):

PR #1384, opened 2024-03-21
Refactor loading of environment variables

### LINKED ISSUE #1383: Rewrite Config Block
https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/cantusdb/settings.py#L43-L52

The use of single `if` statements on the same variable is not correct, and it does not provide a suitable (and safe) fallback if the variable does not match. Also if no environment variable is set, the `getenv` will try to split on `None` which will make it cranky. 

A couple suggestions:

The "minimal":

```py
if PROJECT_ENVIRONMENT == "STAGING":
    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_STAGING").split(" ")
    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_STAGING").split(" ")
elif PROJECT_ENVIRONMENT == "PRODUCTION":
    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_PRODUCTION").split(" ")
    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_PRODUCTION").split(" ")
else:
    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_DEVELOPMENT").split(" ")
    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_DEVELOPMENT").split(" ")
    DEBUG = True
```

The "rewrite":

```py
allowed_hosts = os.getenv(f"CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT}")
trusted_origins = os.getenv(f"CANTUSDB_ORIGINS_{PROJECT_ENVIRONMENT}")

ALLOWED_HOSTS = allowed_hosts.split(" ") if allowed_hosts else []
CSRF_TRUSTED_ORIGINS = trusted_origins.split(" ") if trusted_origins else []

if PROJECT_ENVIRONMENT not in ("STAGING", "PRODUCTION"):
    DEBUG = True
```

jacobdgm:
This PR reworks how environment variables are read:

- In `settings.py`, we've gotten rid of the series of simple `if` statements checking the value of `PROJECT_ENVIRONMENT`, replacing them with an `if` and a series of `elses`
- in `settings.py`, if `PROJECT_ENVIRONMENT` is not set to something reasonable, an exception with an informative error message is raised.
  - a similar error message used to be generated in `context_processors.py`, but it makes more sense that it be generated in `settings.py`, so it's been removed from the context processors file.
- `context_processors.py` now reads `PROJECT_ENVIRONMENT` from the project settings (as set in `settings.py`), rather than reading it from the environment variables.
- type annotations have been added throughout `settings.py`

In doing the above, it fixes #1383.

It was suggested that in case of misconfiguration, a safe fallback be specified. In talking about this with @dchiller, we both agreed that a misconfiguration should not pass silently, hence the exception that remains.

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -38,36 +38,25 @@ SECRET_KEY = os.getenv("CANTUSDB_SECRET_KEY")
 
 PROJECT_ENVIRONMENT: Optional[str] = os.getenv("PROJECT_ENVIRONMENT")
 
-if PROJECT_ENVIRONMENT not in ("DEVELOPMENT", "STAGING", "PRODUCTION"):
+# DEBUG gets set to True below when PROJECT_ENVIRONMENT == "DEVELOPMENT"
+# SECURITY WARNING: don't run with debug turned on in production!
+DEBUG: bool = False
+if PROJECT_ENVIRONMENT == "DEVELOPMENT":
+    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_DEVELOPMENT").split(" ")
+    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_DEVELOPMENT").split(" ")
+    DEBUG = True
+elif PROJECT_ENVIRONMENT == "STAGING":
+    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_STAGING").split(" ")
+    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_STAGING").split(" ")
+elif PROJECT_ENVIRONMENT == "PRODUCTION":
+    ALLOWED_HOSTS = os.getenv("CANTUSDB_HOSTS_PRODUCTION").split(" ")
+    CSRF_TRUSTED_ORIGINS = os.getenv("CANTUSDB_ORIGINS_PRODUCTION").split(" ")
+else:
     raise ValueError(
         "The PROJECT_ENVIRONMENT environment variable must be either "
         "DEVELOPMENT, STAGING, or PRODUCTION. "
         f"Its current value is {PROJECT_ENVIRONMENT}."
     )
-
-cantusdb_hosts: Optional[str] = os.getenv(f"CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT}")
-cantusdb_origins: Optional[str] = os.getenv(f"CANTUSDB_ORIGINS_{PROJECT_ENVIRONMENT}")
-
-try:
-    ALLOWED_HOSTS: list = cantusdb_hosts.split(" ")
-except AttributeError:  # cantusdb_hosts is None
-    raise ValueError(
-        f"The PROJECT_ENVIRONMENT environment variable is set to {PROJECT_ENVIRONMENT}, "
-        f"yet the CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT} environment variable is not set."
-    )
-
-try:
-    CSRF_TRUSTED_ORIGINS: list = cantusdb_origins.split(" ")
-except AttributeError:  # cantusdb_origins is None
-    raise ValueError(
-        f"The PROJECT_ENVIRONMENT environment variable is set to {PROJECT_ENVIRONMENT}, "
-        f"yet the CANTUSDB_ORIGINS_{PROJECT_ENVIRONMENT} environment variable is not set."
-    )
-
-# SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = False
-if PROJECT_ENVIRONMENT == "DEVELOPMENT":
-    DEBUG = True
 
 
 # Application definition
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -25,7 +25,7 @@ MESSAGE_TAGS: dict = {
 }
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR: Optional[str] = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR: str = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 STATIC_ROOT: Optional[str] = os.getenv("CANTUSDB_STATIC_ROOT")
 MEDIA_ROOT: Optional[str] = os.getenv("CANTUSDB_MEDIA_ROOT")
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -16,7 +16,7 @@ from django.contrib.messages import constants as messages
 from typing import Optional
 
 # https://ordinarycoders.com/blog/article/django-messages-framework
-MESSAGE_TAGS = {
+MESSAGE_TAGS: dict = {
     messages.DEBUG: "alert-secondary",
     messages.INFO: "alert-info",
     messages.SUCCESS: "alert-success",
@@ -25,16 +25,16 @@ MESSAGE_TAGS = {
 }
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR: Optional[str] = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
-STATIC_ROOT = os.getenv("CANTUSDB_STATIC_ROOT")
-MEDIA_ROOT = os.getenv("CANTUSDB_MEDIA_ROOT")
+STATIC_ROOT: Optional[str] = os.getenv("CANTUSDB_STATIC_ROOT")
+MEDIA_ROOT: Optional[str] = os.getenv("CANTUSDB_MEDIA_ROOT")
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/3.0/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = os.getenv("CANTUSDB_SECRET_KEY")
+SECRET_KEY: Optional[str] = os.getenv("CANTUSDB_SECRET_KEY")
 
 PROJECT_ENVIRONMENT: Optional[str] = os.getenv("PROJECT_ENVIRONMENT")
 
@@ -61,7 +61,7 @@ else:
 
 # Application definition
 
-INSTALLED_APPS = [
+INSTALLED_APPS: list[str] = [
     "dal",
     "dal_select2",
     "django.contrib.admin",
@@ -82,7 +82,7 @@ INSTALLED_APPS = [
     "users",
 ]
 
-MIDDLEWARE = [
+MIDDLEWARE: list[str] = [
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
@@ -94,9 +94,9 @@ MIDDLEWARE = [
     "reversion.middleware.RevisionMiddleware",
 ]
 
-ROOT_URLCONF = "cantusdb.urls"
+ROOT_URLCONF: str = "cantusdb.urls"
 
-TEMPLATES = [
+TEMPLATES: list[dict] = [
     {
         "BACKEND": "django.template.backends.django.DjangoTemplates",
         "DIRS": [os.path.join(BASE_DIR, "templates")],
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/cantusdb/settings.py
```diff
@@ -35,21 +36,38 @@
...
+        "The PROJECT_ENVIRONMENT environment variable must be either "
+        "DEVELOPMENT, STAGING, or PRODUCTION. "
+        f"Its current value is {PROJECT_ENVIRONMENT}."
+    )
+
+cantusdb_hosts: Optional[str] = os.getenv(f"CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT}")
+cantusdb_origins: Optional[str] = os.getenv(f"CANTUSDB_ORIGINS_{PROJECT_ENVIRONMENT}")
+
+try:
+    ALLOWED_HOSTS: list = cantusdb_hosts.split(" ")
```
ahankinson: You’re effectively checking the same condition 3 times.

django/cantusdb_project/cantusdb/settings.py
```diff
@@ -35,21 +36,38 @@
...
+except AttributeError:  # cantusdb_origins is None
+    raise ValueError(
+        f"The PROJECT_ENVIRONMENT environment variable is set to {PROJECT_ENVIRONMENT}, "
+        f"yet the CANTUSDB_ORIGINS_{PROJECT_ENVIRONMENT} environment variable is not set."
+    )
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = False  # this is switched to True below when PROJECT_ENVIRONMENT=="DEVELOPMENT"
-
+DEBUG = False
```
ahankinson: Where do you ever set this to True now?

django/cantusdb_project/main_app/context_processors.py
```diff
@@ -2,17 +2,9 @@
...
-        project_environment == "DEVELOPMENT"
-        or project_environment == "STAGING"
-        or project_environment == "PRODUCTION"
-    ):
-        raise ValueError(
-            "The PROJECT_ENVIRONMENT environment variable must be either "
-            "DEVELOPMENT, STAGING, or PRODUCTION. "
-            f"Its current value is {project_environment}."
-        )
+    project_environment: str = os.getenv("PROJECT_ENVIRONMENT")
```
ahankinson: Don’t read the environmental variables outside of your settings file. Get the value from there, as I showed in my comment on the issue.

django/cantusdb_project/main_app/tests/test_models.py
```diff
@@ -90,7 +90,7 @@ def test_index_components(self):
         self.assertEqual(weight_search_term_dict, expected_dict)
 
     def test_get_concordances(self):
-        chant = Chant.objects.get(id=1)
+        chant = Chant.objects.first()
```
ahankinson, line 93: Not part of this pr?

django/cantusdb_project/cantusdb/settings.py
dchiller: Line 70

ahankinson: Right. Got lost in the diff

django/cantusdb_project/cantusdb/settings.py
ahankinson: If you really want to do this (I think it's overkill -- there's a such thing as being too cautious, particularly because you're automating the deployment, but anyway...) then you should just check if it's null. This will anyway only catch the instances where you're trying to `split` on `None`. 

So:

```py
if cantusdb_hosts is None:
    raise ValueError(
        f"The PROJECT_ENVIRONMENT environment variable is set to {PROJECT_ENVIRONMENT}, "
        f"yet the CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT} environment variable is not set."
    )

ALLOWED_HOSTS = cantusdb_hosts.split(" ")
```

django/cantusdb_project/main_app/tests/test_models.py
jacobdgm: I'm not sure why this particular test chose now to begin failing, but it did. I don't know why it had been set up differently from other tests in the test suite, but bringing it into alignment with the others causes it to behave as expected.

I can revert this and open a separate PR if you think it's worth doing.

django/cantusdb_project/cantusdb/settings.py
jacobdgm: > You’re effectively checking the same condition 3 times.

I don't understand what you mean by this. Line 41, we check that `PROJECT_ENVIRONMENT` is properly set. Line 51, we check that the relevant `CANTUSDB_HOSTS_` variable is properly set, and analogously Line 54 and `CANTUSDB_ORIGINS_`.

> If you really want to do this (I think it's overkill -- there's a such thing as being too cautious, particularly because you're automating the deployment, but anyway...)

Fair enough. But if we expect configuration errors to be caught by Ansible, then why should we bother handling a situation in our Python that will never arise? If we decide to prepare for this case that will never arise, isn't it worth setting something up something to aid future developers in debugging, rather than setting a safe fallback value which could obscure the cause of a problem?

Maybe your argument turns on some detail I'm missing, but I don't understand why we wouldn't want to have error messages here.

jacobdgm:
```
if cantusdb_hosts is None:
    raise ValueError(
        f"The PROJECT_ENVIRONMENT environment variable is set to {PROJECT_ENVIRONMENT}, "
        f"yet the CANTUSDB_HOSTS_{PROJECT_ENVIRONMENT} environment variable is not set."
    )

ALLOWED_HOSTS = cantusdb_hosts.split(" ")
```

You have more experience in this than I do, but my understanding was that in Python, doing control flow via exception handling is preferred over checking the value of a variable before running a command. Is there a reason why the former is better in this particular situation?

django/cantusdb_project/main_app/tests/test_models.py
ahankinson: Nah, just pointing it out in case it slipped in.

django/cantusdb_project/cantusdb/settings.py
ahankinson: EAFP style has its place, but this probably isn't it. You know exactly what the types of `cantusdb_hosts` will be -- either `None` or a string. There's no need for a try/except if you know how to otherwise handle it. (Can you imagine trying to code where every `if` statement is a try/except? That would be a nightmare!)

(Part of me thinks EAFP style was useful back when Python didn't have any real concept of first-class types, so you could never really be sure what the type for any given variable was, and rather than `if/elif/elif/else` check all possible types -- so just do something and catch the failure. But now we have type annotations, so you know exactly what it will be, and the type checker will tell you if you're not doing it right. So no need to do a try/except block.)

> But if we expect configuration errors to be caught by Ansible, then why should we bother handling a situation in our Python that will never arise?  If we decide to prepare for this case that will never arise, isn't it worth setting something up something to aid future developers in debugging, rather than setting a safe fallback value which could obscure the cause of a problem?

Indeed, that is the question. You're not coding an application for broad distribution, you have an install base of < 5, and you also control the end-to-end environment where these things get deployed. Automating the values in your settings files for staging / production and setting sane defaults for local development setups is the way I go.

Your "safe fallback" is the development environment. You can monkey around with those files all you like. But when it comes to a deployment, the rule is you *never* touch the config files by hand (Unless you're debugging a problem). If you need to make a change, make it in Ansible and deploy it. Want to provision a new server? Point Ansible at it and go. You should never get to a point where you're hand-deploying a server, and thus you'll never get to a point where these variables are not set. 

Your "future developers" will then just need to look at the Ansible deployment to see how the file is templated for staging / production.

jacobdgm: > (Can you imagine trying to code where every `if` statement is a try/except? That would be a nightmare!)

Sounds like [Hurl](https://ntietz.com/blog/introducing-hurl/) to me. Which is maybe a nightmare?

jacobdgm: Okay, I still don't feel I understand 100%, I feel I'm getting closer. Let me rework a few things and see if it's more along the lines of what you have in mind.

django/cantusdb_project/cantusdb/settings.py
```diff
@@ -24,37 +25,43 @@
 }
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+BASE_DIR: Optional[str] = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
ahankinson: ? os.path.dirname always returns a str, AFAIK

ahankinson:
I wrote the stuff below first. I'll keep it there, because I think there might be something useful in it, but upon further reflection -- whatever man. If it works for you, then that's what matters. Feel free to merge. 

------

I'm going to put the blame on Docker for the confusion here.  (Have I mentioned how much I LOOOOVE Docker... It doesn't get in the way of anything at all.)

You're assuming(?) that I meant that the **environment variables** would always be set, so you reverted back to the old way. I was assuming that Ansible would template out the `settings.py` directly, and bypass any need to read from the environment variables at all. 

This means that you were maybe assuming that I meant to revert back to `os.getenv(BLAH).split(" ")` which, if `BLAH` isn't set, will blow up. Maybe that's what you thought I meant when I said you should assume an environment where they're always set, but that's not what I meant. I still think calling `split()` on something that you know can be `None` is wrong. I was simply saying that, since you know this, you can check and act, rather than either a) letting it blow up or b) going through all the effort of writing a try / catch. 

Beyond that, my larger point was that, if something else is writing exactly what it needs to the settings file directly, you don't need to muck around with "if PROJECT_ENVIRONMENT == BLAH then ..." in your settings file ***at all***. You don't have to deal with all the cases, you just have to set it to the values that are needed **for that environment**. 

In Ansible-land, this looks like a template like this (excerpt):

```
PROJECT_ENVIRONMENT: Optional[str] = {{ cantus_deployment_env }}
DEBUG: bool = {{ True if cantus_deployment_env == DEVELOPMENT else False }}
ALLOWED_HOSTS = {{ cantus_allowed_hosts }}
CSRF_TRUSTED_ORIGINS = {{ cantus_trusted_origins }}
```

That's it. Each of those `{{ statements }}` gets templated out to the appropriate values when you deploy, based on the environment you're deploying to. No need for if / then / else /try / except / throw / catch / fail / getenv / split / yadda yadda yadda or whatever else. 

For local development, you can have your own settings.py that has sensible development values, with the understanding that this doesn't need to ever translate to the deployment. 

```
PROJECT_ENVIRONMENT: Optional[str] = "DEVELOPMENT"
DEBUG: bool = True
ALLOWED_HOSTS = []
CSRF_TRUSTED_ORIGINS = []
```

For DIAMM I have two settings files. `settings_local.py` which contains the local environment settings, and `settings.py` which imports `settings_local`. I can keep all the things in `settings.py` that apply to the whole app, and only template out a `settings_local.py` with the appropriate values. See [here](https://github.com/diamm/diamm/blob/main/diamm/settings_local.py.template) if you're at all interested... I add `settings_local.py` to my .gitignore.

The only fly in the soup, so to speak, is that your settings.py is "locked" inside your docker container. But it seems to me that somewhere about [here](https://github.com/dact-chant/ansible.cantus-db/blob/main/roles/cantusdb-app/tasks/main.yml#L58) you (or @dchiller) can template out a deployment settings.py file in your docker container just before you build it.

dchiller:
Two overall comments:

1. I find the `Optional[some_type]` annotations confusing. (To be honest, all of the annotations are in some ways -- I can't see a case where you would set these again in any other place -- in fact, we probably shouldn't?, so I'm not sure why `SOME_SETTING: str = "some string here"` isn't totally superfluous. But leaving this to the side because I don't think it really matters either way).  Are any of these actually optional? Can I actually set, for example, `SECRET_KEY` to `None`? 

2. I think we should approve and merge #1386 first. We have an ansible set-up and whether we want to template additional pieces of this with ansible (per @ahankinson's suggestion above) or we want to make any changes to the `settings.py` file in this repo a la this PR, I think enough changes with ansible-based configuration and deployment that it's worth adding that in and then determining what problems we are still trying to fix. I think the concreteness of that will help.

ahankinson:
+1 to merging #1386 first.

jacobdgm:
Okay, I think I'm coming to understand:

- since the project has been Dockerized since I began working on it, I've only seen this approach of setting the values of project variables within `settings.py`, i.e., by specifying them in `Dockerfile`s/`dockercompose.yml`s, and then reading them within `settings.py` using `os.getenv`. I've been trying to refactor towards the local maximum of minimizing the number of commands that need to be run, i.e., simplifying the set of modifications that need to be made to files, in order to set the project environment.
- you're suggesting, however, that we switch to specifying the development environment within `settings.py` itself. And this approach will become even more preferable when we have Ansible set up, which can easily make this change and allow us to bypass environment variables entirely.

If this is all correct, then I agree - let's get #1386 merged, then revisit this. It will probably make most sense to simply close this PR, to avoid several rounds of commits going back and forth ending up in the git history. But let's merge #1386 and then come back to this.

Thanks for taking the time to help us (i.e., me) understand!

ahankinson:
Is this still on hold?

jacobdgm:
I'm closing this, for two reasons:

- 1. There are now merge conflicts, so in the interests of preserving a clean git history, it would be best to implement these couple of changes in a new branch forked from the `develop` branch in its current state.
- 2. I've been instructed to prioritize other issues in my last couple of weeks working in the lab. Especially if done by consulting this PR and the discussion above, this should not be a difficult thing for someone else to implement in the future.

ahankinson:
See #1412 for the fix.

jacobdgm:
Ah, I had forgotten that the context processors was where this all started. Thanks, @ahankinson.

PR #1392, opened 2024-03-29
Remove ajax_melody_view, ajax_concordance_list views

### LINKED ISSUE #1378: We have an `ajax_concordance_list` function-based view we aren't using
https://github.com/DDMAL/CantusDB/blob/9057bdd71b5e7f73c0cc6e083bdaea4017edbf87/django/cantusdb_project/main_app/views/views.py#L72-L124

I initially thought we might be able to use it for #1171, but we'll be getting this from CI directly. Thus, we should simply get rid of this code - it's in the Git history if we need it, but as long as it remains, its main effect will be to confuse future developers.

jacobdgm:
This PR removes two unused views in `main_app/views/views.py`:
- `ajax_concordance_list` is no longer required, since we can get this data directly from https://cantusindex.uwaterloo.ca/json-cid/{CantusID}
  - in doing so, it fixes #1378
- `ajax_melody_view` is no longer required, since we can get this data directly from https://cantusindex.uwaterloo.ca/json-cid-mel/{CantusID}

This PR also removes tests for `ajax_melody_view` (turns out we didn't have tests for `ajax_concordance_list` - yikes!), and reformats the files that were touched to comply with the updated Black style (see #1387)

We haven't been using either of these views, so it's good that they can finally be cleaned up!


dchiller:
Does this function use the `ajax-melody-view`? https://github.com/DDMAL/CantusDB/blob/491968b533ff5812cba6a0e74688858025281d1a/django/cantusdb_project/static/js/chant_detail.js#L34

It looks to me like it constructs a URL to that view...but perhaps this function is never called? If so, I would get rid of the function too at the same time...

jacobdgm:
Ah. We still do need `ajax-melody-view` - thanks for catching this. And also - oof, you should not have needed to catch this, and I should have searched the codebase before asserting that the view is no longer required.

It would be nice to have integration tests set up, since our current test suite doesn't test any of our JavaScript.

Going to close this PR and open another that only removes `ajax_concordance_list`.

PR #1395, opened 2024-04-02
Update csv-export view to use chants' sources' sigla

### LINKED ISSUE #1365: Contruct source-level csv export with source `siglum` rather than chant `siglum`
Brought to our attention by the fact that the csv export for `D-WI1 2` (https://cantusdatabase.org/source/588308) contains chants with two different sigla, one correct and one incorrect. This seems to be because we are using the chant's `siglum` field in the construction of this csv, which may contain legacy data, rather than the source's `siglum` field, which was corrected.

jacobdgm:
This PR updates our `csv-export` view, ensuring that we display chants' sources' sigla rather than the chants' sigla (which are legacy data which is sometimes out-of-date, and should never be displayed). This PR thus fixes #1365.

As part of this PR,
- CsvExportTest was updated to check for this change.
- I reworked CsvExportTest a bit, mostly adding subtests to annotate the various `assert`ions we were making.

### DIFF
--- a/django/cantusdb_project/main_app/views/views.py
+++ b/django/cantusdb_project/main_app/views/views.py
@@ -236,6 +236,7 @@ def csv_export(request, source_id):
         ]
     )
     for entry in entries:
+        siglum = entry.source.siglum if entry.source else ""
         feast = entry.feast.name if entry.feast else ""
         office = entry.office.name if entry.office else ""
         genre = entry.genre.name if entry.genre else ""
@@ -243,7 +244,7 @@ def csv_export(request, source_id):
 
         writer.writerow(
             [
-                entry.siglum,
+                siglum,
                 entry.marginalia,
                 entry.folio,
                 # if entry has a c_sequence, it's a Chant. If it doesn't, it's a Sequence, so write its s_sequence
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/views/views.py
```diff
@@ -236,14 +236,15 @@ def csv_export(request, source_id):
         ]
     )
     for entry in entries:
```
dchiller, line 239: Do you need to include this in the loop? Because we are using the source siglum, it will always be the same for each entry? 

I am just wondering if this is a case that will introduce a lot of duplicate queries?

jacobdgm: Good call

jacobdgm: @lucasmarchd01 - I see we're setting `diff_db = entry.diff_db.id if entry.diff_db else ""` near here, but then we're writing `entry.diff_db` rather than the value of the `diff_db` variable. I can't recall the specifics, but weren't you trying to fix something lately related to differentiae and the CSV export view? Might this be relevant?

dchiller: Hmm...interesting. Approving cause the bit about the siglum looks good to me, but this certainly seems like something to investigate.

dchiller (approved):

PR #1396, opened 2024-04-02
Pre-populate office selector on Chant Create page

### LINKED ISSUE #915: Chant Create page: we could auto-populate the Office selector
In an email forwarded to us by Debra, a CantusDB contributor commented:

> Office/Mass box. When adding chants, the selection made here does not carry over from the previous chant. My preference is that it pre-populate from the previous chant in the source. (But I don't remember whether this used to pre-populate from the previous or not.)

jacobdgm:
This PR fixes #915, pre-populating the office selector with the same office as the most recently created chant within the relevant source.

It adds a new test, `ChantCreateViewTest.test_initial_values`, with a subtest for the existing Feast pre-population feature, and a subtest for the new Office pre-population feature.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -2910,30 +2910,50 @@ class ChantCreateViewTest(TestCase):
         self.assertEqual(chant_2.volpiano_intervals, "1-12-23-34-45-56-67-78-8")
 
     def test_initial_values(self):
+        # create a chant with a known folio, feast, office, c_sequence and image_link
         source: Source = make_fake_source()
+        folio: str = "001r"
+        sequence: int = 1
         feast: Feast = make_fake_feast()
         office: Office = make_fake_office()
-        # create a chant with a known feast and office
+        image_link: str = "https://www.youtube.com/watch?v=9bZkp7q19f0"
         self.client.post(
             reverse("chant-create", args=[source.id]),
             {
-                "manuscript_full_text_std_spelling": "this is a bog standard manuscript spelling textful",
-                "folio": "001r",
-                "c_sequence": "1",
+                "manuscript_full_text_std_spelling": "this is a bog standard manuscript textful spelling",
+                "folio": folio,
+                "c_sequence": str(sequence),
                 "feast": feast.id,
                 "office": office.id,
+                "image_link": image_link,
             },
         )
-        # when we request the page, the same feast and office should be preselected
-        request = self.client.get(
+
+        # when we request the Chant Create page, the same folio, feast, office and image_link should
+        # be preselected, and c_sequence should be incremented by 1.
+        response = self.client.get(
             reverse("chant-create", args=[source.id]),
         )
-        observed_initial_feast: int = request.context["form"].initial["feast"]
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -805,13 +805,15 @@ def get_initial(self):
             }
         latest_folio = latest_chant.folio if latest_chant.folio else "001r"
```
dchiller, line 806: I'm wondering if it would make more sense to used `date_created` as the way to find the "latest chant" rather than `date_updated`. 

Let's say I create a chant, then realize a chant I created 5 folios back needs an edit, make that edit, and then come back to creating chants. Then, `latest_chant` won't be what I want. But if you went by `date_created` then you would get the "right" chant.

Maybe there is a case where `date_updated` leads to the "correct" result more often? 

Of course, in the case I mentioned one can just make the edits manually, but this seem like a simple change that might make more sense?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2909,6 +2909,32 @@ def test_volpiano_signal(self):
...
+        # when we request the page, the same feast and office should be preselected
+        request = self.client.get(
+            reverse("chant-create", args=[source.id]),
+        )
+        observed_initial_feast: int = request.context["form"].initial["feast"]
+        observed_intitial_office: int = request.context["form"].initial["office"]
+        with self.subTest(subtest="test initial value of feast feild"):
+            self.assertEqual(observed_initial_feast, feast.id)
+        with self.subTest(subtest="test initial value of office field"):
+            self.assertEqual(observed_intitial_office, office.id)
```
dchiller: Do you want to test the other initial values too?

django/cantusdb_project/main_app/views/chant.py
annamorphism: I think you're right.  You will encounter some scenarios where date_updated and date_created are really far apart--e.g. if you are adding full texts to some manuscript added in the 90s--but in those cases you probably wouldn't be using chant-create all that often. It's probably fair to assume that if you are creating chants, you want to know about the last chants that were created in the source. 
(Actually, in cases where a whole group is working on a MS at once, you might want to know about the last chants _you_ created, as opposed to your teammates. But that seems a lot harder!)

jacobdgm: While I agree, I think that changing the behavior of the feast dropdown pre-populator is probably beyond the scope of this PR.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: good point. done!

django/cantusdb_project/main_app/views/chant.py
dchiller: Looking back at the original issue, I'm now interpreting the request as neither of these options: pre-populating the `Office` field with the previous chant in the source (in other words, determined by folio and sequence rather than by when updated or created).

I imagine that for the most common workflow (someone entering all the chants on a folio in sequence) these will all be equivalent. And the code looks good, so I'll approve.

dchiller (approved):

PR #1398, opened 2024-04-04
Merge staging into production, 2024 Apr 04 edition

jacobdgm:
Changes:

- Dependencies:
  - black bumped from 21.10b0 to 24.3.0
  - typing-extensions bumped from 3.10.0.0 to 4.0.1
  - click bumped from 7.1.2 to 8.0.0
- Tests
  - `main_app.tests.test_views.CsvExportTest` updated to use subtests
  - `main_app.tests.test_views.ChantCreateViewTest` had several tests added to ensure the `get_initial` method works as expected
- Ajax Concordance List - view removed
- [x] Chant Create
  - we now pre-populate the form's "Office" field using the same office as the most recently modified chant in the same source (should work the same way as the "Feast" field)
- [x] CSV Export
  - We now display the source's siglum in the "siglum" column, rather than each chant's individual siglum

### DIFF
--- a/django/cantusdb_project/cantusdb/urls.py
+++ b/django/cantusdb_project/cantusdb/urls.py
@@ -13,6 +13,7 @@ Including another URLconf
     1. Import the include() function: from django.urls import include, path
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
+
 from django.contrib import admin
 from django.urls import path
 from django.urls import include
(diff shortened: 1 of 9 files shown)


jacobdgm:
There wasn't much to test for this one. @dchiller and/or @lucasmarchd01, can you quickly glance this over to make sure nothing is amiss? With an approval, we should be able to deploy this tomorrow.

dchiller (approved):
Looks good to me!

PR #1407, opened 2024-04-11
Remove Query Parameters from Links to Homepage

### LINKED ISSUE #1399: Header: we should remove query parameters on links to the homepage
This is a really simple fix - in the header, we have two links to the homepage, and each is labelled with a query parameter that shows us which link was clicked on to go to the homepage. We wanted to experimentally test how people were using the site, and [the results of this experiment have come in](https://github.com/DDMAL/CantusDB/pull/1338#issuecomment-2038235671). We can get rid of these query parameters now so as to de-complicate our `base.html` template.

candlecao:
This PR removes query parameters from the 2 links to the home page in the NavBar.  It fixes #1399.


jacobdgm (approved):
Looks good! Thanks for fixing this!

PR #1422, opened 2024-04-19
Chant Create: Reimplement Suggested Chants feature

### LINKED ISSUE #1024: Chant Create suggested chant autofill buttons could be refactored to use a single function
Currently, on the Chant Create page, for the buttons in the sidebar providing suggested texts/genres for chants, we create a JavaScript function named after each suggestion's Cantus ID. Since Cantus IDs can contain `.`s and `:`s, and JS function names cannot, we have to convert Cantus IDs into valid alphanumeric strings for this to work.

A better possibility could be to create a single function into which we can pass a Cantus ID and it would handle all the autofilling. Discussion of how to implement this can be found in #1019.

Thanks, @ahankinson, for this suggestion!

### LINKED ISSUE #1388: Chant Create view: We need to reimplement the "suggested chants" feature
This feature was turned off by #1142, and we need to turn it back on again.

https://github.com/DDMAL/CantusDB/commit/92fa04590885aeb02a8475e547af5a6cf5110ad5 was the commit responsible for removing this functionality. (https://github.com/DDMAL/CantusDB/commit/fc188e94e2581fdb613e9948df1eec4309448196 and https://github.com/DDMAL/CantusDB/commit/d65f4246ab9e2870fec57a9660a3366d1cfd75b3 could also be useful for writing tests for this view)

I think it could be worth exploring implementing this feature client-side - when a user requests this page, we can return a page without sending a request to CI ourself. When the user's page loads, we get the user's browser to send requests to CI for the data necessary for creating this list of chants, and update the content of the page using JavaScript.
- It's less necessary to do this client-side than on publicly-accessible views, since we don't expect bots to be trawling through Chant Create pages.
- But this could allow us to return the page more quickly, allowing the user to start interacting with the page even before CI has returned its responses.

We can use the following endpoints:
- https://cantusindex.uwaterloo.ca/json-nextchants/{CantusID} to get the list of potential successor chants (note that CI's [documentation on this endpoint](https://github.com/dact-chant/cantus-index) indicates it is "under construction" - it seems to be working fine currently, but we should perhaps check in with Jan to confirm it will behave reliably)
- https://cantusindex.uwaterloo.ca/json-cid/{CantusID} to get the fulltext, genre, etc. of each potential successor chants

jacobdgm:
This PR reinstates the Suggested Chants buttons on the Chant Create page. It adds several functions in a new file, `cantusindex.py`. The main purpose of these functions is to request data from Cantus Index and parse it.

As part of writing these functions, I set up a system that allows us to replace `requests.get` calls with a mock function, so that CI doesn't get blitzed with requests whenever someone runs the test suite (and also so our test suite runs more quickly, and also so our tests don't start failing if the state of CI's database changes). These mock objects/functions are used extensively in `test_functions`, and are also used in `test_views.ChantCreateViewTest` wherever we had been requesting the Chant Create page. As we turn on features that request data from Cantus Index in other views, we should remember to patch any calls to `requests.get` in those views as well.

Once I had figured out how to refactor the `autoFillSuggestedChant` function(s) (see #1024), I went ahead and refactored the `autoFillFeast` functions as well.

This PR fixes #1388 and fixes #1024.

### DIFF
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -119,7 +119,7 @@ def get_json_from_ci_api(
     try:
         response: requests.Response = requests.get(uri, timeout=timeout)
     except requests.exceptions.Timeout:
-        return {}
+        return None
 
     response.encoding = "utf-8-sig"
     return response.json()
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -33,7 +33,7 @@ def get_suggested_chants(
     if not isinstance(first_suggestion, dict):
         return None
 
-    sort_by_occurrences: function = lambda suggestion: int(suggestion["count"])
+    sort_by_occurrences: Callable[[dict], int] = lambda suggestion: int(suggestion["count"])
     sorted_suggestions: list = sorted(
         all_suggestions, key=sort_by_occurrences, reverse=True
     )
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -76,11 +76,11 @@ def get_suggested_chant(
         return None
 
     fulltext: str = json["info"]["field_full_text"]
-    incipit = " ".join(fulltext.split(" ")[:5])
+    incipit: str = " ".join(fulltext.split(" ")[:5])
     genre_name: str = json["info"]["field_genre"]
     genre_id: Optional[int] = None
     try:
-        genre_id: Optional[Genre] = Genre.objects.get(name=genre_name).id
+        genre_id = Genre.objects.get(name=genre_name).id
     except Genre.DoesNotExist:
         pass
 
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -9,7 +9,7 @@ from main_app.models import Genre
 
 CANTUS_INDEX_DOMAIN: str = "https://cantusindex.uwaterloo.ca"
 DEFAULT_TIMEOUT: float = 2  # seconds
-NUMBER_OF_SUGGESTED_CHANTS: int = 5  # this number can't be too large,
+NUMBER_OF_SUGGESTED_CHANTS: int = 3  # this number can't be too large,
 # since for each suggested chant, we make a request to Cantus Index.
 # We haven't yet parallelized this process, so setting this number
 # too high will cause the Chant Create page to take a very long time
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -22,8 +22,11 @@ def get_suggested_chants(
 ) -> Optional[list[dict]]:
     endpoint_path: str = f"/json-nextchants/{cantus_id}"
     all_suggestions: Optional[list] = get_json_from_ci_api(endpoint_path)
+
     if not all_suggestions:
-        return None  # get_json_from_ci_api timed out
+        # get_json_from_ci_api timed out
+        # or CI returned a response with no suggestions.
+        return None
 
     # when Cantus ID doesn't exist within CI, CI's api returns a 200 response with `['Cantus ID is not valid']`
     first_suggestion = all_suggestions[0]
@@ -138,4 +141,10 @@ def get_json_from_ci_api(
         return None  # /json-cid/Non-existentCantusId returns a 500 page
 
     response.encoding = "utf-8-sig"
+
+    if not response.text.strip():
+        # /json-nextchants returns a response with text='\r\n' in situations where
+        # there are no suggested chants
+        return None
+
     return response.json()
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -8,9 +8,9 @@ from codecs import decode
 from typing import Optional, Union
 from main_app.models import Genre
 
-CANTUS_INDEX_DOMAIN = "https://cantusindex.uwaterloo.ca"
-DEFAULT_TIMEOUT = 2  # seconds
-NUMBER_OF_SUGGESTED_CHANTS = 5  # this number can't be too large,
+CANTUS_INDEX_DOMAIN: str = "https://cantusindex.uwaterloo.ca"
+DEFAULT_TIMEOUT: float = 2  # seconds
+NUMBER_OF_SUGGESTED_CHANTS: int = 5  # this number can't be too large,
 # since for each suggested chant, we make a request to Cantus Index.
 # We haven't yet parallelized this process, so setting this number
 # too high will cause the Chant Create page to take a very long time
@@ -29,17 +29,21 @@ def get_suggested_chants(
 
     suggested_chants: list[dict] = []
     for suggestion in trimmed_suggestions:
-        cantus_id = suggestion["cid"]
-        occurrences = suggestion["count"]
+        cantus_id: str = suggestion["cid"]
+        occurrences: int = int(suggestion["count"])
         suggested_chants.append(get_suggested_chant(cantus_id, occurrences))
 
     # filter out Cantus IDs where get_suggested_chant timed out
-    filtered_suggestions = [sugg for sugg in suggested_chants if sugg is not None]
+    filtered_suggestions: list[dict] = [
+        sugg for sugg in suggested_chants if sugg is not None
+    ]
 
     return filtered_suggestions
 
 
-def get_suggested_chant(cantus_id: str, occurrences: int) -> Optional[dict]:
+def get_suggested_chant(
+    cantus_id: str, occurrences: int, timeout: float = DEFAULT_TIMEOUT
+) -> Optional[dict]:
     """Given a Cantus ID and a number of occurrences, query one of Cantus Index's
     APIs for information on that Cantus ID and return a dictionary
     containing a full text, an incipit, the ID of that Cantus ID's genre, and
@@ -66,7 +70,7 @@ def get_suggested_chant(cantus_id: str, occurrences: int) -> Optional[dict]:
             ...but if get_json_from_ci_api timed out, returns None instead
     """
     endpoint_path: str = f"/json-cid/{cantus_id}"
-    json: dict = get_json_from_ci_api(endpoint_path)
+    json: dict = get_json_from_ci_api(endpoint_path, timeout=timeout)
 
     if not json:  # mostly, in case of a timeout within get_json_from_ci_api
         return None
@@ -74,8 +78,11 @@ def get_suggested_chant(cantus_id: str, occurrences: int) -> Optional[dict]:
     fulltext: str = json["info"]["field_full_text"]
     incipit = " ".join(fulltext.split(" ")[:5])
     genre_name: str = json["info"]["field_genre"]
-    genre: Genre = Genre.objects.get(name=genre_name)
-    genre_id: int = genre.id
+    genre_id: Optional[int] = None
+    try:
+        genre_id: Optional[Genre] = Genre.objects.get(name=genre_name).id
+    except Genre.DoesNotExist:
+        pass
 
     return {
         "cantus_id": cantus_id,
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -4,7 +4,7 @@ Cantus Index's (CI's) various APIs.
 """
 
 import requests
-from typing import Optional, Union
+from typing import Optional, Union, Callable
 from main_app.models import Genre
 
 CANTUS_INDEX_DOMAIN: str = "https://cantusindex.uwaterloo.ca"
@@ -33,7 +33,9 @@ def get_suggested_chants(
     if not isinstance(first_suggestion, dict):
         return None
 
-    sort_by_occurrences: Callable[[dict], int] = lambda suggestion: int(suggestion["count"])
+    sort_by_occurrences: Callable[[dict], int] = lambda suggestion: int(
+        suggestion["count"]
+    )
     sorted_suggestions: list = sorted(
         all_suggestions, key=sort_by_occurrences, reverse=True
     )
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -25,6 +25,11 @@ def get_suggested_chants(
     if not all_suggestions:
         return None  # get_json_from_ci_api timed out
 
+    # when Cantus ID doesn't exist within CI, CI's api returns a 200 response with `['Cantus ID is not valid']`
+    first_suggestion = all_suggestions[0]
+    if not isinstance(first_suggestion, dict):
+        return None
+
     sort_by_occurrences: function = lambda suggestion: int(suggestion["count"])
     sorted_suggestions: list = sorted(
         all_suggestions, key=sort_by_occurrences, reverse=True
@@ -88,12 +93,16 @@ def get_suggested_chant(
     except Genre.DoesNotExist:
         pass
 
+    clean_cantus_id = cantus_id.replace(".", "d").replace(":", "c")
+    #                                        "d"ot             "c"olon
     return {
         "cantus_id": cantus_id,
         "occurrences": occurrences,
         "fulltext": fulltext,
         "incipit": incipit,
+        "genre_name": genre_name,
         "genre_id": genre_id,
+        "clean_cantus_id": clean_cantus_id,
     }
 
 
@@ -124,6 +133,9 @@ def get_json_from_ci_api(
         response: requests.Response = requests.get(uri, timeout=timeout)
     except requests.exceptions.Timeout:
         return None
+
+    if not response.status_code == 200:
+        return None  # /json-cid/Non-existentCantusId returns a 500 page
 
     response.encoding = "utf-8-sig"
     return response.json()
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -19,12 +19,17 @@ NUMBER_OF_SUGGESTED_CHANTS: int = 5  # this number can't be too large,
 
 def get_suggested_chants(
     cantus_id: str, number_of_suggestions: int = NUMBER_OF_SUGGESTED_CHANTS
-) -> list[dict]:
+) -> Optional[list[dict]]:
     endpoint_path: str = f"/json-nextchants/{cantus_id}"
-    all_suggestions: list = get_json_from_ci_api(endpoint_path)
+    all_suggestions: Optional[list] = get_json_from_ci_api(endpoint_path)
+    if not all_suggestions:
+        return None  # get_json_from_ci_api timed out
+
     sort_by_occurrences: function = lambda suggestion: int(suggestion["count"])
-    sorted_suggestions = sorted(all_suggestions, key=sort_by_occurrences, reverse=True)
-    trimmed_suggestions = sorted_suggestions[:number_of_suggestions]
+    sorted_suggestions: list = sorted(
+        all_suggestions, key=sort_by_occurrences, reverse=True
+    )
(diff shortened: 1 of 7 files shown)


django/cantusdb_project/cantusindex.py
```diff
@@ -0,0 +1,150 @@
...
+        "fulltext": fulltext,
+        "incipit": incipit,
+        "genre_name": genre_name,
+        "genre_id": genre_id,
+        "clean_cantus_id": clean_cantus_id,
+    }
+
+
+def get_json_from_ci_api(
+    path: str, timeout: int = DEFAULT_TIMEOUT
```
dchiller:
```suggestion
    path: str, timeout: float = DEFAULT_TIMEOUT
```

`DEFAULT_TIMEOUT` is a float.

django/cantusdb_project/cantusindex.py
```diff
@@ -0,0 +1,150 @@
...
+        Optional[dict]: A dictionary with the following keys:
+            - "cantus_id"
+            - "occurrences"
+            - "fulltext"
+            - "incipit"
+            - "genre_id"
+            ...but if get_json_from_ci_api timed out, returns None instead
+    """
+    endpoint_path: str = f"/json-cid/{cantus_id}"
+    json: dict = get_json_from_ci_api(endpoint_path, timeout=timeout)
```
dchiller: Two things:
- This might be `None`. This case is handled below.
- This might be a List. Where is that handled?

django/cantusdb_project/cantusindex.py
```diff
@@ -0,0 +1,150 @@
...
+# too high will cause the Chant Create page to take a very long time
+# to load. If/when we parallelize this process, we want to limit
+# the size of the burst of requests sent to CantusIndex.
+
+
+def get_suggested_chants(
+    cantus_id: str, number_of_suggestions: int = NUMBER_OF_SUGGESTED_CHANTS
+) -> Optional[list[dict]]:
+    endpoint_path: str = f"/json-nextchants/{cantus_id}"
+    all_suggestions: Optional[list] = get_json_from_ci_api(endpoint_path)
```
dchiller: Return type shows that this might be a `dict`. Is that handled?

django/cantusdb_project/cantusindex.py
```diff
@@ -0,0 +1,150 @@
...
+        # get_json_from_ci_api timed out
+        # or CI returned a response with no suggestions.
+        return None
+
+    # when Cantus ID doesn't exist within CI, CI's api returns a 200 response with `['Cantus ID is not valid']`
+    first_suggestion = all_suggestions[0]
+    if not isinstance(first_suggestion, dict):
+        return None
+
+    sort_by_occurrences: function = lambda suggestion: int(suggestion["count"])
```
dchiller: `function` is not a valid type. 
```suggestion
    sort_by_occurrences: Callable[[dict], int] = lambda suggestion: int(suggestion["count"])
```

django/cantusdb_project/cantusindex.py
```diff
@@ -0,0 +1,150 @@
...
+    sorted_suggestions: list = sorted(
+        all_suggestions, key=sort_by_occurrences, reverse=True
+    )
+    trimmed_suggestions: list = sorted_suggestions[:number_of_suggestions]
+
+    suggested_chants: list[dict] = []
```
dchiller:
```suggestion
    suggested_chants: list[Optional[dict]] = []
```

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -255,11 +255,59 @@ <h3>Create Chant</h3>
             <div class="search-bar mb-3">
                 {% include "global_search_bar.html" %}
             </div>
-            <div class="card w-100">
+
+            <div class="card w-100 mb-3">
                 <div class="card-header">
-                    <h5><a id="source" href="{% url 'source-detail' source.id %}">{{ source.title }}</a></h5>
+                    <h5><a id="source" href="{% url 'source-detail' source.id %}">{{ source.siglum }}</a></h5>
```
dchiller, line 245: Is this intentional?

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -255,11 +255,59 @@ <h3>Create Chant</h3>
...
+                                {% endif %}
+                            </div>
+                        {% if suggested_chants %}
+                            {% for suggestion in suggested_chants %}
+                                <input type="button" style="width: 80px" value="{{ suggestion.cantus_id }}" onclick="autoFill{{ suggestion.clean_cantus_id }}()" title="{{ suggestion.cantus_id }}"></input>
+                                <strong>{{ suggestion.genre_name }}</strong> - <span title="{{ suggestion.fulltext }}">{{ suggestion.incipit }}</span> (<strong>{{ suggestion.occurrences }}x</strong>)<br>
+                                <script>
+                                    function autoFill{{ suggestion.clean_cantus_id }}() {
+                                        document.getElementById('id_cantus_id').value = "{{ suggestion.cantus_id }}";
+                                        {% comment %} document.getElementById('id_genre').value = {{ suggestion.genre_id }}; {% endcomment %}
```
dchiller: Should this be deleted?

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -255,11 +255,59 @@ <h3>Create Chant</h3>
...
+                                {% if previous_chant.cantus_id %}
+                                    (Cantus ID: <b><a href="https://cantusindex.org/id/{{ previous_chant.cantus_id }}" target="_blank">{{ previous_chant.cantus_id }}</a></b>)
+                                {% endif %}
+                            </div>
+                        {% if suggested_chants %}
+                            {% for suggestion in suggested_chants %}
+                                <input type="button" style="width: 80px" value="{{ suggestion.cantus_id }}" onclick="autoFill{{ suggestion.clean_cantus_id }}()" title="{{ suggestion.cantus_id }}"></input>
+                                <strong>{{ suggestion.genre_name }}</strong> - <span title="{{ suggestion.fulltext }}">{{ suggestion.incipit }}</span> (<strong>{{ suggestion.occurrences }}x</strong>)<br>
+                                <script>
+                                    function autoFill{{ suggestion.clean_cantus_id }}() {
```
dchiller: Why not have one function and then pass in arguments?

dchiller: E.g.: 

```javascript
function autoFill(cantus_id, full_text, genre){

}
```

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -255,11 +255,59 @@ <h3>Create Chant</h3>
...
+                                        // in case suggestion.genre_id is None, don't try to change the #id_genre selector
+                                        if (genre_id == null) {
+                                            return
+                                        }
+                                        // Set the value, creating a new option if necessary
+                                        if ($('#id_genre').find("option[value='" + genre_id + "']").length) {
+                                            $('#id_genre').val(genre_id).trigger('change');
+                                        } else { 
+                                            // Create a DOM Option and pre-select by default
+                                            var newOption = new Option(genre_name, genre_id, true, true);
```
dchiller: Why is this necessary with genres? What is the situation where we want a genre that doesn't exist as an option to be created by a suggested chant?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2772,46 +2774,55 @@ def setUp(self):
...
-        response = self.client.post(
-            reverse("chant-create", args=[source.id]),
-            {
-                "manuscript_full_text_std_spelling": "initial",
-                "folio": "001r",
-                "c_sequence": "1",
-            },
-        )
+        with patch("requests.get", mock_requests_get):
+            response = self.client.post(
```
dchiller: Does a post request to `chant-create` make a call using `requests.get`?

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2772,46 +2774,55 @@ def setUp(self):
...
+        with patch("requests.get", mock_requests_get):
+            response = self.client.get(url)
         self.assertEqual(response.context["source"].title, source.title)
 
     def test_post_error(self):
         """post with correct source and empty full-text"""
         source = make_fake_source()
         url = reverse("chant-create", args=[source.id])
-        response = self.client.post(url, data={"manuscript_full_text_std_spelling": ""})
+        with patch("requests.get", mock_requests_get):
```
dchiller: Same here re: post requests

django/cantusdb_project/main_app/tests/test_views.py
```diff
@@ -2846,15 +2858,16 @@ def test_repeated_seq(self):
...
-        response = self.client.post(
-            url,
-            data={
-                "manuscript_full_text_std_spelling": fake_text,
-                "folio": TEST_FOLIO,
-                "c_sequence": random.randint(1, 4),
-            },
-            follow=True,
-        )
+        with patch("requests.get", mock_requests_get):
```
dchiller: Same here re: post request

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -1,20 +1,162 @@
...
+            return MockResponse(
+                status_code=200,
+                content=mock_cantusindex_data.mock_json_cid_909030_content,
+                text=mock_cantusindex_data.mock_json_cid_909030_text,
+                json=mock_cantusindex_data.mock_json_cid_909030_json,
+            )
+        else:  # imitating CI's behavior when a made-up Cantus ID is entered.
+            return MockResponse(
+                status_code=500,
+                content=None,
```
dchiller, line 145: `content` and `text` are not valid as `None` according to definition of `__init__` of `MockResponse`

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -1,20 +1,162 @@
...
+    Args:
+        url (str): a URL - a necessary argument for requests.get
+        timeout (int): we pass timeout as an argument to requests.get in get_json_from_ci_api,
+            so mock_requests_get is configured to accept this argument.
+
+    Raises:
+        ValueError: This function is configured to mock requests to specific URLs only, including
+            - /json-nextchants/001010
+        If a call to requests.get with a different URL is made while mock_requests_get is patching it,
+        a ValueError is raised.
```
dchiller:
```suggestion
        a NotImplementedError is raised.
```

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -1,20 +1,162 @@
...
+
+    Returns:
+        MockResponse: A mock response object
+    """
+    if timeout < 0.001:
+        raise requests.exceptions.ConnectTimeout
+
+    if not (
+        "https://cantusindex.uwaterloo.ca" in url or "https://cantusindex.org" in url
```
dchiller: Import this string from `cantusindex.py`?

django/cantusdb_project/main_app/tests/mock_cantusindex_data.py
```diff
@@ -0,0 +1,644 @@
...
+            "db": "PEM"
+		}
+    }
+}
+"""
+mock_json_cid_909030_content: bytes = bytes(
+    mock_json_cid_909030_text,
+    encoding="utf-8-sig",
+)
+mock_json_cid_909030_json: dict = {
```
dchiller: Is this the same as what is in `_content` and `_bytes`? Showing the conversion to json would make more sense to me.

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -118,3 +260,156 @@ def test_generate_incipit(self):
...
+
+        # I can't figure out how to get assertRaises to work - even when the assertion should fail, it doesn't.
+        # It's not of vital importance that this argument check work correctly, I think, so I'm giving up in
+        # an effort not to get bogged down.
+        # Leaving a couple of commented-out attempts here with things I've tried, for the hopeful benefit of some
+        # future, cleverer developer
+        # - Jacob dGM, April 2024
+
+        # with self.subTest(test="Ensure raises ValueError when path improperly formatted, attempt 1"):
+        #     self.assertRaises(ValueError, get_json_from_ci_api, "/for/some/reason/this/passes/even/with/a/leading/slash")
```
dchiller: As I read it, your function `get_json_from_ci_api` would raise `ValueError` if there *isn't* a leading slash.

django/cantusdb_project/main_app/templates/chant_create.html
```diff
@@ -282,9 +282,13 @@ <h5><a id="source" href="{% url 'source-detail' source.id %}">{{ source.siglum }
                                         document.getElementById('id_cantus_id').value = "{{ suggestion.cantus_id }}";
                                         {% comment %} document.getElementById('id_genre').value = {{ suggestion.genre_id }}; {% endcomment %}
                                         document.getElementById('id_manuscript_full_text_std_spelling').value = "{{ suggestion.fulltext }}";
-                                        var genre_id = {{ suggestion.genre_id }};
+                                        var genre_id = {{ suggestion.genre_id | default_if_none:"null" }};
                                         var genre_name = "{{ suggestion.genre_name }}";
 
+                                        // in case suggestion.genre_id is None, don't try to change the #id_genre selector
+                                        if (genre_id == null) {
```
ahankinson: Mmmmm. Python writing Jinja writing JavaScript. My favourite. 

You probably want `===` here.

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: @annamorphism was looking over my shoulder as I did a bit of wireframing. This, iirc, is to make it match the sidebar cards on other pages.

But it should have been in its own commit. Thanks for checking on this.

@annamorphism, all this sounds right, yes?

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: The question of `===` vs. `==` was on my mind while writing this. I think someone in a StackOverflow discussion said that `==` was correct, since professional JS developers will understand the difference between the two or some similar justification, but looking at it now, I'm inclined to agree with you. Let me make this change - thanks.

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: This is necessary because we're using the django-autocomplete-light widget for the Genre selector. I had these same questions of @lucasmarchd01 when he implemented the Suggested Feasts feature - apparently, although it doesn't feel like it should be necessary to add options to the selector, this is the recommended way to use JS to change the options when using django-autocomplete-light. I copied and lightly adapted the code from later in this template.

See #1063.

django/cantusdb_project/cantusindex.py
jacobdgm: You need to help me set up the Black VSCode extension sometime - thank you for pointing out all of the not-quite-correct type annotations in this PR...

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: I've added comments to the places where we are setting both select elements.

django/cantusdb_project/main_app/tests/test_functions.py
jacobdgm: Right. So considering there _is_ a leading slash, and no `ValueError` is raised, we expect `AssertRaises` to fail. And yet it doesn't, for reasons I don't understand.

As far as I can figure, the most important part of a test suite is not that all the tests pass, but that the tests fail when they're supposed to. Before writing `self.assertRaises(ValueError, get_json_from_ci_api, "path/without/leading/slash")`, which passes, I want to be sure that `self.assertRaises(ValueError, get_json_from_ci_api, "/path/with/leading/slash")` fails.

django/cantusdb_project/main_app/tests/test_views.py
jacobdgm: oh. this is a good point.

Let me think this through aloud: when we send a post request, the response will be an indication that the post request was received, along with a redirection status code indicating the next page the client should load. But `self.client.post` will not load the next page itself.

So long as my reasoning is correct, this patch is not necessary. Let me remove it. Thanks!

django/cantusdb_project/cantusindex.py
jacobdgm: thank you

django/cantusdb_project/main_app/tests/test_functions.py
jacobdgm: thank you

django/cantusdb_project/main_app/templates/chant_create.html
dchiller: Ok! That makes sense... I really was just making sure it didn't slip in unintentionally.

annamorphism: yes--OldCantus had the full title here (for some reason), which is probably why this got a title as well, but all the other similar sidebar cards use the siglum and we decided to make it match (better use of space, too.)

dchiller: Perfect! Going to "resolve"...

django/cantusdb_project/main_app/tests/test_functions.py
dchiller: Ok, so this test does fail for me:

image

dchiller: It seems to me that if what I'm getting is the desired outcome, then this would be a place for an `expectedFailure` decorator?

jacobdgm: It's bizarre that this would fail for you but not for me.

I've also been having problems with one `ChantModelTest.test_get_concordances` - if I run `python manage.py test main_app.tests` as a whole, it errors, but if I run `python manage.py test main_app.tests.test_models.ChantModelTest`, it passes.

Did you try the `self.AssertRaises(several_arguments)` version or the `with self.AssertRaises(ValueError)` version? Should I just go ahead and uncomment that one and delete the other?

django/cantusdb_project/main_app/templates/chant_create.html
jacobdgm: I think the bit I have been missing up until now is that I can potentially go `onclick="autoFillFeast({{feast.cantus_id}}, {{feast.name}}, etc.)`. Let me try to make this change. Thanks.

django/cantusdb_project/main_app/tests/test_functions.py
dchiller: > I've also been having problems with one `ChantModelTest.test_get_concordances` - if I run `python manage.py test main_app.tests` as a whole, it errors, but if I run `python manage.py test main_app.tests.test_models.ChantModelTest`, it passes.

You should definitely make an issue about this...this is bad!

> Did you try the `self.AssertRaises(several_arguments)` version or the `with self.AssertRaises(ValueError)` version? Should I just go ahead and uncomment that one and delete the other?

The latter.

jacobdgm: oh! I think I figured it out.

This assertion was behaving weirdly because `"/path/with/leading/slash"` or whatever was not in the list of paths that `mock_requests_get` was set up to handle. Now, `mock_requests_get` raises a `NotImplementedError` when an unexpected URL/path is passed, but previously, it raised a - you guessed it - `ValueError`.

jacobdgm:
I think I have addressed all the suggested changes! Thank you, @dchiller and @ahankinson, for your perceptive suggestions!

Once I figured out how to refactor the `autoFillSuggestedChants` functions, I went ahead and applied the same thing to the `autoFillFeast` functions.

dchiller (approved):

Issue #1420, opened 2024-04-19
Chants (not just sources) should be able to belong to a segment


dchiller:
For the to-be-created Benedicamus Domino segment and the to-be-merged Clavis Sequentiarum, we need chants to be able to belong to a segment, not just sources (a source may contain chants in multiple segments).

dchiller:
PR #1434 goes part of the way towards addressing this issue. The remaining steps are:

- [ ] Run the new `assign_chants_to_segments` command when these changes are introduced to staging and production servers.
- [ ] Remove that command from the repo (once this initial change has been made, there is no further need for the command)

dchiller:
Argh! I also realized that the `assign_chants_to_segments` command in #1434 does not assign sequences to their segment!  So I'll add that additional step here too: 

- [x] Revise `assign_chants_to_segments` command to assign `Sequence` objects to a segment as well. (Completed with #1499)

dchiller:
This issue is obviated by the new chant-level `project` relationship introduced in #1548.

PR #1434, opened 2024-04-30
Add `segment` field to `BaseChant` model

dchiller:
This PR introduces a change to the `BaseChant` model that allows chants to belong to segments. The field is also added to the `ChantCreate` and `SourceChantEdit` views. Relevant forms, templates, and tests are updated.

The PR also adds the new command `assign_chants_to_segments`. This is meant to be run one time to assign existing chants to the segment of the source they belong to. 

Contributes towards #1420. After this PR is merged on staging and production, the temporary command `assign_chants_to_segments` should be run. This command and the associated test can then be removed. These further steps have been added to #1420.


django/cantusdb_project/main_app/management/commands/assign_chants_to_segments.py
```diff
@@ -0,0 +1,23 @@
...
+    help = "Assigns all chants in the database to the segment of the source they belong to."
+
+    def handle(self, *args, **options):
+        sources = Source.objects.all()
+        for source in sources:
+            segment = Segment.objects.get(id=source.segment_id)
+            chants = Chant.objects.filter(source=source)
+            for chant in chants:
+                chant.segment = segment
+                chant.save()
```
lucasmarchd01: Maybe this won't be a problem since we are only looking at a smaller number of chants within each source, but typically in these management commands we have to process the chants in chunks so that we don't run into memory issues. In other commands it looks something like this (from `touch_all_chants.py`)
```
        CHUNK_SIZE = 1_000
        chants = Chant.objects.all()
        chants_count = chants.count()
        start_index = 0
        while start_index <= chants_count:
            self.stdout.write(f"processing chunk with {start_index=}")
            chunk = chants[start_index : start_index + CHUNK_SIZE]

            for chant in chunk:
                chant.save()
            del chunk  # make sure we don't use too much RAM
            start_index += CHUNK_SIZE
```

lucasmarchd01 (approved):
Do we also want the segment to appear on the Chant Detail page?

django/cantusdb_project/main_app/management/commands/assign_chants_to_segments.py
dchiller: ok, I'll implement something like that too here before merging.

dchiller:
> Do we also want the segment to appear on the Chant Detail page?

This is a good question...I'm not sure. In convo's with Debra I know we need the chants to be able to have this, but I'm not sure where/how she wants to display the information. There was talk of making a separate search view. 

Although it seems like it couldn't hurt to also have it on the Chant Detail page. Let me check in with her and I'll make a new issue with what she says.

PR #1437, opened 2024-05-02
Merge staging into production, 2 May 2024

dchiller:
To test:

- [x] Query parameters removed from links to home page in upper-left corner image and "About" menu (#1407)
- [x] `/search` permanent redirects to `/chant-search` (#1451)
- [x] `Resources > PDF Documents` --> `Resources > Reference and Instructional Documents` (#1449)
- [x]  Addition of segment field on chant model (#1434 and #1452)
    - [x] Existing chants are successfully assigned the segment of their source with the `assign_chants_to_segments` command
    - [x] No `segment` details show on Chant Create, Chant Edit, or Chant Detail pages
- [x] Office has been renamed to service on all public-facing views (#1400)
    - [x] `/chant-search`
    - [x] `/offices`
    - [x] Terms menu drop-down
    - [x] Chant Create form
    - [x] Chant edit form
    - [x] Chant detail page
    - [x] `/source/<source_id>/inventory header 
- [x] "Manuscript" has been replaced with "source" (#1446)
    - [x] Chant list
    - [x] Source detail
    - [x] Source edit
- [x] Chant create: suggested chants feature (#1422)
- [x] Chant edit: suggested fulltexts feature (#1436)
- [x] Duplicates removed from feast-folio dropdowns (#1447)
- [ ] ~Populate missing `diff_id` field with `populate_diff_id_fields` command (#1445)~
- [x] Minor html updates (#1461)
    - [x] Copyright year in footer should be "2024"
    - [x] No pagination footer should be shown when Source search yields no results
    - [x] Link to field descriptions should work on Chant Create form
- [x] Test `populate_exists_on_cantus_ultimus` command (#1456)
- [x] Test ability to manually change existence on CU (#1456)
- [x] Links to images should be to CU where possible or to third-party institution when not on the following pages (#1456):
    - [x] Chant Detail
    - [x] Source Detail
    - [x] Browse chants in source
    - [x] Source edit

Full Change Log at https://github.com/DDMAL/CantusDB/releases/tag/17May2024


lucasmarchd01 (approved):

dchiller:
I ran `assign_chants_to_segments` overnight and thus far < 20,000 chants have been assigned, which means running the command to completion would take days. So, I am going to play around with this and try to make it more efficient.

Update: It seems like this would be a better job for `QuerySet.bulk_update` than `.save`, so giving that a go.

dchiller:
The `populate_diff_id_fields` command ran to completion. I'm collating all of the non-status messages here:

```
Differentia not found for chant: "Argentum et aurum non est " (710130)
Differentia not found for chant: "Quarum mentium inhabitator est Christus " (709595)
Differentia not found for chant: "Ille me clarificabit quia de " (702282)
```

The status message (i.e. "processing chunk with start_index N") did change partway through, so I am wondering if there were more of these chants for which Differentiae were not found that for some reason were not logged, because after running the command, we still have this:

image

This is worse, somehow, than what is currently on the production site:

image

annamorphism:
> The `populate_diff_id_fields` command ran to completion. I'm collating all of the non-status messages here:
> 
> ```
> Differentia not found for chant: "Argentum et aurum non est " (710130)
> Differentia not found for chant: "Quarum mentium inhabitator est Christus " (709595)
> Differentia not found for chant: "Ille me clarificabit quia de " (702282)
> ```

weirdly, this third one (which I indexed) doesn't seem to have a differentia database value at all, while the other two do, so I'm not sure why they both get "Differentia not found." All three are unpublished sources, fwiw.

lucasmarchd01:
This is such a mystery for me because when I ran this command locally it ran to completion and all chants were updated properly! I guess back to the drawing board for this issue...

dchiller:
> weirdly, this third one (which I indexed) doesn't seem to have a differentia database value at all

It does! We just don't display it anymore in the UI.
image

dchiller:
> This is such a mystery for me because when I ran this command locally it ran to completion and all chants were updated properly! I guess back to the drawing board for this issue...

I have two thoughts:
1. The command takes a while to run so some sort of weird timeout is happening, and we should run this in the background. 
2. Back before I used the update command and was doing a similar process for filling in the `segment` field, I was getting a similar issue, and I wonder if this kind of bulk iteration is kind of unstable with the ORM and if a more explicit transaction management is necessary. I don't immediately know, but it's maybe something to look into.

Given that we need to deploy to new production today, I'm thinking that we keep this issue open and just don't run this command on production as of yet (as long as we don't run the command, nothing changes with this). If the only thing that's happening is that this command misses some chants for some reason, then running it multiple times (which I am currently attempting on staging) would theoretically work. It's just a little too hacky, though, for me to want to try on production without knowing what's actually going on.

lucasmarchd01:
> Given that we need to deploy to new production today, I'm thinking that we keep this issue open and just don't run this command on production as of yet 

I agree, I would like to find out why this is happening to avoid it in the future. Have you tried running it on your local machine as well?

PR #1445, opened 2024-05-09
Populate missing `diff_id` Chant fields

### LINKED ISSUE #1368: Some chants with non-empty `differentiae_database` fields have empty `diff_id` fields
See, for example, chant 243157. 

image

I just looked back at the management command added in #1104 which I think was supposed to populate this field. If I manually go through the steps in that command with this chant, it looks like it should work. Did that script just not run to completion? Were there errors when it ran? 

Although I suspect that a re-run of this without errors would clear up the problem, this issue should also include a check that we aren't missing any differentiae database data from OldCantus.

lucasmarchd01:
In #1368, we noticed that there were many instances where a Chant's `differentiae_database` field was non-empty while the newly created `diff_db` field was empty.

This has occurred after running the management command in #1104 that had created new `Differentia` objects in the database and subsequently populated each chant with a ForeignKey `diff_id` that points to it's corresponding `Differentia` object.

For some reason, many chants were left unchanged (17634 on production). While I'm not 100% sure what caused this issue, I suspect it could be due to memory/resource problems while iterating through each chant in the database.

In this PR, I have created a new management command that will specifically target these chants with empty `diff_id` fields. Additionally, to reduce the risk of resource constraints, I have reduced the chunk size for batch processing. Running this script locally has eliminated all of the missing `diff_id` fields.

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
+++ b/django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
@@ -6,7 +6,7 @@ from typing import Optional
 
 class Command(BaseCommand):
     def handle(self, *args, **kwargs):
-        CHUNK_SIZE = 1_000
+        CHUNK_SIZE = 500
         chants = Chant.objects.filter(
             Q(differentiae_database__isnull=False) & Q(diff_db__isnull=True)
         )
--- a/django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
+++ b/django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
@@ -20,20 +20,16 @@ class Command(BaseCommand):
             for chant in chunk:
                 try:
                     differentia_id: Optional[str] = chant.differentiae_database
-                    if not differentia_id is None:
-
-                        differentia = Differentia.objects.get(
-                            differentia_id=differentia_id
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
```diff
@@ -0,0 +1,50 @@
...
+        start_index = 0
+        count = 0
+        while start_index <= chants_count:
+            self.stdout.write(f"processing chunk with {start_index=}")
+            chunk = chants[start_index : start_index + CHUNK_SIZE]
+
+            for chant in chunk:
+                try:
+                    differentia_id: Optional[str] = chant.differentiae_database
+                    if not differentia_id is None:
```
dchiller: just to be clear, we wouldn't expect this to catch anything because our query in line 10 already filtered out cases where `differentiae_database` was `None`, right?

dchiller (approved):
Looks good...my only comment is maybe don't close the issue yet since we have to remember to run this command (and then potentially remove the command) once we've deployed!

django/cantusdb_project/main_app/management/commands/populate_diff_id_fields.py
lucasmarchd01: Yes, that is true. I guess this was more of a sanity check when I was trying to figure out what was going wrong but it can be removed.

PR #1456, opened 2024-05-14
Display links to Cantus Ultimus where possible

### LINKED ISSUE #429: On Chant Detail page, we should add links to corresponding Cantus Ultimus page

### LINKED ISSUE #1079: Source Detail Page/Sidebar: Allow for multiple image views (Cantus Ultimus and host institution)

### LINKED ISSUE #1418: Change image links on individual chants to Cantus Ultimus where possible
After discussion with Debra, if a source is available on Cantus Ultimus, we want the image links on individual chants to be links to those folios on Cantus Ultimus. Note that this doesn't affect the various image views at the source level described in #1079

lucasmarchd01:
In this PR, we improve the connectivity between Cantus Database and Cantus Ultimus by adding links on various pages to Sources and Chants that exist on Cantus Ultimus.

To achieve this, there is a new BooleanField `exists_on_cantus_ultimus` on the `Source` model that can be manually set to True if the source has been added to Cantus Ultimus. This field is then used by the source and all corresponding Chants to display links where necessary. Here is a summary of the changes:

- Add field `exists_on_cantus_ultimus` to the `Source` model
- Add a management command `populate_exists_on_cantus_ultimus` that updates this field on a single Source or all Sources.
- Update sidebar templates (source detail/edit and browse chants) to include a "View inventory with images" link which points to CU, and "View images on external site" which points to the host institution.
- Update the chant detail template to check the Chant's Source and display the link to Cantus Ultimus for that particular folio and chant number. Also changes the "Display facsimile" link
Screenshot 2024-05-14 at 1 01 14 PM

- Update the browse chants template to check the Chant's Source and change the "Image" link to Cantus Ultimus instead of the host institution

On staging and production, we will have to run makemigrations and migrate, as well as running the new management command to see the changes reflected on the templates.

Resolves #1079, Resolves #1418, Resolves #429


django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
```diff
@@ -0,0 +1,93 @@
...
+            sources = Source.objects.filter(segment=4063)
+            for source in sources:
+                try:
+                    response: Response = requests.get(
+                        f"https://cantus.simssa.ca/manuscript/{source.id}", timeout=10
+                    )
+                except (TimeoutError, ConnectionError) as e:
+                    self.stdout.write(self.style.ERROR(f"{e} for source: {source.id}"))
+                # We expect a 404 response if the source doesn't exist on Cantus Ultimus
+                if response.status_code == 200:
```
ahankinson: This is almost completely a duplicate of the previous section. If you extract this to a separate method then you only have one chunk of code to maintain, instead of two!

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -215,6 +215,14 @@ <h3>{{ chant.incipit }}</h3>
                         <a href="{{ chant.image_link }}" target="_blank" style="word-break: break-all">{{ chant.image_link }}</a>
                     </dd>
                 {% endif %}
+                {% if chant.source.exists_on_cantus_ultimus %}
```
ahankinson: Remember that traversing multiple models in a template is ~the path to the dark side~ a surefire way to explode the number of queries that are performed on a page load. Make sure you test this!

django/cantusdb_project/main_app/templates/browse_chants.html
```diff
@@ -52,79 +52,84 @@ <h3>Browse Chants</h3>
...
-            {% else %}
-                No chants found.
-            {% endif %}
+                                <td class="text-wrap" style="text-align:center">{{ chant.position|default:"" }}</td>
+                                <td class="text-wrap" style="text-align:center">
+                                    <a href="{{ chant.get_ci_url }}" target="_blank">{{ chant.cantus_id|default:"" }}</a>
+                                </td>
+                                <td class="text-wrap" style="text-align:center">{{ chant.mode|default:"" }}</td>
+                                <td class="text-wrap" style="text-align:center">
+                                    {% if chant.source.exists_on_cantus_ultimus %}
```
dchiller: Since we are setting this on a source-by-source basis (setting aside the discussion in the related issue about whether there will ever be chants not on CU that are in sources that do exist on CU -- I think there we've determined we just won't handle those cases at the moment), do you need to check this? Won't it be the same for every chant in the source...so you could just check from the source?

dchiller: This point also makes me realize that we might need to update the `chant_search` template, because the chant-level image links are also shown there. I realize that is not discussed in any of the related issues, so I will create a new one for a separate PR.

dchiller: This is now #1459

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
lucasmarchd01: Good point, thanks!

django/cantusdb_project/main_app/templates/chant_detail.html
lucasmarchd01: Makes sense. I can't exactly test how many queries are being made because of #1457, but I assume passing this as a `context` template variable in the view would be better than accessing it through the template.

ahankinson: Or, even better, checking your `queryset` in your view before it goes to the template.

django/cantusdb_project/main_app/templates/browse_chants.html
```diff
@@ -52,79 +52,84 @@ <h3>Browse Chants</h3>
...
-                            </td>
-                            <td class="text-wrap" style="text-align:center">{{ chant.position|default:"" }}</td>
-                            <td class="text-wrap" style="text-align:center">
-                                <a href="{{ chant.get_ci_url }}" target="_blank">{{ chant.cantus_id|default:"" }}</a>
-                            </td>
-                            <td class="text-wrap" style="text-align:center">{{ chant.mode|default:"" }}</td>
-                            <td class="text-wrap" style="text-align:center">
-                                {% if chant.image_link %}
-                                    <a href="{{ chant.image_link }}" target="_blank">Image</a>
+            {% with exists_on_cantus_ultimus=source.exists_on_cantus_ultimus%}
```
lucasmarchd01: > do you need to check this? Won't it be the same for every chant in the source...so you could just check from the source?

You're right, and I actually put a source-level check earlier in the code here at line 55. I guess i just forgot I put it there.

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def update_exists_on_cantus_ultimus(self, source: Source) -> None:
+        try:
+            response: Response = requests.get(
+                f"https://cantus.simssa.ca/manuscript/{source.id}", timeout=10
+            )
+        except (TimeoutError, ConnectionError) as e:
+            self.stdout.write(self.style.ERROR(f"{e} for source: {source.id}"))
+        # We expect a 404 response if the source doesn't exist on Cantus Ultimus
+        if response.status_code == 200:
```
ahankinson: "Exit early, exit often"

Consider handling the negative case first and exiting, since this is the shortest path. Then you don't need the `else` block. 

(It's even a pylint rule, if you're into that sort of thing. https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html)

django/cantusdb_project/main_app/templates/chant_detail.html
lucasmarchd01: True, right now our `ChantDetailView` isn't using a `queryset` attribute and everything is being passed in as context data. If I'm understanding correctly, this is probably part of a larger issue where we could clean up and optimize our code by using `get_queryset` where possible.

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
```diff
@@ -0,0 +1,66 @@
...
+        source_id: int = kwargs["source_id"]
+        if source_id:  # update exists_on_cantus_ultimus for a specified source
+            try:
+                source: Source = Source.objects.get(id=source_id)
+                self.update_exists_on_cantus_ultimus(source)
+            except Source.DoesNotExist:
+                self.stdout.write(
+                    self.style.ERROR(f"Source {source_id} does not exist")
+                )
+                return
```
dchiller, line 37: I think this is superfluous?

```suggestion
```

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -156,6 +156,8 @@ def get_context_data(self, **kwargs):
         chants_current_folio = chants_in_source.filter(
             folio=chant.folio
         ).prefetch_related("feast")
+        print(self.queryset)
```
dchiller: Leftover print statement?
```suggestion
```

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
lucasmarchd01: Since we removed the else block we don't want the code under this if statement to run, and thus we need to exit the function early.

ahankinson: I also think that if you catch an exception but don't raise another one, then it will continue.

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def update_exists_on_cantus_ultimus(self, source: Source) -> None:
+        try:
+            response: Response = requests.get(
+                f"https://cantus.simssa.ca/manuscript/{source.id}", timeout=10
+            )
+        except (TimeoutError, ConnectionError) as e:
+            self.stdout.write(self.style.ERROR(f"{e} for source: {source.id}"))
+        # We expect a 404 response if the source doesn't exist on Cantus Ultimus
+        if response.status_code == 404:
```
ahankinson: Better to do `!= 200` since you don’t also handle other possible codes like 503 or 302.

django/cantusdb_project/main_app/management/commands/populate_exists_on_cantus_ultimus_field.py
lucasmarchd01: Oops, I was looking at a different part of the code.

ahankinson: That return was pretty useful, though, since it made it explicit that it exited after writing the log message. Now it exits because it passes the `else`. Functionally they're the same, but in terms of readability it's useful to signal that the function ends there by returning. 

Also, if you ever add anything to the end of the handler, it will introduce a problem because it's not immediately obvious that it keeps executing after it passes that `except`.

lucasmarchd01: Good point, I'll leave it in then!

dchiller (approved):

PR #1458, opened 2024-05-15
Use update rather than save in assign_chants_to_segments

dchiller:
In testing the `assign_chants_to_segments` command (introduces in #1434) for #1437, I was finding the command was taking a very long time (about 20,000 chants assigned over the course of about 12 hours). This meant we were looking at a weeks-long command run, which seemed ripe for things going wrong! 

This PR changes the command to use the `update` method to add the segment to the chants in a source. Run-time is considerably reduced in tests.

@lucasmarchd01 raised the point in #1434 that batching chants would reduce the possibility for memory error. The use of `update` also obviates that need because chant objects are never loaded into memory here, and the updates happen entirely on the SQL side (as such, the `update` method does not work on sliced query sets).

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/assign_chants_to_segments.py
+++ b/django/cantusdb_project/main_app/management/commands/assign_chants_to_segments.py
@@ -9,8 +9,6 @@ all chants in the database to the segment of the source they belong to.
 from django.core.management.base import BaseCommand
 from main_app.models import Source, Chant, Segment
 
-CHANT_CHUNK_SIZE = 1_000
-
 
 class Command(BaseCommand):
     help = "Assigns all chants in the database to the segment of the source they belong to."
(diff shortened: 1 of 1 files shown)


lucasmarchd01:
> @lucasmarchd01 raised the point in #1434 that batching chants would reduce the possibility for memory error. The use of `update` also obviates that need because chant objects are never loaded into memory here, and the updates happen entirely on the SQL side (as such, the `update` method does not work on sliced query sets).

Good to know, thanks!

lucasmarchd01 (approved):

dchiller:
> Good to know, thanks!

Yeah, it was a nice discovery this morning!

lucasmarchd01:
Something to keep in mind is that when using the `update` method on a chant wont call the `on_chant_save` method in `signals.py` won't be triggered. I don't think it will matter in this case but maybe a check to make sure would be good!

dchiller:
Yes...I checked here and the segment change won't implicate any of the signals, but an important consideration for other uses.

PR #1464, opened 2024-05-16
Fix django debug toolbar 

### LINKED ISSUE #1457: Django debug toolbar is not working
I just noticed that while working in my development environment (devcontainer setup), the Django debug toolbar doesn't show. Perhaps this was as a result of #1454?

lucasmarchd01:
This PR fixes the django debug toolbar not appearing in our development environment.

The issue seems to affect developers using docker (https://github.com/jazzband/django-debug-toolbar/issues/1854), and their solution assumes that the last segment of the IP should be 1, which doesn't work in our case.

In this solution, we add the `SHOW_TOOLBAR_CALLBACK` setting to the `DEBUG_TOOLBAR_CONFIG` which, when set to True, ignores the Internal IP setting.

The function `lambda request: (
        False if request.headers.get("x-requested-with") == "XMLHttpRequest" else True` checks if the request is Ajax, and returns True if it's not. 

`request.is_ajax()` is deprecated as of Django 3.1 ([release notes](https://docs.djangoproject.com/en/3.1/releases/3.1/#id2)) but this function does the same thing. 

Also, we upgrade the `django-debug-toolbar` to 4.3.0 because the current version we are using is incompatible with `psycopg  3.1.18`

@dchiller if you can, try this on your environment before approving to make sure it works

Fixes #1457

### DIFF
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -204,6 +204,12 @@ INTERNAL_IPS = [
 ]
 
 if DEBUG:
+    import socket
+
+    # Add the internal IP of the Docker container hosting the Django application.
+    # This line dynamically resolves the hostname to its internal IP address
+    INTERNAL_IPS.append(socket.gethostbyname("cantusdb-nginx-1.cantusdb_default"))
+
     INSTALLED_APPS.append("debug_toolbar")
     # debug toolbar must be inserted as early in the middleware as possible
     MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")
--- a/django/cantusdb_project/cantusdb/settings.py
+++ b/django/cantusdb_project/cantusdb/settings.py
@@ -197,6 +197,9 @@ DEBUG_TOOLBAR_CONFIG = {
         "debug_toolbar.panels.redirects.RedirectsPanel",
         "debug_toolbar.panels.profiling.ProfilingPanel",
     },
+    "SHOW_TOOLBAR_CALLBACK": lambda request: (
+        False if request.headers.get("x-requested-with") == "XMLHttpRequest" else True
+    ),
 }
 
 INTERNAL_IPS = [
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/cantusdb/settings.py
```diff
@@ -204,6 +204,12 @@
 ]
 
 if DEBUG:
+    import socket
+
+    # Add the internal IP of the Docker container hosting the Django application.
+    # This line dynamically resolves the hostname to its internal IP address
+    INTERNAL_IPS.append(socket.gethostbyname("cantusdb-nginx-1.cantusdb_default"))
```
dchiller: This is particular to your local system...for example, my nginx container is called `cdb-nginx-1`

dchiller: `socket.gethostbyname("cdb-nginx-1")` doesn't error but `socket.gethostbyname("cdb-nginx-1.cantusdb_default")` and `socket.gethostbyname("cantusdb-nginx-1.cantusdb_default")` do.

Update: `socket.gethostbyname("cdb-nginx-1.cdb_default")` does work as well.

dchiller: We could specify `container_name` in the docker compose files so everyone will always have the same container names

ahankinson: Or what about modifying option 3 here: https://ranjanmp.medium.com/django-debug-toolbar-not-showing-up-when-using-docker-django-docker-e79585813bc6

This solution could be modified to return true if the request is Ajax and the request is docker. I don’t know much about the docker setup, but this seems preferable over trying to list all the possible Ip addresses…

dchiller: Nice! I tried it earlier just setting the callback to true always, but there were some conflicts -- which now that I see this I know were from AJAX requests. This seems more robust

lucasmarchd01: Nice, option 3 works for me so I've added the changes here!

dchiller (approved):

PR #1473, opened 2024-05-23
Update Chant model and related forms for Benedicamus Domino-specific fields

### LINKED ISSUE #1469: Add chant fields for Benedicamus Domino segment
The following fields need to be added or "turned on" for the new Benedicamus Domino segment:

CM melody ID* (text box): Only accessible for BD segment chants
Function* (pull-down): Benedicamus Domino (“BD”), Processional (“P”), Historiae (“H”), Dramatic Element (“D”); Default blank
Incipit of the refrain* (text box): Only accessible for BD segment chants
Polyphony* (choice field): blank (default), 2-voice, 3-voice, 4-voice
Rubrics* (text box): Only accessible for BD segment chants for editing
Later addition* (text box): Only accessible for BD segment chants for editing

All fields, except for the `Function` and `Polyphony` fields should only be accessible once the "BD" segment has been selected.

dchiller:
Closes #1469.

This PR:
- adds some new fields to the BaseChant model that are (at least at this time) specific to the Benedicamus Domino segment:
    - `later_addition`
    - `incipit_of_refrain`
    - `cm_melody_id` (Corpus Monodicum melody id)
- adds two new choice fields (controlled vocab) to the BaseChant model that are not necessarily Benedicamus Domino-specific but that are necessary to begin indexing chants in that segment:
   - `polyphony` 
   - `liturgical_function`
- makes use of the existing `rubrics` field that already exists on the BaseChant model but to-date has not been included in chant creation or editing interfaces
- undoes #1452 (uncomments segment-related fields on the Chant Create and Chant Edit forms
- adds these fields to the Chant Create and Chant Edit forms:
    - `polyphony` and `function` fields are always visible
    - `later_addition`, `incipit_of_refrain`, `cm_melody_id`, and `rubrics` are only visible after the "Benedicamus Domino" segment is selected

This final point is illustrated in these screenshots. 

**Screenshot of Chant Create form with CANTUS segment selected** 
Note the addition of the `polyphony` and `function` fields

image

**Screenshot of Chant Create form with Benedicamus Domino segment selected** 

image

**Screenshot of Chant Edit form with CANTUS segment selected** 
Note the addition of the `polyphony` and `function` fields

image

**Screenshot of Chant Edit form with Benedicamus Domino segment selected** 

image

These additions require updates to our wiki documentation...I will complete this before merging if the PR looks generally good.


django/cantusdb_project/main_app/forms.py
```diff
@@ -67,6 +66,16 @@ def label_from_instance(self, obj):
...
+    """
+    This class inherits from NameModelChoiceField, but uses the
+    the custom SelectWidget defined in widgets.py as its widget
+    (for styling).
+    """
+
+    widget = SelectWidget()
+
+
 # 3 best
```
lucasmarchd01:
```suggestion

```
I don't know why this comment is here I saw it earlier when I was working on a different issue. Maybe we can just remove it.

django/cantusdb_project/main_app/models/base_chant.py
```diff
@@ -59,6 +60,25 @@ class Meta:
         "Office", on_delete=models.PROTECT, null=True, blank=True
     )
     position = models.CharField(max_length=63, null=True, blank=True)
+
+    BENEDICAMUS_DOMINO = "BD"
+    PROCESSIONAL = "P"
+    HISTORIAE = "H"
+    DRAMATIC_ELEMENT = "D"
```
lucasmarchd01, line 67: Why do we need to define these variables outside of `LITURGICAL_FUNCTION_CHOICES` if their full names are still represented in the list?

django/cantusdb_project/main_app/models/base_chant.py
```diff
@@ -116,6 +136,18 @@ class Meta:
     volpiano_notes = models.TextField(null=True, blank=True)
     volpiano_intervals = models.TextField(null=True, blank=True)
 
+    P2V = "2v"
+    P3V = "3v"
+    P4V = "4v"
```
lucasmarchd01, line 141: Same for these ones

lucasmarchd01 (approved):
I left a few comments that don't really make a difference if they're included or not, but everything looks good!

django/cantusdb_project/main_app/forms.py
dchiller: Yeah, I was wondering the same thing....I think if neither of us want to keep it, I'll remove!

django/cantusdb_project/main_app/models/base_chant.py
dchiller: I followed the model [here](https://docs.djangoproject.com/en/5.0/ref/models/fields/#choices). 

The string actually stored in the database is the first element of each tuple, and the second element is the "public-facing name". My understanding of defining them separately is so that you can do something like:

```python
from main_app.models import Chant
chants_processional = Chant.objects.filter(liturgical_function = Chant.PROCESSIONAL)
```

rather than having to remember that PROCESSIONAL is stored as "P" in the database. But you certainly don't *need* to do that

lucasmarchd01: Oh okay makes sense! Let's keep it this way then.

PR #1479, opened 2024-05-24
Add Management Command to Reflect Cantus ID Merge Events on CI to CD

### LINKED ISSUE #877: Cantus IDs sometimes get merged; Cantus Index has an API to show which Cantus ID have been merged

### LINKED ISSUE #1158: We need some method to track which Cantus IDs have been merged
Directly related to https://github.com/DDMAL/CantusDB/issues/877. For each Chant object in the database, it would be a good idea to keep track of each `cantus_id` it has had in the past. In case there is a problem with Cantus Index's https://cantusindex.org/json-merged-chants API, this would allow us to roll back erroneous changes.

There are several ways I can imagine setting this up.

One way would be to create a `CantusIDMergeEvent` model. I imagine it looking something like:
```python
class CantusIDMergeEvent(BaseModel):
    old_cantus_id = CharField() # or ForeignKey to CantusIdentifier model?
    new_cantus_id = CharField() # or ForeignKey to CantusIdentifier model?
    applied = BoolField() # has the merge been applied yet, y/n
    date_applied = DateTimeField() # if this merge has been applied, when did it happen
    ci_timestamp = CharField() # we've asked Jan to attach a timestamp to each merge event in the
        # json-merged-chants API, which we're hoping we can use as a unique identifier. This would
        # be the time that the merge was submitted on CantusIndex, i.e., when it was posted in the
        # json-merged-chants API
    applied_by = ForeignKey(model=User)

    chants = OneToMany(model=Chant)
    sequences = OneToMany(model=Sequence)
    # when a merge occurs, we use these two fields to record which chants/sequences were affected

    # we could consider adding a apply() and revert() methods, which would switch the `cantus_id` of
    # chant/sequence in the `chants`/`sequences` from `old_cantus_id` to `new_cantus_id` and back

    # N.B.: an apply() method would certainly require more thought - is `chants` calculated when a new
    # merger gets posted to CI's API, or when we commit the merge?
    # In any case, you get the general idea.
```

Since we're worried about Cantus IDs being changed programmatically when new merge events get posted to https://cantusindex.org/json-merged-chants (see https://github.com/DDMAL/CantusDB/pull/1141#issuecomment-1812749551 etc.), it would be easy to create a view that lists all the chants/sequences that would be affected by a given Merge Event, and would allow Debra to approve the merge. I'm hoping this would not be too much work for Debra to press "approve" on CD as well as on CI, and this would ensure that there's always a human in the loop when we're applying changes specified by Cantus Index.

Worth noting: this record will need to be attached to individual Chant/Sequence objects, and not to CantusIdentifier objects themselves (in case we go ahead with implementing them - see https://github.com/DDMAL/CantusDB/pull/1141 and https://github.com/DDMAL/CantusDB/issues/410). In case mutiple IDs get folded into a single ID and we need to roll back a single merge event, we'll need to know which of the chants associated with the new ID needs to be reset to the earlier ID.

lucasmarchd01:
In this PR, we set up a management command script to check the Cantus Index API endpoint `json-merged-chants`, which returns a list of dictionaries, each containing information about a merge event for Cantus ID's on Cantus Index.

- In `cantusindex.py`, a new function `get_merged_cantus_ids` is added that will retrieve the response from the `json-merged-chants` endpoint
- A management command `add_cantus_index_merge_events` is added, which will run under a cron job (I would say every week should be sufficient).
    - It uses the response from `get_merged_cantus_ids` to update the chants in our database with the new CantusIDs.
    - An optional "date" argument is added in case a user wants to run the management command manually and filter the response from a specific date.
- Fix typing in one other function inside `cantusindex.py`

We will still have to set up a cron job in https://github.com/dact-chant/ansible.cantus-db in order to get this working.

Resolves #877, Resolves #1158

### DIFF
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -6,6 +6,7 @@ Cantus Index's (CI's) various APIs.
 import requests
 from typing import Optional, Union, Callable
 from main_app.models import Genre
+import json
 
 CANTUS_INDEX_DOMAIN: str = "https://cantusindex.uwaterloo.ca"
 DEFAULT_TIMEOUT: float = 2  # seconds
@@ -112,7 +113,7 @@ def get_suggested_chant(
     }
 
 
-def get_suggested_fulltext(cantus_id: str) -> str:
+def get_suggested_fulltext(cantus_id: str) -> Optional[str]:
     endpoint_path: str = f"/json-cid/{cantus_id}"
     json: Union[dict, list, None] = get_json_from_ci_api(endpoint_path)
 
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/cantusindex.py
```diff
@@ -112,7 +113,7 @@ def get_suggested_chant(
     }
 
 
-def get_suggested_fulltext(cantus_id: str) -> str:
+def get_suggested_fulltext(cantus_id: str) -> Optional[str]:
```
dchiller, line 116: Nice catch!

django/cantusdb_project/main_app/management/commands/add_cantus_index_merge_events.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Filter merge events by the provided date
+        if date_filter:
+            merge_events = self.filter_merge_events_by_date(merge_events, date_filter)
+
+        # Apply the new merge events on Cantus DB
+        for tx in merge_events:
+            self.apply_transaction(tx)
+
+    def filter_merge_events_by_date(self, merge_events: list, date_filter: str) -> list:
```
dchiller: I think the type of `date_filter` won't be `str` here cause you already converted it to a datetime object at line 25.

dchiller (approved):
Looks good to me! (Found one minor type thing)

1. Do you want to also try to make the ansible change? I'm happy to do it, but if that would be interesting, go for it!
2. Love the logging on this command, and it's making me wish we had a better logging set-up so we could actually recover this when the cron command runs. Ah well....something for the future.

django/cantusdb_project/main_app/management/commands/add_cantus_index_merge_events.py
lucasmarchd01: Right, thanks!

lucasmarchd01:
> Do you want to also try to make the ansible change? I'm happy to do it, but if that would be interesting, go for it!

Sure! I'll try and set this up.

> Love the logging on this command, and it's making me wish we had a better logging set-up so we could actually recover this when the cron command runs. Ah well....something for the future.

Yeah... I figured it would be pretty useless for now since no one will see it but at least when we finally do set up logging it will be there!

PR #1482, opened 2024-05-27
Update footer logos

### LINKED ISSUE #1405: We need to update the footer
Debra sent us the following request:

> Hi Developers,
> Following our meeting in Halifax and the discussion about the footer and the acknowledgements, I have drafted up a possible version of the footer. See "mock up" document, attached.
> 
> Also, here is the link to images of all the logos: `(there was a link here; all the images within the folder are attached below)`

I will open another issue for the Acknowledgements page.

## Mockup:
Screenshot 2024-04-11 at 9 31 14 AM

## Logos:
image
image
image
image
image
image
image
image
image

## (Legacy Logos)
image
image
image
image
image
image

lucasmarchd01:
Debra has requested that we change the logos in the footer based on a mockup that she provided. Here is a possible configuration of the footer. Looking for feedback from anyone who uses the site!

Updated screenshot:
Screenshot 2024-05-28 at 9 57 08 AM

- I've added the static files for these logos under a new directory `static/footer` to keep things organized and deleted the now unused logos from the static folder.
- The "Cantus Team" list item for the developers is now changed to "Development and Maintenance" which links to the DDMAL website. 

Resolves #1405


annamorphism:
my personal feeling is it would be good to have "DDMAL | Development and Maintenance" (in parallel with the other two [entity]|[role] lines)--but I've been trying to find other examples of footers like this and not finding many, so I'm not sure!

lucasmarchd01:
I was thinking something similar to that as well. I wasn't exactly sure how Debra meant for us to implement https://github.com/DDMAL/CantusDB/issues/1405#issuecomment-2133394652 because I couldn't find the email. @dchiller what do you think?

annamorphism:
floated this edit by the authorities and they agreed!
so, for the third line, do DDMAL | Development & Maintenance (with "DDMAL" linked in red like the earlier lines.)

dchiller:
The logos look great imho! 

Ok, a little pedantic, but I think at the very least it would be "DDMAL Lab"? I could also see spelling it out (for clarity) though there may be no way to do that and make it look close to nice given how much more text is is than Jennifer or Debra's names.

dchiller (approved):
Looks good to me...just see comment about "DDMAL" text.

fujinaga:
Yes to "DDMAL Lab".

annamorphism (approved):

Issue #1485, opened 2024-05-28
IndexError at /chant/396856


lucasmarchd01:
Exception Type: 	IndexError
Exception Value: 	string index out of range

 `/code/.venv/lib/python3.9/site-packages/volpiano_display_utilities/text_volpiano_alignment.py, line 354, in align_text_and_volpiano `

chant_text :	`('Tu es vas electionis sancte paule apostole predicator veritatis in universo '
 'mundo | Exaudi deus orationem #')`

preprocessed_volpiano: 	`''`

review_encoding_flag: `True`

syllabified_text: 	`[SyllabifiedSection: [['Tu'], ['es'], ['vas'], ['e-', 'lec-', 'ti-', 'o-', 'nis'], ['sanc-', 'te'], ['pau-', 'le'], ['a-', 'po-', 'sto-', 'le'], ['pre-', 'di-', 'ca-', 'tor'], ['ve-', 'ri-', 'ta-', 'tis'], ['in'], ['u-', 'ni-', 'ver-', 'so'], ['mun-', 'do']],
 SyllabifiedSection: [['|']],
 SyllabifiedSection: [['Ex-', 'au-', 'di'], ['de-', 'us'], ['o-', 'ra-', 'ti-', 'o-', 'nem'], ['#']]]`

text_presyllabified: 	`False`

vol_chars_rmvd_flag: 	`True`

volpiano: 	`'\n'`

dchiller:
This is probably closed by #1487, but we should test on staging before we mark it as such.

dchiller:
Verified now that #1487 fixed this on production!

PR #1487, opened 2024-05-29
Upgrade to volpiano-display-utilities 1.1.3

dchiller:
Upgrade to the newest release of volpiano display utilities.

This should address #1485 once implemented, but I'll wait to close that issue until we can test.


lucasmarchd01 (approved):
Ok great!

Issue #425, opened 2022-11-17
Chant Search page: some fields are sorted by ID rather than name


jacobdgm:
On the chant search page, the sorting/ordering upon clicking on a column header is weird - for offices and genres, I think it is sorting by ID rather than by the display name for the office/genre (And since the offices were added in order of time during the day I think [their ID numbers are thus ordered chronologically], this has the felicitous but unintended effect of displaying results sorted by time-of-day when sorting by office, when the expected behaviour is that they be sorted alphabetically by abbreviation).

This may cause actual problems when sorting by source, since we sort by chants' `siglum` field, but we display the chant's source's `siglum` in the template (since chants' `siglum` fields are known to occasionally be out of date).

jacobdgm:
An attempt to fix this by sorting by `office__name` rather than just by `office` has the effect of causing an `ORDER BY term does not match any column in the result set django` error. Some preliminary googling suggested that this could be solved by  using `.prefetch_related()` during the querying of the database, but trying this did not immediately resolve the error. More research is necessary!

jacobdgm:
I tried what seemed to be another promising way to fix this, which also didn't work: applying

```python
class Meta:
    ordering = ["name"]
```
to the `Office` model causes `Office.objects.all()` to be ordered differently, but seems not to have any effect on the Search Chants page: swapping `"-name"` for `"name"` and running migrations, the search results are still in the same order. The order in which Offices appear on the chant page is consistent with them being ordered by `id`.

PR #1497, opened 2024-06-03
Fix sort headers on chant search page when using global search bar

### LINKED ISSUE #1089: sort headers don't work
When in chant search, the headers of each column LOOK like you can sort by them, and clicking on them gives you a dutiful little arrow pointing up or down. It doesn't, however, actually sort anything:

image

https://cantusdatabase.org/chant-search/?search_bar=notum%20fecit&order=office&sort=desc

lucasmarchd01:
This PR fixes the sort columns on the chant search page. It was discovered that the sort headers will not work if the chant search page is accessed via the global search bar, but work fine if a search is made using the keyword search bar.

The code fixes the logic inside `ChantSearchView` so that the sorting is applied at the end of building the queryset.

Test cases are also made to ensure that sorting while getting a search from the global search bar works as expected.

Fixes #1089

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views.py
+++ b/django/cantusdb_project/main_app/tests/test_views.py
@@ -1130,6 +1130,44 @@ class ChantSearchViewTest(TestCase):
         last_result_incipit = descending_results[1]["incipit"]
         self.assertEqual(last_result_incipit, chant_1.incipit)
 
+    def test_order_by_siglum_global_search(self):
+        source_1 = make_fake_source(published=True, siglum="sigl-1")
+        chant_1 = make_fake_chant(
+            manuscript_full_text_std_spelling="thing 1", source=source_1
+        )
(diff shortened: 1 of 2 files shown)


dchiller (approved):
Looks good! 

In reviewing this I learned that when we order by some fields we order by their ID rather than their name. So, we order by genre ID rather than name, when I think most people would assume that the ordering there is alphabetical by name. Odd! It turns out we have this referenced by #425 -- I'm going to change the name of that issue to something a little more descriptive.

lucasmarchd01:
True, definitely makes more sense to order some fields by name instead of ID! I can put that change in a separate PR.

dchiller:
> True, definitely makes more sense to order some fields by name instead of ID! I can put that change in a separate PR.

Cool! And yeah, definitely wan't expecting you ti change that in this PR -- but it does seems like something one of us should get to eventually

PR #1501, opened 2024-06-04
Preserve proofreader fields for contributor changes on chants

### LINKED ISSUE #1491: Contributor changes to a chant after proofreading overwrite proofreading fields
To replicate: 

- Log in as a proofreader of a source and edit a chant in that source by marking any of the "proofread" checkboxes, or adding someone to the `proofread_by` field, or by specifying a chant range.
- Log in as a contributor (ie. someone without proofread access to that source), and then make any edit to the chant.
- Log back in as the proofreader and see that all of your changes to those fields (proofread checkboxes, the `proofread_by` field, and the chant range field) have been overwritten. 

In emails with Debra, it sounds like the following would be preferable:
- contributor changes should never overwrite the `proofread_by` or `chant_range` fields
- the checkboxes are only "unchecked" by a contributor change if the contributor has actually changed something in one of the related fields ("volpiano proofread" gets unchecked only if the "volpiano" has changed, and so on)

lucasmarchd01:
This PR modifies the model form for the chant edit page so that contributor changes to a chant do not affect certain proofreader fields. If a contributor makes a change to a chant, the following conditions are met:

1. When a contributor edits a Chant, the following fields are preserved:
    - `chant_range`
    - `proofread_by`
    - `volpiano_proofread`
    - `manuscript_full_text_std_proofread`
    - `manuscript_full_text_proofread`

2. However, if the contributor makes changes to the following fields:
    - `volpiano`
    - `manuscript_full_text_std_spelling`
    - `manuscript_full_text`

Then the respective proofreader checkboxes (`volpiano_proofread`, `manuscript_full_text_std_proofread`, `manuscript_full_text_proofread`) are set to `False`. 

Fixes #1491

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -1105,7 +1105,7 @@ class SourceEditChantsView(LoginRequiredMixin, UserPassesTestMixin, UpdateView):
 
     def form_valid(self, form):
         if form.is_valid():
-            user = self.request.user
+            user: User = self.request.user
             chant: Chant = form.instance
 
             if not user_can_proofread_chant(user, chant):
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -25,7 +25,6 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.http import Http404, HttpResponse
 
 from django.contrib.auth.mixins import UserPassesTestMixin
-from django.contrib.auth import get_user_model
 
 
 from main_app.forms import (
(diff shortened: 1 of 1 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -1103,12 +1105,42 @@ def get_chants_with_folios(chants_in_feast):
...
+                    chant.manuscript_full_text_proofread = False
+
+            chant.last_updated_by = user
+            return_response: HttpResponse = super().form_valid(form)
+
+            # The many-to-many `proofread_by` field is reset when the
+            # parent class's `form_valid` method calls `save()` on the model instance.
+            if not user_can_proofread_chant(user, chant):
+                for user in proofreaders:
+                    chant.proofread_by.add(user)
```
dchiller: My only suggestion is that I think this can be simplified like this:

```suggestion
            if not user_can_proofread_chant(user, chant):
                chant.proofread_by.set(user)
```

dchiller (approved):

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: Nice, I've added this change with `chant.proofread_by.set(proofreaders)`

dchiller: Ok! Looks great to me!

PR #1513, opened 2024-06-06
Fixed: Break up admin.py

### LINKED ISSUE #110: Created sources not showing up in "sources created by user"

ahankinson:
- Moves each model admin to its own file
- Creates a shared base_admin file for the base model admin


dchiller (approved):

ahankinson:
Typically I would expect the approving reviewer to merge the PR, if that's OK?

PR #1514, opened 2024-06-06
Add institutions model

### LINKED ISSUE #1505: Create an Institutions Model
An "institution" model should be added to track the holding institution of a source. This will necessitate moving the "siglum" out of the source title into its own model. 

This means also adding a "holding institution" field to a source. 

The institution model will be based on (but not exactly) the DIAMM 'Archive" model: https://github.com/diamm/diamm/blob/main/diamm/models/data/archive.py

This will also include adding an "institution identifiers" table with several options for indicating the presence of that institution in other authorities -- DIAMM and RISM, in particular, but also probably wikidata and VIAF.

ahankinson:
This PR adds an institution model, an institution identifiers model, and adjusts the source model to accept a FK relationship to the holding institution. 

The fields are set to `blank=True, null=True` so that they can be added without triggering problems for existing data. I believe the intention is that every institution will have a name and a sigla. 

This PR uses the new layout of the `admin` site to add the new models, **so should be merged after** #1513 

Fixes #1505


ahankinson:
Just realized that with the slglum, I can probably also add the country to the institution, but this will need its own field.

ahankinson:
This is now ready for review

ahankinson:
Request to add 'former sigla' field from Debra.

ahankinson:
@dchiller @lucasmarchd01 this is ready for review when you get a moment

lucasmarchd01 (approved):
I can't exactly speak on the choices we are making for changing the data and models for CantusDB since I wasn't part of the entire discussion, but from a code standpoint everything looks okay to me!

django/cantusdb_project/main_app/models/institution.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+    former_sigla = models.TextField(
+        blank=True, null=True, help_text="Enter former sigla on separate lines."
+    )
+
+    def __str__(self) -> str:
+        names: list = [self.name]
+        if self.siglum:
+            names.append(f"({self.siglum})")
+        return " ".join(names)
```
dchiller: Isn't this just
```suggestion
        return f"{self.name} ({self.siglum})"
```

django/cantusdb_project/main_app/admin/institution.py
```diff
@@ -1,5 +1,26 @@
...
+
+@admin.register(Institution)
+class InstitutionAdmin(BaseModelAdmin):
+    list_display = ("name", "siglum", "get_city_region", "country")
+    search_fields = ("name", "siglum", "city", "region", "alternate_names")
+    list_filter = ("city",)
+    inlines = (InstitutionIdentifierInline,)
+
+    def get_city_region(self, obj):
+        name = [obj.city]
```
dchiller: `obj.city` might be None?

django/cantusdb_project/main_app/models/institution.py
ahankinson: The original checks if self.siglum is null and appends it to the list with the parens. In your version if it’s null you’ll get empty parens, or maybe a parenthesised “None”.

django/cantusdb_project/main_app/admin/institution.py
ahankinson: Yes! Will fix it tomorrow.

django/cantusdb_project/main_app/models/institution.py
dchiller: Yes...but self.siglum is a non-nullable field with a default value, right?

ahankinson: Yeah, that's true. But I guess I'm also hedging my bets that at some point it may not be a requirement, and I don't want this bit of code to break.

ahankinson: Fixed.

django/cantusdb_project/main_app/admin/institution.py
ahankinson: Fixed.

PR #1517, opened 2024-06-07
Fix `TypeError` in `get_suggested_chant()`

### LINKED ISSUE #1515: some IDs make create-chant break?
I just got an email from an indexer who had created the first chant in a source, but got a 500 error when trying to add a second. 
I tried myself and also can't seem to add a new chant to https://cantusdatabase.org/source/1000051. 

When I created a chant with identical information in a test source, it also gave me a 500 error, but added the chant to the source. When I then tried to add new chants to the test source, it gave a 500.  Remove the new chant--> can create as normal. I tried removing the content in various fields in the new chant, and strangely, deleting or replacing the Cantus ID seemed to solve the problem. 
Why would the ID [g00791](https://cantusindex.org/id/g00791) be making it so you can't create a new chant in that source?  Further investigation required.

lucasmarchd01:
This PR adds a try/except block inside the `get_suggested_chant` function.

Sometimes, the `json-cid` endpoint path will return a json response where the `info` key is `None`, causing a 500 error on the chant create page for certain CantusIDs.

Here, we return `None` if a `TypeError` exception is caught.

Will add more robust error handling for issue #1515, but we will leave the issue open until further checks are made on Cantus Index

### DIFF
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -91,9 +91,12 @@ def get_suggested_chant(
         # mostly, in case of a timeout within get_json_from_ci_api
         return None
 
-    fulltext: str = json["info"]["field_full_text"]
-    incipit: str = " ".join(fulltext.split(" ")[:5])
-    genre_name: str = json["info"]["field_genre"]
+    try:
+        fulltext: str = json["info"]["field_full_text"]
+        incipit: str = " ".join(fulltext.split(" ")[:5])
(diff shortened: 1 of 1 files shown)


dchiller (approved):
Looks good to me! 

Even if this closes the issue, I want to make sure @annamorphism takes a look at that specific Cantus ID in Cantus Index because in this specific case it is also a data integrity problem.

lucasmarchd01:
Yup, I just read more of the discussion that I missed so we can leave the issue open for now!

PR #1518, opened 2024-06-10
Update .gitignore

lucasmarchd01:
Adds `config/nginx/conf.d/cantusdb.conf.development` to `.gitignore` so we aren't tracking changes to this file. I found this while setting up my environment locally this week.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,6 +110,7 @@ config/envs/dev_env
 certificates/
 docker-compose.yml
 config/nginx/conf.d/cantusdb.conf
+config/nginx/conf.d/cantusdb.conf.development
 cron/cron.txt
 
 # Drupal scripts from old Cantus


dchiller:
I don't think we want to do this? 

This file is already in the repo, and I think is how we track any changes needed in the `cantusdb.conf` for local development.  See https://github.com/DDMAL/CantusDB/wiki/Deploying-CantusDB-locally-for-development#set-up.

lucasmarchd01:
Okay, that makes sense. git was still tracking `cantusdb.conf.development` as a file that has been deleted since it was renamed. I think then, instead, we should change the documentation from this:

`Rename config/nginx/conf.d/cantusdb.conf.development to cantusdb.conf.`

To this:
`Duplicate config/nginx/conf.d/cantusdb.conf.development and rename it to cantusdb.conf.`

dchiller:
Ah, yes. That documentation change makes sense (it was somehow what I did locally so I didn't noticed that the documentation was wrong :(

lucasmarchd01:
Yeah it wasn't an issue on my lab computer either so I guess I must have done the same thing! I'll go and update the page.

Issue #1474, opened 2024-05-23
Add simple password to staging server


dchiller:
Discussion began in #1303 

Adding a simple password would continue to provide wide access to those we might call on to look at staging but would prevent indexing/crawling of the staging site.

dchiller:
I think this is up to @fujinaga

fujinaga:
Ok. Please go ahead: medium priority?

PR #1520, opened 2024-06-11
Nginx: add custom 500 page, refactor error page static files, add simple password

### LINKED ISSUE #1519: Add custom 500 error page
This way, 500 errors will be returned with similar pages to 404, 502, and 504 responses.

dchiller:
This PR: 

- adds a custom 500 error page to match our custom 502 and 504 error pages
- makes the necessary changed to this repo for #1474 by adding an empty password file to the nginx container during the build process (the actual password file is provisioned to the staging server by ansible -- see https://github.com/dact-chant/ansible.cantus-db/pull/40 -- but this mock file allows builds to occur locally and on production)

Additionally, I learned in the course of solving #1519 that we were keeping an extra copy of a few static resources for use by the custom error pages. This was unnecessary and added some additional complexity to our nginx configuration. These custom error page static files are refactored here.

Closes #1519. Changes build process for the nginx container for #1474.


lucasmarchd01:
> This was unnecessary and added some additional complexity to our nginx configuration. 

Are you sure that they aren't kept there because resources located in the django container can't be located once we reach a 500 error? That could be why they're stored here as well, I'm not sure though.

dchiller:
> Are you sure that they aren't kept there because resources located in the django container can't be located once we reach a 500 error? That could be why they're stored here as well, I'm not sure though.

The resources are located in a volume that is shared between the nginx container and the django container. The collectstatic command in the django container moves static files to that volume, which is then accessible to the nginx container, but once so moved, the volume is independent of the django container. 

If you run the app, and the stop just the django container, the static files are still accessible (e.g. stop the django container, and then go to `http://localhost/static/js/chant_create.js` and you'll still get some js.

lucasmarchd01 (approved):

PR #1526, opened 2024-06-12
Add GitHub action to run tests; fix erroring test for ChantEditSyllabificationView

### LINKED ISSUE #1524: ChantEditSyllabificationViewTest fails

### LINKED ISSUE #1525: Add GitHub action for running test suite

dchiller:
This PR adds a GitHub action that will run the django test suite when a PR is opened or new commits are pushed. 

It also fixes an error in the `ChantEditSyllabificationView`. Version 1.1.0 of volpiano-display-utilties introduced an additional output to the `syllabify_text` function, which was not accounted for in the view. 

Finally, during the course of fixing these issues, I ran into a build error having to do with how `psycopg-binary` was specified in `pyproject.toml`. With this PR, a specific version of that package is specified.

Closes #1524. Closes #1525.

### DIFF
--- a/.github/workflows/django_tests.yml
+++ b/.github/workflows/django_tests.yml
@@ -1,0 +1,29 @@
+name: django-tests
+on:
+    pull_request:
+        types: [opened, synchronize]
+jobs:
+    run-tests:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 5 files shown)


lucasmarchd01 (approved):
This is exciting!

PR #1538, opened 2024-06-17
Re-implement cantus index input tool

### LINKED ISSUE #1391: We need to reimplement the CI Search View
This feature was turned off by https://github.com/DDMAL/CantusDB/pull/1142, and we need to turn it back on again.

https://github.com/DDMAL/CantusDB/commit/7b18c99a911bb0aac88886842af1ef011d1c6359 was the commit responsible for removing this functionality (https://github.com/DDMAL/CantusDB/commit/49a1978e20d65d06aacf407573dc3f2b1ce7789e could also be useful for writing tests for this view).

It occurs to me that I'm not aware of any CI endpoint that is directly helpful for turning this feature on. Previously, we had made a request to the page at https://cantusindex.org/search and parsed the HTML it returned; hopefully there's a better way to implement this feature. It might be worth asking Jan about how OldCantus did this.

lucasmarchd01:
This PR re-implements the Cantus Index search view on the chant create page. This feature was turned off in https://github.com/DDMAL/CantusDB/pull/1142, and many users have been requesting the feature again.

The previous implementation parsed HTML from https://cantusindex.org/search API. Now, we use https://cantusindex.org/json-text/. This API is still not available on cantusindex.uwaterloo.ca, so we will have to switch over once it is.

We still have to write tests for this view, so I will convert that to a new issue to get this up and running sooner.

Chant Create:
Screenshot 2024-06-17 at 12 26 54 PM
CISearchView:
Screenshot 2024-06-17 at 12 27 28 PM

Resolves #1391


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -869,6 +873,40 @@ def get_success_url(self):
...
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["genres"] = list(
+            Genre.objects.all().order_by("name").values("id", "name")
+        )
+        search_term: str = kwargs["search_term"]
+        search_term: str = search_term.replace(" ", "+")  # for multiple keywords
+
+        text_search_results: Optional[list[dict]] = get_ci_text_search(search_term)
```
dchiller: This type doesn't agree with the `get_ci_text_search` output type? The type output of the function suggests you'd need to check for `None`s in the `for` loop below.

dchiller (approved):

dchiller:
I find some things weird about the ci_search template, but am assuming you just re-added what we had before?

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: Ah, yes!

lucasmarchd01:
> I find some things weird about the ci_search template, but am assuming you just re-added what we had before?

Yeah, I thought about making it look more like our other pages by adding `base.html` and so on but I think it was made this way to save space and scrolling time. The template is exactly the same as before but with some added javascript to handle populating the django-autocomplete-light selector for genre.

lucasmarchd01:
Also @dchiller (or @annamorphism ) check out https://github.com/DDMAL/CantusDB/issues/1391#issuecomment-2176157470 just to make sure this is how we'd like to implement this.

dchiller:
> Also @dchiller (or @annamorphism ) check out [#1391 (comment)](https://github.com/DDMAL/CantusDB/issues/1391#issuecomment-2176157470) just to make sure this is how we'd like to implement this.

I think this is good for now ... If we figure out a way to increase the number of results in future then we make that a separate issue, but at least this feature will be available...

I do think the current message might therefore be a little misleading, so would agree about changing it to be something like the other option you suggested.

PR #1563, opened 2024-07-16
Put view classes in separate files

### LINKED ISSUE #1533: Separate views into their own files
The `views/views.py` module is over 1000 lines long. This file should be separated into more views, organized by logical grouping. Here's one proposal:

```
views/
... existing views
- autocomplete.py  # Autocompletes
- api.py  # any JSON API views
- export.py # data export, e.g., csv. Not sure if this should be in api too? 
- auth.py # any authentication views
- shared.py  # any functions that are shared across views
- flatpages.py  # any views for the flat pages (contact, about, etc.)
- redirect.py  # all redirects

### PR DESCRIPTION
**Pull request by dchiller**
Closes #1533.

This PR distributes view previously in `views/views.py` into:
- `api.py`: Contains any "api" views (views with responses that are not rendered and don't fit into redirect or autocomplete view categories)
- `auth.py`: Contains the "change password" view
- `autocomplete.py`: Contains views used in the implementation of autocomplete fields
- `contact.py`: Contains a view for the contact flatpage
- `redirect.py`: Contains redirect views
- `site_stats.py`: Contains views for the "item count" and "content overview" pages

For the future, it probably makes sense to manage the contact page with the flatpages app. I've added this as a low priority issue #1572.

### CODE REVIEW

File: django/cantusdb_project/main_app/urls.py [hunk 1]
```diff
@@ -7,6 +7,7 @@
     PasswordResetCompleteView,
 )
 from main_app.views import views
+from main_app.views import redirect
```
ahankinson: Seems strange that the others import the functions directly but these two use the module prefix...

django/cantusdb_project/main_app/views/site_stats.py
```diff
@@ -0,0 +1,102 @@
...
+def project_manager_check(user):
+    """
+    A callback function that will be called by the user_passes_test decorator of content_overview.
+
+    Takes in a logged-in user as an argument.
+    Returns True if they are in a "project manager" group, raises PermissionDenied otherwise.
+    """
+    if user.groups.filter(name="project manager").exists():
+        return True
+    raise PermissionDenied
```
lucasmarchd01: I feel like this function would make more sense in either `auth.py` or `permissions.py`. Maybe `permissions.py` would be better since this function isn't actually a view.

dchiller: Yeah, that makes sense. There are some other functions too (especially ones that are now in api.py) that are not views and are more helper functions that might similarly have a more logical home. 

I'll make this change as you suggest and might also move some of those other ones now too.

dchiller: Ok, never mind about other refactoring...I think it is good as is for now.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -167,3 +168,15 @@ def user_can_manage_source_editors(user: User) -> bool:
         or user.is_staff
         or user.groups.filter(name="project manager").exists()
     )
+
+
+def project_manager_check(user: User) -> bool:
```
lucasmarchd01: Great. Can we just rename it similarly to other functions in this file? Something like `user_is_project_manager`.

dchiller: Yep...and I'll fix the failing test.

django/cantusdb_project/main_app/views/site_stats.py
```diff
@@ -0,0 +1,90 @@
...
+    Sequence,
+    Source,
+    Feast,
+    Office,
+    Provenance,
+    Genre,
+    Notation,
+    Century,
+)
+from permissions import project_manager_check
```
lucasmarchd01:
```suggestion
from main_app.permissions import project_manager_check
```

lucasmarchd01 (approved):

Issue #1560, opened 2024-07-12
Combine feasts: 2456 -> 4474 and 2094 -> 4475


dchiller:
We need to: 

- ensure that all the feast data for feasts 2456 and 2094 have been transferred to 4474 and 4475 respectively
- "re-feast" all chants and sequences that have the old feast ID's so that they have the new feast ID's
- delete the old feasts

lucasmarchd01:
I think we can set up a management command that can do this. There's one thing I've noticed:

The feast date is different for feasts 2456 and 4474, which are August 8 and August 4, respectively. Should we use the date from the old feast (2456) or the new one (4474) when merging the data?

annamorphism:
My understanding is that the merged feast (4474) should have the 4 August date, corresponding to the new code 14080430.

dchiller:
> I think we can set up a management command that can do this.

Great point -- I could see this as being potentially useful functionality going forward.

lucasmarchd01:
The command has now been run successfully on staging. Once this reaches production we can run the command again and close the issue.

dchiller:
@lucasmarchd01 Thanks! I assigned you mostly so that we could see from the issue list that this was already under control -- but thanks for testing this too!

dchiller:
Run on production. 21 Oct 2024

PR #1564, opened 2024-07-16
Add command to reassign feast data and update chants

lucasmarchd01:
Adds a management command that will transfer data between feasts and update chants accordingly.

- Ensure data from old feasts are transferred to new feasts
- Update chants and sequences to use new feast IDs
- Delete old feasts after reassignment

Towards #1560, once we run the command on production we can close the issue.

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/reassign_feasts.py
+++ b/django/cantusdb_project/main_app/management/commands/reassign_feasts.py
@@ -2,16 +2,18 @@ from django.core.management.base import BaseCommand
 from main_app.models import Feast, Chant, Sequence
 
 
+FEAST_MAPPING = {
+    2456: 4474,
+    2094: 4475,
+}
+
+
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


dchiller (approved):
Looks good! My only thought would be to make the old feast, new feast parameters into command line arguments.

On the plus side, that would mean you could use this command additionally where needed without having to go through the review process of getting a change to the code merged.

On the minus side, maybe we want that kind of review for something that makes these kinds of bulk changes. Also this way we have a record of past "feast merges".

Ok, I think the way you did it was better!

lucasmarchd01:
> On the plus side, that would mean you could use this command additionally where needed without having to go through the review process of getting a change to the code merged.
> 
> On the minus side, maybe we want that kind of review for something that makes these kinds of bulk changes. Also this way we have a record of past "feast merges".

Good points! You're right, it's probably safest to have a review process for reassigning feasts. But I'll make the mapping a global variable for readability/modification in the future.

PR #1566, opened 2024-07-16
Fixed: A follow-up to the source rework

ahankinson:
This is a follow-up on the source / institution re-work. 

 - There were a few places where the source siglum was still being used, so I fixed this. I'll keep looking. 
 - Optimized some admin queryset with select_related
 - The check for what groups a user was in was triggering several SQL queries. A new template filter was added that takes a comma-separated string and uses that in a single `filter()` query.
 - The source browse chants view was doing a lot of work before a simple check that would trigger a 404. This check was moved up and simplified (It no longer checks the queryset with `in` but just checks the segment id matches the expected one).
 - The pagination custom tags were also slowing down the page loads, so they were optimized as well.

### DIFF
--- a/django/cantusdb_project/main_app/admin/chant.py
+++ b/django/cantusdb_project/main_app/admin/chant.py
@@ -8,10 +8,13 @@ from main_app.models import Chant
 @admin.register(Chant)
 class ChantAdmin(BaseModelAdmin):
 
+    def get_queryset(self, request):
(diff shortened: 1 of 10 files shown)


dchiller (approved):

dchiller:
@ahankinson Looks good! Realize you didn't explicitly request review yet so lmk if this is good to merge or I was premature...

ahankinson:
Yeah a merge is okay.

PR #1570, opened 2024-07-19
Further fixes for institution migration

ahankinson:
This PR introduces a couple small changes in support of the institution migration. 

 - Makes "title" and "siglum" read-only for sources. Since these are no longer used they should not be edited, but there was a request to keep them around for a while. Setting them to read-only in the admin interface removes all practical methods of adding or editing these fields until they are dropped entirely.
 - Makes a small adjustment to the migration script to add a RISM id to the "XX-NN" institution (Unknown).

### DIFF
--- a/django/cantusdb_project/main_app/admin/source.py
+++ b/django/cantusdb_project/main_app/admin/source.py
@@ -28,7 +28,7 @@ class SourceAdmin(BaseModelAdmin):
         "id",
         "provenance_notes",
     )
-    readonly_fields = READ_ONLY + (
+    readonly_fields = ("title", "siglum") + READ_ONLY + (
         "number_of_chants",
         "number_of_melodies",
         "date_created",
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/admin/source.py
```diff
@@ -28,7 +28,7 @@ class SourceAdmin(BaseModelAdmin):
         "id",
         "provenance_notes",
     )
-    readonly_fields = READ_ONLY + (
+    readonly_fields = ("title", "siglum") + READ_ONLY + (
         "number_of_chants",
         "number_of_melodies",
         "date_created",
```
lucasmarchd01, line 34:
```suggestion
    readonly_fields = READ_ONLY + (
        "title", 
        "siglum",
        "number_of_chants",
        "number_of_melodies",
        "date_created",
```

lucasmarchd01 (approved):

dchiller (approved):

django/cantusdb_project/main_app/admin/source.py
ahankinson: I did it this way so that title and siglum appear at the top of the read-only field block

PR #1574, opened 2024-07-24
Add legacy logos for acknowledgements page

### LINKED ISSUE #1406: We should add logos to the Acknowledgements page
In the same email that prompted #1405, Debra made the following request:

> For the "acknowledgements" page, could one of you figure out what looks good regarding putting the logos in as well as the names of the places?
> https://cantusdatabase.org/about/acknowledgements/
> I have logos for UWaterloo, Catholic University of America, Compute Canada, SIMSSA and Western. The listing of previous funders in the acknowledgements page is in the order that I like, so if there's a way to design it to include logos, that would be great.

(all logos are included in #1405)

https://cantusdatabase.org/about/acknowledgements is a flatpage, and unlike Articles, we don't have a wysiwyg editor set up for flatpages. If I were to implement this now, I would probably upload the necessary images as static files, and then hardcode links to them in the flatpages, manually styling them in the flatpage html. But this feels a bit hacky, and would leave several static files in the repository that we don't reference anywhere in the code. It might make sense to convert this one flatpage into a full-fledged view with its own template? I'd be interested to hear others' thoughts on how best to approach this.

lucasmarchd01:
This PR adds the necessary logos that were provided by Debra for the acknowledgements page. Once these static resources are available on staging and production, we can update the flatpage to include these logos.

Here is a potential mockup of the new acknowledgements page:
Screenshot 2024-07-24 at 12 28 46 PM

Towards #1406, we will close the issue once the flatpage is updated on production


dchiller (approved):
Looks good...should we not resolve the issue until the flatpage has been updated?

Issue #1484, opened 2024-05-28
500 Error for invalid search parameters


lucasmarchd01:
Following up on #1483, one of the main causes of 500 errors on our site is the use of invalid search parameters. These would only appear if the user is manually changing the value in the URL. For example:

https://cantusdatabase.org/source/620437/chants/?folio=104 will cause
Exception Type: 	ValueError
Exception Value: 	'104' is not in list

http://localhost/source/620437/chants/?feast=hello will cause
Exception Type: 	ValueError
Exception Value: 	Field 'id' expected a number but got 'hello'

We should include an exception block that checks for `ValueError` when retrieving search parameters anywhere on the site.

annamorphism:
marking as low, since this only happens if users modify the URL and not if they do expected site interactions.

PR #1576, opened 2024-07-29
Add indexing notes search to chant search page

### LINKED ISSUE #843: Indexing notes in "search manuscript"

lucasmarchd01:
This PR adds functionality to the chant search page by adding the ability to search for keywords in the indexing notes of a chant. This feature is available only on the "searching in source" page. It was a relatively new feature on Old Cantus, which is why it was probably missed when re-building the new site.

Screenshot 2024-07-29 at 10 36 32 AM

Resolves #843

### DIFF
--- a/django/cantusdb_project/main_app/views/chant.py
+++ b/django/cantusdb_project/main_app/views/chant.py
@@ -629,6 +629,12 @@ class ChantSearchMSView(ListView):
         search_melodies = self.request.GET.get("melodies")
         if search_melodies:
             search_parameters.append(f"melodies={search_melodies}")
+        search_indexing_notes_op = self.request.GET.get("indexing_notes_op")
+        if search_indexing_notes_op:
+            search_parameters.append(f"indexing_notes_op={search_indexing_notes_op}")
+        search_indexing_notes = self.request.GET.get("indexing_notes")
+        if search_indexing_notes:
+            search_parameters.append(f"indexing_notes={search_indexing_notes}")
 
         if search_parameters:
             joined_search_parameters = "&".join(search_parameters)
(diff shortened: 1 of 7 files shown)


django/cantusdb_project/main_app/views/chant.py
```diff
@@ -724,6 +730,14 @@ def get_queryset(self) -> QuerySet:
 
             keyword_filter = ms_spelling_filter | std_spelling_filter | incipit_filter
             queryset = queryset.filter(keyword_filter)
+        if notes := self.request.GET.get("indexing_notes"):
+            operation = self.request.GET.get("indexing_notes_op")
+            # the operation parameter can be "contains" or "starts_with"
+            if operation == "contains":
+                indexing_notes_filter = Q(indexing_notes__icontains=notes)
+            else:
```
dchiller, line 738: My only question here is about this `else`, and not doing a check that the operation = "starts_with"... I'm only thinking about this in relation to the kind of thing in #1484. Is this the kind of place you are envisioning checking? Or was your thought about #1484 to do that separately.

dchiller (approved):

django/cantusdb_project/main_app/views/chant.py
lucasmarchd01: Oh, you're right. I completely forgot about that issue. #1484 is concerned with the actual parameter values themselves, for example if a user inputs a string type value for an expected int type. In this case, the expected value can be anything, since we are doing a text search.

I copied this logic from another part of the code and did some testing, I don't think we'll encounter the 500 error here.

dchiller: Ok, great! I'm good to merge

PR #1577, opened 2024-07-29
Add many-to-many segment field to source model

### LINKED ISSUE #1549: Make `segment` into a many-to-many field
Change the `segment` relationship to be many-to-many so sources can belong to multiple segments.

Make the source's `segment` be editable on the Source Create and Source Edit page.

dchiller:
This PR adds a new many-to-many field, `segment_m2m`, to the source model. The name differentiates this field from the current foreign key `segment` field. Once the data currently in the `segment` field is transferred to the new field, the old `segment` field can be deleted and the new field renamed. 

Changes to the source create and source edit forms make this new field editable. Screenshots of these new fields are below. 

image

image

Closes #1549.


lucasmarchd01:
Might be pedantic, but do you think it's necessary to show the segment ID (4063 or 4064) to the user on the source create/edit page? I have a feeling someone might get confused that it might be a count value instead of an ID. Although I don't know if it's worth the trouble since it's just using the __str__ method of the segment model.

lucasmarchd01 (approved):

dchiller:
> Might be pedantic, but do you think it's necessary to show the segment ID (4063 or 4064) to the user on the source create/edit page?

Good point...I kinda tuned those out but you're right that this is confusing unless you know what's going on... I'll fix this.

PR #1579, opened 2024-08-06
Chant create: optimize suggested chants feature

### LINKED ISSUE #1424: Chant Create: Suggested Chants feature needs to be optimized
When putting together #1422, I had to set the number of suggested chants at 3 (instead of 5, as it had been previously). Even so, adding this feature has increased the load time on the Chant Create page by an order of magnitude.

The slowdown is caused by the fact that we're sending a bunch of requests to CI, and waiting for each response to come back before sending the next request.

There are a few ways I can think of to optimize this:
- parallelize the request.get calls to CI's `/json-cid` endpoint, using a library like `threading`
- implement a cache system to save CI's responses and only send a new request if our cached response is out of date by a certain amount of time

dchiller:
This PR eliminates all calls to the `json-cid` endpoint for the "suggested chants" feature on the Chant Create view. All data required for the suggested chants is now supplied directly in the response to the `json-nextchants` endpoint. 

Given that we no longer need to make these extra calls, this PR also returns the number of suggested chants presented to at most 5 rather than at most 3.

Modifies tests and relevant mock data to conform to this change in endpoint behaviour.

In the process, adds a few development-related items:
- adds the django-extensions and werkzeug packages so that the development django container runs `runserver_plus`. This server provides a shell for debugging any time a view errors -- a useful development feature.
- add the `types-requests` package with type stubs for the `requests` package.

Closes #1424


dchiller:
Tests are currently failing because we install the "production" version on the test runner but use the development docker-compose file (which now requires a development dependency for the `runserver_plus` command). I'll solve.

dchiller:
Ok, I think it makes the most sense to replicate the production environment on the test runner... so in that case we'll need to use a different command for the `django` container on the test runner.

lucasmarchd01 (approved):

PR #1584, opened 2024-08-08
New: Add source key filter to chants

ahankinson:
This small change adds a text input filter to the chants for the source primary key. This allows chants to be filtered by source in the list.


lucasmarchd01 (approved):

lucasmarchd01:
Nice, I didn't know about this feature!

dchiller (approved):

PR #1586, opened 2024-08-08
New: Add views for institutions

ahankinson:
This PR adds views for institutions, giving them each their own page on the Cantus site.

Special care was taken to get the source counts and source links correct for logged in / not logged in users and published / unpublished sources.

In addition external authorities are shown -- currently only RISM Online entries are in the external authorities, but others can be added. An entry for DIAMM as an external authority was also added so that links between Cantus and DIAMM are possible.


lucasmarchd01:
Great! Could you include screenshots of these new views?

ahankinson:
Here's the entry for the Baviarian State Library: 

image

Here's the list of all institutions:

image

dchiller (approved):
Nice!

PR #1592, opened 2024-08-09
Make "Country" and "Source" columns sortable on Source List page

### LINKED ISSUE #1217: Make source list page sortable by column headers
In an email from Debra:

> On the Source List page, we'd like to be able to sort search results by various available columns, like we can on the Chant Search page. We want to be able to sort by the Siglum and Title fields (Title field will be in column headed "Source" in the table).

dchiller:
Closes #1217 by making the source list page sortable by the "country" and "source" (city + institution + siglum) columns.

Additionally, this PR adds the `sortable_header` template tag. This tag renders a column header for a sortable column, reducing the need to repeat this construction multiple times in the `source_list` and other templates. This template tag can be reused in any template where we sort by column headers (I'm thinking here especially of the Chant Search table; I've made #1593 for this). 

Finally, makes some minor code cleanliness changes to the SourceList view: annotates function return type and makes use of the assignment operator to reduce the number of statements in the `get_queryset` view.

### DIFF
--- a/django/cantusdb_project/main_app/templates/source_list.html
+++ b/django/cantusdb_project/main_app/templates/source_list.html
@@ -1,4 +1,6 @@
 {% extends "base.html" %}
+{% load helper_tags %}
+
 
 {% block title %}
     <title>Browse Sources | Cantus Database</title>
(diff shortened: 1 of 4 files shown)


lucasmarchd01 (approved):
Nice, the inclusion_tag is a neat feature

Issue #1598, opened 2024-08-14
Production server is not using custom 500 error pages


lucasmarchd01:
For example:
https://cantusdatabase.org/source/620437/chants/?folio=104

lucasmarchd01:
I believe that the changes that were made in #1530 should be applied to the `cantusdb.conf` file on the production server instead of `cantusdb.conf.development`. 

We should also probably have the `cantusdb.conf` file stored on the DDMAL wiki resources page.

lucasmarchd01:
> We should also probably have the cantusdb.conf file stored on the DDMAL wiki resources page.

Nvm, they're on the ansible github page.

PR #1599, opened 2024-08-14
Update nginx configuration and switch to `runserver_plus` in devcontainer setup

lucasmarchd01:
This change disables nginx's error page interception for the development environment configuration file to ensure that Django's debug pages are displayed correctly when `DEBUG` is set to True. I think that this change was introduced as part of #1530 which should be done instead on the `cantusdb.conf` file on the production server. (Related: #1598)

Replaced `runserver` with `runserver_plus` in the Docker Compose configuration for the development container.


dchiller:
Good catch on the `proxy_intercept_errors`. I made the change in development (where you're right, it shouldn't be) but never actually did it in the playbooks :( 

Made a PR for that here: https://github.com/dact-chant/ansible.cantus-db/pull/48

dchiller (approved):

Issue #1550, opened 2024-06-19
Incorrect usage of the `<small>` element


ahankinson:
In several templates the `<small>` element is used to make everything in that block smaller. However, this element is not a block-level element, and should not be used this way.

The Bootstrap Toolkit has the `.small` class that can be used to make things smaller.

For reference: 

![image](https://github.com/DDMAL/CantusDB/assets/163183/dce8bdd6-8529-4604-8ed3-642b2525c65a)

![image](https://github.com/DDMAL/CantusDB/assets/163183/14e8cbc1-8b39-4e3e-9073-5e1a990332cb)

dchiller:
I'm going to hold off on this until we upgrade to Django 5. In starting to correct this, I discovered that most places we use this element are in forms, and that we do form fields inconsistently. Django 5 looks like it contains [changes](https://github.com/DDMAL/cantus/tree/main/cantaloupe) to form templating that would simplify templating form fields (in the process of this we could handle many instances of this issue).

PR #1606, opened 2024-08-19
Make `shelfmark` a required field on source create and edit forms

### LINKED ISSUE #1604: TypeError on Source Create page
Try and create a source with all of the fields empty. We get the following error:

`TypeError: sequence item 0: expected str instance, NoneType found`

This is currently on the develop and staging branches.

lucasmarchd01:
This was causing a TypeError on the source create and edit forms. Also adds the fix to the admin source create form. Fixes #1604


ahankinson:
Shelfmark should also be marked as required in the `Source` model. Also, `holding_institution` needs to be marked as required in both the editing form and the model.

dchiller:
> Shelfmark should also be marked as required in the `Source` model. Also, `holding_institution` needs to be marked as required in both the editing form and the model.

Was the fact that these weren't previously required because null needed to be allowed while we transitioned the records? But the intention was that they really shouldn't allow null after that?

ahankinson:
> Was the fact that these weren't previously required because null needed to be allowed while we transitioned the records? 

Yep.

dchiller:
Great... I think it makes sense to make those changes, and then when we update production we just won't apply this migration, then run `migrate_records`, and then apply this last migration.

@lucasmarchd01 Do you want to add those changes here or we can create another issue?

lucasmarchd01:
> @lucasmarchd01 Do you want to add those changes here or we can create another issue?

Sure, we can add those changes here.

django/cantusdb_project/main_app/templates/source_create.html
```diff
@@ -57,7 +57,7 @@ <h3>Create Source</h3>
                 <div class="form-row">
                     <div class="form-group m-1 col-lg-3">
                         <label for="{{ form.shelfmark.id_for_label }}">
-                            <small>Shelfmark:</small>
+                            <small>Shelfmark:</small><span class="text-danger" title="This field is required">*</span></small>
```
ahankinson: See: https://github.com/DDMAL/CantusDB/issues/1550

Better would be to add the `.small` class to the `<label>` element.

lucasmarchd01:
There's a migration script that needs to be added, but I can generate that after #1615 is merged.

django/cantusdb_project/main_app/templates/source_create.html
lucasmarchd01: I think I've fixes the ones for these fields by using `class='form-control-sm` on the `<label>` element. But there's many more of these cases we'll want to fix in the future.

dchiller (dismissed):
Cool. Good to merge imo with the migration added :)

dchiller (approved):

PR #1629, opened 2024-09-13
Institution source changes

ahankinson:
This PR addresses the outstanding issues that were brought out in our e-mail conversations:

 - Creates a "Source Identifiers" model and adds it to the admin UI (no public view of this data yet)
 - Adds the 'name' field for colloquial names for sources
 - Changes "Private" to "Cantus" for privately-held sources
 - Adjusts some labels in the source view and list templates


django/cantusdb_project/main_app/forms.py
```diff
@@ -720,22 +720,27 @@ class Meta:
...
+    #     widget=TextInputWidget,
+    #     help_text="Full Source Identification (City, Archive, Shelf-mark)",
+    # )
+    # title.widget.attrs.update({"style": "width: 610px;"})
+    #
+    # siglum = forms.CharField(
+    #     required=True,
+    #     widget=TextInputWidget,
+    #     help_text="RISM-style siglum + Shelf-mark (e.g. GB-Ob 202).",
+    # )
```
lucasmarchd01, line 734: If we aren't planning on reintroducing these fields in the admin form (I see currently they're read only in the admin page) then we can probably remove this comment.

Is the idea that eventually we will remove title and siglum data all together?

lucasmarchd01 (approved):

django/cantusdb_project/main_app/forms.py
ahankinson: Yeah, I guess I’m just hedging my bets here, waiting for someone to ask “but wait, where did the siglum field go” (and then another 128 emails follow ;-)

dchiller (approved):

Issue #1631, opened 2024-09-18
Remove requirement for sources to have a holding institution if source is a reconstruction


dchiller:
As is being discussed in #1630, the holding institution relationship on a source is ill-suited for source reconstructions. While source reconstructions may, at some point, require additional attention in terms of how they are modeled, at present they can be accommodated by removing the requirement for a source to have a holding institution. 

- [ ] Make "Reconstruction" an option in the `full_source` field. (Change to ChoiceField).
- [ ] Remove the requirement that a source have a holding institution relationship

ahankinson:
> Add a Boolean field to the source model tracking whether or not the source is a reconstruction

Right now `full_source` is a Boolean field in the database, but a choice field in the form. This seems really weird. 

Why not make an actual choice field with `Full Source`, `Fragment`, and `Reconstruction` as options?

dchiller:
> Right now `full_source` is a Boolean field in the database, but a choice field in the form. This seems really weird.

Hmm... that is weird. Yeah, this would be a good chance to clean that up. Revised steps above.

dchiller:
Decided with Debra, Anna, Lucas that no source should have to have a holding institution.

PR #1645, opened 2024-10-02
Adjust source fields: holding_institution, name, production_method, source_completeness

### LINKED ISSUE #1628: Add a field to identify if a Source is a printed book or a manuscript
This should be a checkmark or dropdown for printed.

dchiller:
NOTE: This PR removes some previously-committed migrations. You will need to re-create your local/staging database to do development after this PR is merged. I decided to do this because it will make the update of production go smoother (ie. not have to run to a certain migration, then run the migrate_records command, then apply the rest of the migrations), which ultimately felt preferable to prioritizing people's individual current development database status.

Makes significant changes to the `Source` model:
- removes the requirement for a `Source` to have a `holding_institution`. #1631 requests that holding_institution-less sources be possible only for reconstruction sources. This will be added once `source_completeness` field is available to mark reconstructions as such (see To-do's below). 
- adds default `[No Shelfmark]` to `Source.shelfmark` field.
- adds the `source_completeness` choice field to replace the `full_source` boolean field and add the "Reconstruction" option. A command, `populate_source_completeness`, will populate the initial values of this field. Note that after this new field is populated, the `full_source` field will be redundant and can be removed (see To-do's below).
- adds the `production_method` choice field to note (currently) whether a source is a manuscript or printed source. 

These changes are reflected in source create, edit, detail, and list pages. Relevant tests are also updated.

Makes cosmetic changes to the `Institution` model by changing various field defaults from latin abbreviations to english words.

Adds the source's `name` field to source create, edit, and detail views and to the source heading.

Performs some minor cosmetic changes: refactors `forms.py` a bit with a new `StyledChoiceField` class so that repeated changes to widget style attributes are no longer necessary and changes the ordering of the options in the "Holding Institution" input field on the Source admin page. 

### To-Do's
- run `populate_source_completeness` command on servers
- mark source reconstructions as reconstructions and fix orphaned sources (those sources that don't have proper institutions in the migration)
- add constraint to sources that require holding institutions unless they are reconstructions (completing #1631)

Closes #1628.

Most display changes occur on the source list and source detail pages. In screenshots below, disregard repetitions of "(fragment)" on new version...these will be fixed in underlying data.

| Page | Old version | New version |
|-----|------------|--------------|
| Source list (colloquial name/fragment) | image | image |
| Source list | private collection as shown in previous row | image |
| Source detail (colloquial name) | image | image |
| Source detail (fragment/private collection) | image | image |


dchiller:
The failing test will be fixed with the merge of #1643 -- once that is reviewed and merged I'll fix this.

django/cantusdb_project/main_app/forms.py
```diff
@@ -733,19 +714,9 @@ class Meta:
...
-
-    name = forms.CharField(
-        required=False,
-        widget=TextInputWidget
-    )
-
     holding_institution = forms.ModelChoiceField(
-        queryset=Institution.objects.all().order_by("name"),
-        required=True,
+        queryset=Institution.objects.all().order_by("siglum"),
```
ahankinson: Is this still correct? As noted previously, this will cause Germany to fall under "D", Switzerland under "CH" and the UK under "GB".

django/cantusdb_project/main_app/models/source.py
```diff
@@ -176,13 +204,13 @@ def short_heading(self) -> str:
...
-            elif holdinst.is_private_collector:
-                title.append("Cantus")
+        else:
+            title.append("Cantus")
 
         tt = self.shelfmark if self.shelfmark else self.title
         title.append(tt)
 
-        if not self.full_source:
+        if self.source_completeness == self.SourceCompletenessChoices.FRAGMENT:
```
ahankinson, line 219: Will this start to make the "Short" heading not-so-short?

django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -34,15 +34,20 @@ <h3>{{ source.heading }}</h3>
 {% endif %}
 
 <dl>
+    <dt>Cantus Siglum</dt>
+    <dd>{{ source.short_heading }}{% if source.name %} <i>"{{ source.name }}"</i>{% endif %}</dd>
```
ahankinson: The point of the `heading` and `short_heading` methods is to make display consistent across the site. So if the name should be displayed with the short heading, it should probably go in the helper method, and not on a case-by-case basis in the templates.

django/cantusdb_project/main_app/models/source.py
dchiller: Idk ... I can ask Debra to confirm.

django/cantusdb_project/main_app/forms.py
dchiller: We can change to `city`. What I don't like about current is that it is hard to tell in the display what the order is. Siglum is marked off with parentheses and city is at the start of the display, so I think one of those is preferable to name.

image

django/cantusdb_project/main_app/templates/source_detail.html
dchiller: I'm not sure it should in all cases?...in any case, realizing there are a few different persistent source display discrepancies, this among them, so will ask Debra.

django/cantusdb_project/main_app/forms.py
ahankinson: Whatever it is, it should be obvious in the display...

django/cantusdb_project/main_app/models/source.py
ahankinson: I can imagine something like `(f)` in the short display, and `(fragment)` in the full one?

django/cantusdb_project/main_app/forms.py
dchiller: Ok, I'll make it city then. I think alphabetical by first word the user sees is probably the most straightforward.

dchiller:
Tests are now passing. Have not heard back confirmation on some of the remaining formatting questions, so holding for the moment.

dchiller:
Ok, after discussing the remaining questions with Anna, Debra, and Jennifer: 

- `(fragment)` goes after source title and source Cantus siglum where relevant
- `("Colloquial name")` goes after title on source detail page
- `"Colloquial name"` goes at the start of the summary column on the browse sources page

dchiller:
@lucasmarchd01 this is ready for you now

lucasmarchd01 (approved):
Looks good to me!

PR #1653, opened 2024-10-04
Handle errors with improper chant text field characters/syllabification fails

### LINKED ISSUE #1423: Handle `LatinError` exception in views that use volpiano display utilities
https://github.com/DDMAL/volpiano-display-utilities/pull/52 introduces the raising of `LatinError` when non-latin characters are passed in a word to be syllabified.

Views in CantusDB that use volpiano display utilities should handle this error. In particular, if entered text results in this error, users should be prompted to fix their text.

### LINKED ISSUE #1626: mispaired brackets lead to server error
In a variation of #1417 , pairing [ and } in the source spelling field caused an error that couldn't be fixed on the user end. (I haven't tested if this is also true in the standard spelling field, or what happens if the open/close bracket is missing altogether)

dchiller:
Adds custom validators to all forms involving chant texts (chant create, chant edit, and syllabification edit) so that invalid texts cannot be saved.

Nests most attempted syllabifications and text-melody alignments in `try` blocks so that existing syllabification errors won't cause pages to fail to load.

Does some refactoring on a variety of `test_views.test_chant.py` tests to simplify, remove duplicate tests, and catch linting errors. In particular, two tests (`test_view_url_reverse_name` and `test_template_used`) repeated the logic of other tests for the same view (`test_url_and_templates` and `test_view_url_path`).

Closes #1423. Closes #1626.


django/cantusdb_project/main_app/forms.py
```diff
@@ -153,14 +189,18 @@ class Meta:
...
 
-    manuscript_full_text_std_spelling = forms.CharField(
+    manuscript_full_text_std_spelling = CantusDBLatinField(
+        widget=TextAreaWidget,
+        help_text=Chant._meta.get_field("manuscript_full_text_std_spelling").help_text,
+        label="Full text as in Source (standardized spelling)",
         required=True,
+    )
+
+    manuscript_full_text = CantusDBLatinField(
```
lucasmarchd01, line 209: Love this.

lucasmarchd01 (approved):

PR #1691, opened 2024-11-01
Fix `TypeError` when using CI json-cid endpoint

### LINKED ISSUE #1690: `TypeError` when using CI endpoint for suggested full text
In an email, Debra pointed out that she's been encountering 500 errors when trying to edit certain chants. She pointed out that all of these chants don't have full texts.

The error occurs when we try to access the `json-cid/{cantus_id}` endpoint on Cantus Index. We should add an exception that handles when `field_full_text` is `None`.

lucasmarchd01:
This pull request fixes a `TypeError` in the` get_suggested_fulltext` function by adding exception handling for `TypeError`.

Previously, the function raised a `TypeError` when attempting to access a key in json_response that was `None` for chants that don't have a full text.

Also adds some black formatting changes.

Fixes #1690

### DIFF
--- a/django/cantusdb_project/cantusindex.py
+++ b/django/cantusdb_project/cantusindex.py
@@ -112,11 +112,10 @@ def get_suggested_fulltext(cantus_id: str) -> Optional[str]:
         # mostly, in case of a timeout within get_json_from_ci_api
         return None
 
-    try:
-        suggested_fulltext = json_response["info"]["field_full_text"]
-    except (KeyError, TypeError):
+    info: Optional[dict] = json_response.get("info", {})
+    if not isinstance(info, dict):
         return None
-
+    suggested_fulltext: Optional[str] = info.get("field_full_text")
     return suggested_fulltext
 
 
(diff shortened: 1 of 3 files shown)


dchiller (approved):

ahankinson:
In situations like this I generally prefer the `.get()` method, since it doesn’t raise anything. You can also chain them by making the default an empty dict.

`suggested_fulltext:Optional[str] = json_response.get("info", {}).get (“field_full_text")`

lucasmarchd01:
> In situations like this I generally prefer the `.get()` method, since it doesn’t raise anything. You can also chain them by making the default an empty dict.
> 
> `suggested_fulltext:Optional[str] = json_response.get("info", {}).get (“field_full_text")`

The only problem I came across when testing this is that sometimes "info" will be `None` leading to an AttributeError on the second chain. So we'll have to do something like this instead:

```
    info: Optional[dict] = json_response.get("info", {})
    if not isinstance(info, dict):
        return None
    suggested_fulltext: Optional[str] = info.get("field_full_text")
    return suggested_fulltext
```

ahankinson:
This will do pretty much the same thing without needing the `isinstance` check...

```
info: dict = json_response.get("info", {}) or {}
return info.get("field_full_text")
```

PR #1706, opened 2024-11-11
Implement proofread filter functionality to Browse Chants page

### LINKED ISSUE #1492: Proofreaders need some easier way to see their "task list" (chants that need proofreading)
In discussions with Debra:

lucasmarchd01:
This PR introduces functionality for filtering by proofreading status on the "Browse Chants" page. It includes:

- Filtering functionality for `manuscript_full_text_proofread`, `manuscript_full_text_std_proofread`, and `volpiano_proofread fields`.
- JavaScript updates to update URL parameters dynamically as the user selects different proofreading options.
- Forms with ChoiceField and RadioSelect widgets for filtering chants based on proofreading status.
- User permission checks to determine if a user can proofread a specific source.
- Templates are updated to reflect the new proofreading filter options.
- Add a test suite for these changes

Updated screenshot:
image

Towards #1492


lucasmarchd01:
Just adding another thought here...
I implemented this on the Source Browse Chants page. However, it could be on the source inventory page instead (or both). My thought was that since the browse chants page also has the "Edit" links it would be easier for a proofreader to filter and edit these chants directly. But if this looks to cluttered or belongs on another page we can definitely switch that before merging.

django/cantusdb_project/main_app/permissions.py
```diff
@@ -61,6 +61,28 @@ def user_can_proofread_chant(user: User, chant: Chant) -> bool:
     return user_is_project_manager or (user_is_editor and user_is_assigned_to_source)
 
 
+def user_can_proofread_source(user: User, source: Source) -> bool:
```
dchiller, line 68: If you've written this function, the function right above could make use of it!

django/cantusdb_project/main_app/templates/browse_chants.html
```diff
@@ -58,6 +58,21 @@ <h3>Browse Chants</h3>
                 </select>
             </div>
         </div>
+        {% if user_can_proofread_source %}
```
dchiller, line 61: I find the radio buttons really close together with the text -- any chance the options could be more spatially differentiated?

django/cantusdb_project/main_app/views/source.py
```diff
@@ -84,8 +93,17 @@ def get_queryset(self):
...
+            "manuscript_full_text_proofread"
+        )
+        manuscript_full_text_std_proofread = self.request.GET.get(
+            "manuscript_full_text_std_proofread"
+        )
+        volpiano_proofread = self.request.GET.get("volpiano_proofread")
+
         # get all chants in the specified source
-        chants = source.chant_set.select_related("feast", "service", "genre")
+        chants: QuerySet = source.chant_set.select_related("feast", "service", "genre")
```
dchiller:
```suggestion
        chants: QuerySet[Chant] = source.chant_set.select_related("feast", "service", "genre")
```

dchiller (approved):
What's here look good, but I think it would be to not close #1492 yet, because what's here is useful for seeing what chants in a source need proofreading, but you have to know which sources have these chants to make it useful. I'm not sure if that's always the case? Maybe @annamorphism can let us know based on experience. If it's helpful to see this at the source level too, we may want to add a similar function to the Source List page.

django/cantusdb_project/main_app/templates/browse_chants.html
annamorphism: > I find the radio buttons really close together with the text -- any chance the options could be more spatially differentiated?

I agree--I think it's that the buttons before and after "Yes" are both about equally close...

annamorphism:
> What's here look good, but I think it would be to not close #1492 yet, because what's here is useful for seeing what chants in a source need proofreading, but you have to know which sources have these chants to make it useful. I'm not sure if that's always the case? Maybe @annamorphism can let us know based on experience. If it's helpful to see this at the source level too, we may want to add a similar function to the Source List page.

@dchiller yes, it's not always known at the source level if a source is "done" or not--for example, I know I've proofread *some* fragments from Trier and not others. This way I can go in and check, but a little flag that "this source has some chants that need proofreading" would help find them.

PR #1712, opened 2024-11-18
fix(global search bar): display result data with embedded search bar

### LINKED ISSUE #1703: Chant details don't show up in search bar results on genre and service pages
They show up fine on the home page but not these (and potentially other) pages that use the global search bar. 

image

dchiller:
Create two new extendable templates (`single_column_with_search_bar.html` and `filterable_table_with_search_bar.html`) for two common use cases with an embedded search bar: one more generally for the embedded search bar and one for views that display filterable tables.

Adjust all current views to use these templates where relevant.

Put the search bar in a bootstrap column so that dynamic search result previews fill the correct horizontal space.

Closes #1703

**Old look of the Genre list page while using the global search bar**

image

**New look of the Genre list page while using global search bar**

image

### DIFF
--- a/django/cantusdb_project/main_app/templates/century_detail.html
+++ b/django/cantusdb_project/main_app/templates/century_detail.html
@@ -1,23 +1,24 @@
-{% extends "base.html" %}
+{% extends "single_column_with_search_bar.html" %}
 
 {% block title %}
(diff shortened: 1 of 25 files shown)


lucasmarchd01 (approved):
Very nice!

PR #1725, opened 2024-12-06
Update bootstrap 5

### LINKED ISSUE #1462: Bootstrap 4 is EOL
Just discovered that the version of Bootstrap we use is EOL -- https://getbootstrap.com/docs/4.6/end-of-life/

### LINKED ISSUE #1721: Bootstrap and DAL jqueries conflict
After some troubleshooting, I found the Django Autocomplete Light uses the version of JQuery included in the Django admin site, but (unlike pages on the Django admin site), doesn't namespace that version of JQuery to `django.jquery`... which leads to conflicts with the JQuery required by Bootstrap 4. 

Since Bootstrap 4 has been EOL since January 2023, this might be a good reason to escalate #1462

dchiller:
Updates the version of Bootstrap used in CantusDB to Bootstrap 5. Closes #1462.

This PR touches many many files, but most of those changes are cosmetic and/or involve replacing deprecated Bootstrap classes with new ones. I suspect the best way to review is going to be to pull this branch and walk through pages locally. Note: static files must be refreshed!

A few other minor additions to this PR:
- all custom stylings were moved to a single static .css file
- all templates were reformatted using `djlint`
- the `docker-compose-development.yml` file was rewritten to not require a pin for debugging with `runserver_plus`. Since this command is only ever run locally, we don't need the security of a pin.
- in the course of this upgrade, I observed that many of our tests that involve checking whether some string appears in rendered templates depend on spacing and other html formatting of the django templates. I'm skeptical of these kinds of checks in general, but at the very least, they shouldn't fail if there's an added or removed new line in a django template. To that end, I've written a custom mixin that provides a new assertion, `assertParsedContains`, that maps closely to Django's `assertContains` method, but, by parsing the html, is more robust to these types of changes to html that do not impact rendered html. The additional of `BeautifulSoup` and `lxml` dev dependencies are used for this assertion; however, no production dependencies are added!

Also closes #1721.


lucasmarchd01:
Screenshot 2024-12-17 at 3 56 03 PM


When I pull the branch and collect static pages it looks like this. Is that what you're seeing as well?

dchiller:
Hmm...that's not what I have. Here's what I get: 

image


Try doing a hard refresh of the page?

Edit: When you look at those links on the side in the developer tools, where is the styling coming from? For example, there should be color styles coming from `style.css`... but those don't seem to be applied here. Is `style.css` downloaded by your browser?

lucasmarchd01:
Hard refresh fixed it! Not sure what was going on, but the css styling was fixed as well.

lucasmarchd01 (approved):

PR #1735, opened 2025-01-16
Add proofread status field to sources and browse sources table

### LINKED ISSUE #1492: Proofreaders need some easier way to see their "task list" (chants that need proofreading)

lucasmarchd01:
This pull request introduces a new field to the Source model that tracks whether the source has any chants that are not fully proofread. The field is updated via the save signal on the Chant model and ignores the `volpiano_proofread` field (as requested by Debra in meetings).

- Added `all_chants_proofread ` boolean field to the Source model.
- Implemented logic to update the proofread status field on chant save.
- Created a management command to update the proofread status for all existing sources.
- Updated the Browse Sources table to include a column showing whether a source has chants needing proofreading
    - Accessible only by project managers.
    - Added a sortable column header.

Resolves #1492

Screenshot 2025-01-16 at 3 34 30 PM


dchiller:
Two comments: 
1. I'm not sure this is how we want to handle chants that have `NULL` text proofread fields. Right now, those will be counted as proofread, right? Seems to me it should be the other way around.
2. Since this field doesn't really capture whether the chant is proofread, but rather just that the text fields have been proofread, I think naming and labeling should reflect that.

lucasmarchd01:
Good points, thanks!

django/cantusdb_project/main_app/signals.py
```diff
@@ -103,6 +105,23 @@ def update_source_melody_count(instance) -> None:
         source.save()
 
 
+def update_source_proofread_status(instance) -> None:
```
dchiller, line 108: I'm concerned about a similar point to #1741 here.

If there is a chant in a source where there is no `manuscript_full_text` given, then I presume that the `manuscript_full_text_proofread` will not be `True`. But should that count as the source not being fully proofread?

Just some quick querying and it looks like there are a number of sources (even in the CANTUS segment) where chants like this exist...

Now, it could be that we want to count these as not proofread because we need a proofreader to come along, realize there's no `manuscript_full_text` and put it in, but if not we will need to adjust this logic. 

Maybe @annamorphism can weigh in here.

For example, here's a chant where there is a standard spelling full text which has been proofread, but no manuscript spelling full text (and therefore a `False` `manuscript_full_text_proofread` field): https://cantusdatabase.org/chant/677608

Do we want this chant to count against its source's `all_chant_texts_proofread` calculation?

annamorphism: Hmm...I guess they are two separate questions. On the one hand, since this MS seems to have some chants with full texts and not others, I might want somebody to put that in. But that seems like a separate task from making sure that everything that is there already is accurate (and can be done by a different person--not necessarily somebody with proofreader status.) And (as with the Volpiano field), there's a different sort of accountability to "this extra information is missing" from "this information is wrong".

dchiller: We could also ask Debra ... the purpose of this notification is really to help you and Debra (and any other proofreaders) know where your attention is needed, so whatever you would like best. Ultimately, these kinds of distinctions would be nice for you to be able to search for on your own (for example, I don't think you can currently search for chants *without* a full text? But I think that's something we add when we move toward Solr search.

lucasmarchd01:
Debra has clarified that we will want a separate field that is not reliant on these chant checkboxes. This new field should indicate that **all** of the fields in the record have been proofread. See #1762

Moving forward I think that we should either change the column name for this field, since it is misleading, or we can pause/remove this feature and implement #1762. Thoughts?

Issue #435, opened 2022-11-30
Implement Solr to improve load times on search views?


jacobdgm:
At the LinkedMusic project meeting, someone raised the point that fast load times are the feature that everyone wants but no one asks for, and that developers often don't bother trying to fix long load times when performing database searches because they don't know there are ways to make things faster. Solr was suggested as a tool for optimizing search times. Solr is already used on [CantusUltimus](https://github.com/DDMAL/cantus), I believe, so this might be a natural thing to implement.

This will require some amount of research, and its implementation is not urgent - just putting this here so as not to forget about it.

fujinaga:
Yes, this will be good.

jacobdgm:
If I understand correctly, we're trying to use only django builtins for our searching, in order to not add an extra layer of complexity by adding Solr. Closing.

fujinaga:
No, we would like to implement Solr at some point unless all searches are blazingly fast.

ahankinson:
Quite a lot of the search problems you’ve been having have very good solutions in Solr. I’ve been avoiding commenting because I think it would be a distraction at this point to introduce Solr, and not helpful for just getting the site stable and launched. But I would be happy to provide some guidance about this when the time comes.

kyrieb-ekat:
Closing in favor of Django search improvements. May revisit in future.

PR #1737, opened 2025-01-24
Add quoted search functionality on browse sources page

### LINKED ISSUE #1632: Source search could be better
If I search North Vancouver in Browse Sources, I get 29 results (although only one source is from North Vancouver)--because 2 sources mention Vancouver and another 27 mention "north". The search "North Vancouver" in quotes gets me nothing. If I search Vancouver I can get the source I want, but it seems counterintuitive to take out information to get what I need. 
Similarly, searching Tours Bibliotheque Municipale gives me 29 sources; adding in the number 149 brings it up to 83. For some reason in this case putting it in quotes still gives me 49 results, not 0 as it did for North Vancouver. Not sure why that is. 
Ideally at least one of "Tours Bibliotheque Municipale" in quotes or out of quotes should get me to the sources from that place!

lucasmarchd01:
Fixed some bugs in the search functionality on the Browse Sources page by adding support for quoted phrase matching, allowing exact matches for phrases (e.g., "North Vancouver"). 

Enabled unaccented search terms to match fields with or without accents, applying both partial (icontains) and exact (iexact) matching logic across key fields like institution name, city, shelfmark, and more. 

Updated the query logic to handle quoted and unquoted terms separately using regex for better granularity and consistency.

Added tests for these changes.

Resolves #1632. We should proceed with #435 as part of a more robust fix for these types of issues.

### DIFF
--- a/django/cantusdb_project/main_app/views/source.py
+++ b/django/cantusdb_project/main_app/views/source.py
@@ -331,7 +331,9 @@ class SourceListView(ListView):  # type: ignore
             # field, allowing for a more flexible search, and a field needs
             # to match only one of the terms
             for term in general_search_terms:
-                holding_institution_q |= Q(holding_institution__name__icontains=term)
+                holding_institution_q |= Q(
+                    holding_institution__name__unaccent__icontains=term
+                )
                 holding_institution_city_q |= Q(
                     holding_institution__city__icontains=term
                 )
--- a/django/cantusdb_project/main_app/views/source.py
+++ b/django/cantusdb_project/main_app/views/source.py
@@ -319,7 +319,7 @@ class SourceListView(ListView):  # type: ignore
                 r'"(.*?)"', general_str
             )  # Extract terms in quotes
             unquoted_terms = re.findall(
-                r"\b[\w-]+\b", re.sub(r'"(.*?)"', "", general_str)
+                r"\b[\w,-.]+\b", re.sub(r'"(.*?)"', "", general_str)
             )
 
             # We need a Q Object for each field we're gonna look into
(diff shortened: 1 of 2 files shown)


django/cantusdb_project/main_app/views/source.py
```diff
@@ -311,8 +312,17 @@ def get_queryset(self) -> QuerySet[Source]:
...
-            general_search_terms = general_str.strip(" ").split(" ")
+            # Strip spaces at the beginning and end
+            general_str = general_str.strip()
+
+            # Use regex to extract quoted and unquoted terms
+            quoted_terms = re.findall(
+                r'"(.*?)"', general_str
+            )  # Extract terms in quotes
+            unquoted_terms = re.findall(
+                r"\b[\w,-.]+\b", re.sub(r'"(.*?)"', "", general_str)
```
dchiller, line 323: You could also do this with one regex with look-aheads/look-behinds:

`(?<!\")\b[\w,-.]+\b(?!\")`

django/cantusdb_project/main_app/views/source.py
```diff
@@ -327,27 +337,35 @@ def get_queryset(self) -> QuerySet[Source]:
...
-            # field
-            # general_search_q = (
-            #     title_q | siglum_q | description_q | provenance_q
-            # )
+                name_q |= Q(name__unaccent__icontains=term)
+
+            # Add quoted terms to the Q object with exact matching (iexact)
+            for term in quoted_terms:
+                holding_institution_q |= Q(
+                    holding_institution__name__unaccent__iexact=term
```
dchiller: I don't think `iexact` is correct here...

If someone searches for "North Vancouver", don't I want any results where "North Vancouver" shows up somewhere in one of these fields, not just results where a field is exactly "North Vancouver"? 

I think `contains` is probably still the method, it's just we're not searching single words, but sometimes groups of words depending on the quotes.

django/cantusdb_project/main_app/views/source.py
dchiller: Probably clearer what's happening here with the way you have it now...

dchiller (approved):

Issue #1239, opened 2024-01-05
Source Detail: we should stop preserving linebreaks in Description field


jacobdgm:
with #1087, we made a change in the Source Detail template to preserve line breaks in the `description` field, as well as several other fields.

Some source `description`s include formatting via HTML tags, so this was causing more spacing than necessary. We asked Debra, and she said we should go with styling via HTML tags and not via line breaks.

When a fix for this issue goes to production, we need to let Debra know, so someone can go through and make sure all necessary `<p>`s and `<br>`s are included in source descriptions.

annamorphism:
Summarizing discussion from this morning's meeting:
It would be good to figure out why the line breaks are not always retained; possibly this has something to do with copy-pasting from Word (or other word processor hijinks.)
Having a solution where html tags have to be put into all entries in the future, and retroactively into a bunch of existing ones, is workable, but not ideal.

jacobdgm:
I haven't been able to figure out how to download Word, but I just tried a couple of experiments in Google Docs and Apple Pages to test whether pasting into `textarea`s introduces html tags:

In both OldCantus and New, copying and pasting from these applications does not lead to the introduction of html tags. Where bullet points exist in the copied text, they are converted into single lines without bullet points.

Unless Word behaves differently than Google Docs and Apple Pages, or unless different operating systems or browsers handle pasting differently, it seems likely that indexers have manually added these html tags to the descriptions of certain sources over the years.

(for what it's worth, I've used a bunch of different browsers and different operating systems over the years, and I've never seen html tags added when I pasted into a text field. I highly doubt this could account for the html tags we're observing in source descriptions)

annamorphism:
I don't think the html tags get added with the paste, but something is a little weird about copy pasting. e.g. these two look fairly similar in the editor:
<img width="448" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/175e5554-7a1f-4214-a264-8d2dd59568cc">
but the one copied from Word has a second line break in it (after "the project") when rendered: 
<img width="579" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/13eae466-6a5d-4854-b823-c424f327f123">

annamorphism:
actually, rephrase: I see some related problems about getting html tags inserted when pasting from Word or Outlook from some time back (around 2011). So maybe some sources got tags inserted (automatically) at the last big Cantus update, and since then it's been tag-less? 

[...]

Issue #1221, opened 2023-12-18
Reduce line spacing between bibliography entries


dchiller:
Debra would like less space between bibliography entries on the Source Detail page.

Here's what they look like now on New Cantus:

<img width="744" alt="image" src="https://github.com/DDMAL/CantusDB/assets/11023634/e5417baa-9ae9-4ec1-b7da-b3ba9da2de01">

And for reference, here's the spacing on Old Cantus:
<img width="654" alt="image" src="https://github.com/DDMAL/CantusDB/assets/11023634/03cd5741-5b17-4c4d-8b9d-2c43ac1c7ab9">

jacobdgm:
I think this was caused by #1087. Each `<li>` tag in the bibilography starts on its own line in the text that was entered. Since we're now preserving line breaks with `|linebreaks`, this causes django to add a `<br>` beneath each `<li>` element.

It seems that the formatting of the bibliographies of some sources (and some of their descriptions as well - see #1219) has been done with html tags, and for others, the formatting was done with line breaks. We can accommodate either case (either by including `linebreaks` or by omitting it), but we can't do both.

jacobdgm:
noting that this issue is similar to https://github.com/DDMAL/CantusDB/issues/1219, and that we can apply a similar approach (checking for html tags before applying the `linebreaks` decorator) here as we do there.

Issue #1219, opened 2023-12-18
GOTTSCHALK source description has weird spacing


dchiller:
Source detail page here: https://cantusdatabase.org/source/123609

I can't get it in a screenshot at normal sizing, but when I zoom out, you can see that there is a table that gets pushed all the way down the page: 

<img width="1321" alt="image" src="https://github.com/DDMAL/CantusDB/assets/11023634/2af9fcd4-cffc-4843-a5c9-c02d3ee11c7a">

dchiller:
In the html, there are many many `<br>` tags between the colon and the table definition. When those are manually removed, the spacing is normal. I suspect those are being added automatically somehow.

There is a single `<br>` tag at this point in the Source Description field, but removing it does not seem to fix the problem.

annamorphism:
Is that what it is! I've been annoyed at this table before...

jacobdgm:
This was introduced by #1087 - when I remove `linebreaks` locally, the extra spaces disappear.

I might suggest that we get rid of the `|linebreaks` "decorator" everywhere on the Source Detail page - it's causing this here problem, and it's causing problems in the Bibliography field (#1221). We may need to tell the musicologists to go through the text fields for all the sources and format everything via html tags, but we can't accommodate both text spaced out with linebreaks and text spaced out by html tags.

annamorphism:
but OldCantus seemed to handle this situation...
https://cantus.uwaterloo.ca/node/123609/edit
<img width="426" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/67bbfca7-a2b2-4d25-8103-69a568c61629">
rendered as 
<img width="598" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/817f7a34-c234-4172-b8d9-8e37de1706e4">

while this also works fine:
<img width="315" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/e77fc96c-3edd-4b96-8574-253e311c993a">
<img width="584" alt="image" src="https://github.com/DDMAL/CantusDB/assets/67451875/1e234902-8ddf-48d1-80b7-6ceb9c942f12">

dchiller:
> We may need to tell the musicologists to go through the text fields for all the sources and format everything via html tags, but we can't accommodate both text spaced out with linebreaks and text spaced out by html tags.

Or do this automatically, no?

annamorphism:
[...]

PR #1749, opened 2025-01-31
Add markdown-powered fields

dchiller:
This PR adds machinery to use (edit and display) markdown in text fields. This PR implements this on the `Source.selected_bibliography` and `Source.description` fields, but the same machinery can be used for other fields (e.g., flat page fields, future landing page fields).

The PR includes the following features:
1. a new `MarkdownWidget` django widget that can be used in forms. This widget provides both a `textarea` editing area and the ability to navigate via tabs to a "preview" area. This preview is generated client side (with the `marked.js` dependencies included by way of the widget definition. 
2. a new `render_markdown` template tag that converts the markdown to html and sanitizes the html on the server side for when fields containing markdown are displayed.
3. stylings for both `MarkdownWidget` and a new class `markdown-field-display` that styles the contents of markdown fields (note: in future, these stylings could be expanded to multiple classes that can be used to style fields generated from markdown differently in different contexts). 

Note: The `render_markdown` tag cleans the generated html to remove raw html included in the markdown. Thus, fields displayed with this template tag that have legacy raw html in them will need to be edited before they will display correctly. This editing should occur concurrently with the implementation of this PR on production. I realize that this limits the kinds of things we can do in these fields (we can't use markup alongside markdown), which may present a future problem; on the other hand, we may find that these limits are perfectly fine for our use of these fields.

Once this is implemented and relevant fields are edited, we will be able to deal with #1239, #1221, and #1219.

Screenshot of edit tab:

image

Screenshot of detail view:

image

### DIFF
--- a/django/cantusdb_project/main_app/templatetags/helper_tags.py
+++ b/django/cantusdb_project/main_app/templatetags/helper_tags.py
@@ -285,4 +285,6 @@ def render_markdown(value: str) -> SafeString:
     Renders markdown text as HTML.
     """
     html: str = github_flavored_markdown_to_html(value)
+    # Generated html is marked safe b/c cmark is run in safe mode in
+    # github_flavored_markdown_to_html
     return mark_safe(html)
--- a/django/cantusdb_project/main_app/templatetags/helper_tags.py
+++ b/django/cantusdb_project/main_app/templatetags/helper_tags.py
@@ -7,7 +7,8 @@ from django.db.models import Q
 from django.template.defaultfilters import stringfilter
 from django.utils.safestring import mark_safe, SafeString
 from django.http import HttpRequest, QueryDict
-from django.utils.html import format_html_join, format_html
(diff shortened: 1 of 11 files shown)


django/cantusdb_project/main_app/templatetags/helper_tags.py
```diff
@@ -276,3 +277,12 @@ def join_absolute_url_links(
...
     )
+
+
+@register.filter
+def render_markdown(value: str) -> SafeString:
+    """
+    Renders markdown text as HTML.
+    """
+    html: str = github_flavored_markdown_to_html(value)
+    return mark_safe(html)
```
dchiller, line 290: Realizing it would be useful if this had a comment...
```suggestion
    # Generated html is marked safe b/c cmark is run in safe mode in
    # github_flavored_markdown_to_html
    return mark_safe(html)
```

lucasmarchd01 (approved):

PR #1758, opened 2025-02-06
Add crediting for source descriptions

### LINKED ISSUE #1732: wanted: credit for descriptions
On the source edit page, we have ways of crediting the indexers, proofreaders, adders of melodies, etc. But we don't have a field to credit descriptions consistently! This should just be a field called "Description by" where a user (or in some cases maybe several  users) can be credited with the description. The user profiles could then also list descriptions written by that user.

lucasmarchd01:
Resolves #1732  

This pr enables credit attribution to users who edit the source description. It functions similarly to other fields in the source model, such as `full_text_entered_by` and `melodies_entered_by`.

- add `description_entered_by` field to source model
- add migration file
- update source edit and create forms to include this field
- update source create, source edit, and source detail pages to display this field
- update user detail page to credit descriptions
- add source description field to the indexing notes search

Source edit page:
Screenshot 2025-02-06 at 2 04 36 PM

User Detail Page:
Screenshot 2025-02-06 at 2 05 15 PM


django/cantusdb_project/main_app/forms.py
```diff
@@ -892,6 +900,14 @@ class Meta:
         ),
     )
 
+    # description_entered_by = forms.ModelMultipleChoiceField(
```
dchiller: Remove?

dchiller (approved):

Issue #1713, opened 2024-11-19
Source-folio-sequence uniqueness requirement


dchiller:
This requirement is currently tested in the `form_valid` method of the ChantCreateView. It is not tested at all for any other chant editing methods (either the Admin panel or the ChantEditView), and indeed, this method is the wrong place for this kind of validation (the validation should occur on the form and not on the view). Eventually, this would be a good place for a uniqueness constraint on the model itself, but there is messy data currently. 

We'll use #1714 to fix this data and then add a model uniqueness constraint. In the meantime, I'll move these validations to a form method.

annamorphism:
if I read this correctly, this would help prevent getting all the way through editing a chant only to get it bounced after saving because something about the automatically generated sequence number was weird--correct?

dchiller:
> if I read this correctly, this would help prevent getting all the way through editing a chant only to get it bounced after saving because something about the automatically generated sequence number was weird--correct?

Not exactly what I was intending with this. Because we currently just check this on the individual form, rather than the data level, there are duplicates in the data (either legacy data or edited through the taxonomy manager). So we should fix it so that no matter where the chant is edited, this requirement is met. 

If it would be helpful to have something that alerted you to this earlier, though, we could include such a feature in this issue.

annamorphism:
oh, ok, I just misunderstood! Yes, this should be checked at the data level. If it comes with a notification, that's just a bonus--I'm not really fussed about it.

annamorphism:
This should be part of the validation tests for proofreading, discussed in #1797

kyrieb-ekat:
Related to data cleaning tasks; in progress due to existing duplicates remaining inside records and the database. Issue is known and in progress of resolution. CFI Data cleaning (DACT org and repo has the details for this; this issue can stand as the anchor issue on the CDB end). @pouyamohseni

PouyaMohseni:
Closing this because the CDB-side detection is now covered by PR #2070's `check_duplicate_folio_sequence` check.

PR #1768, opened 2025-02-26
Create bulk chant edit function

### LINKED ISSUE #1714: Create workflow to easily edit aspects of multiple chants
Sometimes, edits are more easily made (correctly!) across multiple chants. Think here of renumbering folios or fixing issues with sequences. It would be helpful to see all chants on a folio at once.

### LINKED ISSUE #1769: Chant Edit: `c_sequence` is always marked as changed even if it isn't
This is because we are using a CharField, when the model field is an integer. So it's being treated as a change before being coerced to an integer.

dchiller:
This PR modifies the `SourceBrowseChantsView` to allow for bulk edits to certain chant fields.

Closes #1714. 

Authenticated users who are allowed to edit chants in a source see a new button on the browse chants page:

image

When clicked, this button shows a form for editing the displayed aspects of the chants. The chants in the form can be filtered using the browse chants page filters.

image

Changes made through this form are captured by django reversion.

Required fields are maintained, but for the moment, the folio-c_sequence uniqueness requirement is not. This is because we will use this form to fix folio-c_sequence uniqueness issues in the data and then institute a model-level constraint (per #1713).

Also closes #1769, replacing the use of a `CharField` for the `c_sequence` field on both the chant create and chant edit forms with an `IntegerField` to match the type of `c_sequence` on the `Chant` model.

Adds tests to support this feature.

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -214,7 +214,7 @@ class ChantCreateForm(forms.ModelForm):
         help_text="Binding order",
     )
 
-    c_sequence = forms.CharField(
+    c_sequence = forms.IntegerField(
         required=True,
         widget=TextInputWidget,
         help_text="Each folio starts with '1'.",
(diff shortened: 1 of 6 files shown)


lucasmarchd01 (approved):
Nice!

PR #1770, opened 2025-02-28
Create workflow to add chants to a source from a csv file.

### LINKED ISSUE #1259: We need to establish a workflow for creating chants given a CSV file
Debra recently gave us a .csv file with information for a bunch of chants that need created on cantusdatabase.org.

We need a system to create chants based on CSV files. Debra said that this used to happen on OldCantus, and will continue to need to happen from time to time in NewCantus.

**We could create a fully automated system with a specification** - "your file should include all these columns and exactly these columns", and so on. **It might, however, make sense to have a more flexible system** - perhaps a management command that can be adapted by a developer to accommodate whichever individual spreadsheets are sent to us by the musicologists. If we adopt this second approach, however, we will need to attend to these .csv files promptly.

Thoughts on how best to approach this?

dchiller:
Closes #1259. The new page can be found at `source/<source id>/add-chants`.

The page provides a preview of a selected CSV file, and various validations are made and signaled to the user with alerts and tooltips:

image

Additional validation occurs on the back-end.

Along the way, refactored or fixed the following:
1. The `JSONResponseMixin` that does content negotiation for a number of views had an error that caused it not to work nicely with `ListView`s that had more than one page. This fixes that error by making use of the `object_list` attribute, rather than the queryset passed to the `ListView` context.
2. `FeastListView` and `FeastDetailView` were not previously using the `JSONResponseMixin`, but it was useful for the front-end mapping and validation of feasts for this feature, so I added it.
3. Refactored the `AddImageLinksView` and the view for this PR into a new `source_bulk_actions` directory and added some typing to autocomplete views.
4. Previously, on the `ChantCreateForm`, we were validating the uniqueness of the sequence and folio of a new chant regardless of whether the field-level validation of the form had succeeded. But this meant that a form submitted with an empty folio or sequence (an invalid form) would still attempt to validate that the folio-sequence combination was unique, and would error because there was no value for that field. Now, we only complete the validation of the combination if the field values are all valid.


lucasmarchd01 (approved):

dchiller:
Thanks for reviewing @lucasmarchd01 .... in thinking about this, I should have made this a draft because I don't like the way this has validation in two different locations (front-end and back-end), so I am going to make some changes and ask you for a re-review.

lucasmarchd01:
Sounds good. I was also curious why you decided to implement this using a front end to the user instead of something like a management command? In the discussion it sounded like Andrew had some concerns with this. I thought it was fine so I approved regardless, but I was just wondering!

ahankinson:
Another option, in case you're looking for options, is to implement it as an Admin page "action" with intermediate pages.

https://docs.djangoproject.com/en/5.1/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages

This has a couple advantages:

 - Fewer styling / pages to administer. You just need to maintain the custom blocks in the admin
 - It's already restricted to people who have admin access, who I'm imagining will be the people uploading via CSV, so you can reduce most of your permissions code.

liampond:
Stale — Dylan said he would make changes and ask for a re-review, and we are working on a separate CSV import pipeline

PR #1791, opened 2025-03-13
New: Simplify source admin

ahankinson:
Previously the source admin page used a completely separate admin form that was handcrafted for each field. As best I can tell, this was only really used because the "Current editors" m2m select needed to be filtered to only project managers, editors, and contributors.

This PR removes the custom admin form and replaces it with the `ModelAdmin.formfield_for_manytomany` method, which allows you to pass in a custom QuerySet for specific m2m fields.

In addition, by reverting to the default admin behaviour this also makes the "autocomplete" widget available, so this has been added for the holding institutions and the provenance field.

For now I have left the custom admin form in place but have commented it out.

### DIFF
--- a/django/cantusdb_project/main_app/admin/source.py
+++ b/django/cantusdb_project/main_app/admin/source.py
@@ -1,8 +1,8 @@
 from django.contrib import admin
-
-from main_app.admin.base_admin import BaseModelAdmin, EXCLUDE, READ_ONLY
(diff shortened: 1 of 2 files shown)


ahankinson:
I should also mention that my linting tool, ruff, auto-organized the import statements to follow PEP8

dchiller (approved):
Thanks!

PR #1811, opened 2025-04-09
Minor fixes: template styles, code cleanliness/typing, sequence detail header, sequence list page

dchiller:
This PR implements relatively minor fixes to templates and does some code cleanup.
- explicitly exports models so linters stop complaining about imports directly from `main_app.models` (also removes a place where we had previously silenced this complaining)
- adds some vertical margins to the main content containers of a number of pages that was introduced in #1779 refactoring. These margins are now handled by templates inheriting from the base template, so we add a number of those in here.

On the template clean-up front, we:
- adjust the ordering of the sequence list page to be a little more rational. Add in ordering by source shelfmark and folio. 
- removes some verbiage from the header of the concordances table on the sequence detail page that made it seem like we were paginating results. Maybe we should paginate results, but we aren't currently, so now the header conforms more closely to what's actually happening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,7 @@ django/media/
 
 # MyPy
 .mypy_cache
+mypy.ini
 
 # Virtual environments
 .env
@@ -123,3 +124,6 @@ django/cantusdb_project/Drupal_scripts
 
 # cached files for APIs
 api_cache/
+
+
+.python-version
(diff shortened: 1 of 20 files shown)


dchiller:
Close because github wasn't showing new pushed commits....not sure what was going on but I reopened in #1815

PR #1815, opened 2025-04-10
Minor fixes: template styles, code cleanliness/typing, sequence detail header, sequence list page

dchiller:
This PR implements relatively minor fixes to templates and does some code cleanup.

- explicitly exports models so linters stop complaining about imports directly from main_app.models (also removes a place where we had previously silenced this complaining)
- adds some vertical margins to the main content containers of a number of pages that was introduced in #1779 refactoring. These margins are now handled by templates inheriting from the base template, so we add a number of those in here.

On the template clean-up front, we:

- adjust the ordering of the sequence list page to be a little more rational. Add in ordering by source shelfmark and folio.
- removes some verbiage from the header of the concordances table on the sequence detail page that made it seem like we were paginating results. Maybe we should paginate results, but we aren't currently, so now the header conforms more closely to what's actually happening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,7 @@ django/media/
 
 # MyPy
 .mypy_cache
+mypy.ini
 
 # Virtual environments
 .env
@@ -123,3 +124,6 @@ django/cantusdb_project/Drupal_scripts
 
 # cached files for APIs
 api_cache/
+
+
+.python-version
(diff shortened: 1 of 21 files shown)


ahankinson:
I've switched to `ruff` for style parsing as part of my code tooling. It replaces black and flake8. 

It also lets you configure it in the pyproject.toml file: https://github.com/rism-digital/muscatplus_server/blob/develop/pyproject.toml#L44-L77

(I've also switched away from Poetry to UV, and it's gone mostly smoothly. Needs a bit of muscle memory rework, but since Poetry 2 got rid of the `poetry shell` command, I would have had to re-learn a new way anyway.)

dchiller:
> but since Poetry 2 got rid of the `poetry shell` command

that's annoying...

lucasmarchd01 (approved):

PR #1818, opened 2025-04-10
feat: add source url model

### LINKED ISSUE #1789: New Model: Source URLs
In DIAMM we have a "Source URL" model: 

https://github.com/diamm/diamm/blob/main/diamm/models/data/source_url.py

This lets you attach different URLs of different types to a source record. It looks like this:

Image

The idea here is not to have individual fields for links on the source record, but as many different kinds of links as needed. 

So this would also mean migrating the content of, and then removing, the current "Image Link" field.

The motivation for this is to allow for capturing multiple IIIF manifest URLs to show so that users can view the images from multiple providers. I can write a separate ticket that describes this.

dchiller:
Create new model to store links related to sources.

Add new command `populate_source_urls` to move current `Source.image_link` fields to new `SourceURL` objects.

Create an inline admin for `SourceURL`'s on the `Source` admin page.

Closes #1789


django/cantusdb_project/main_app/migrations/0036_sourceurl.py
lucasmarchd01, line 1: Might need to re-run `makemigrations` to change the migration number after merging #1810

dchiller: yes, good point

lucasmarchd01 (approved):

Issue #1774, opened 2025-03-03
Complete move from `segment` to `segment_m2m` field


dchiller:
On production, the contents of the `segment` field have been mapped to `segment_m2m`. Now, we need to:

- make sure we aren't using the `segment` field anywhere
- get rid of the `segment` field
- rename segment_m2m to segment

kyrieb-ekat:
Still in progress, ref'd in PR #1774 but not fully implemented backwards.

PR #1820, opened 2025-04-11
fix: remove all uses of segment foreign key on `Source`

dchiller:
Refs: #1774.

This PR finished up previous work (in particular #1577) to move `Source.segment` field from a foreign key to a many-to-many relationship. This PR adjusts all remaining views that used the previous `segment` field to use the new `segment_m2m` field. 

The goal here is to merge and test this change and ensure that all existing functionality exists with the new field before removing the old field and renaming the `segment_m2m` field to be just `segment`.


django/cantusdb_project/main_app/admin/source.py
```diff
@@ -92,8 +91,14 @@ def get_queryset(self, request):
 
     def formfield_for_manytomany(self, db_field, request, **kwargs):
         if db_field.name == "current_editors":
-            kwargs["queryset"] = get_user_model().objects.filter(
-                Q(groups__name="project manager")
-                | Q(groups__name="editor")
-                | Q(groups__name="contributor")).distinct().order_by("full_name")
+            kwargs["queryset"] = (
```
lucasmarchd01: I'm guessing this might cause conflicts with #1812

dchiller: Probably, I left this here bc I wasn't sure what order we'd merge in...I'm thinking we should merge this before, update production, and then do a second update with just the permissions changes since that'll need a lot of testing and it's a big shift.

lucasmarchd01 (approved):

PR #1832, opened 2025-05-07
Restrict to Python Version 3.9

### LINKED ISSUE #1831: Restrict Python version to 3.11
- Python 3.9 reaches EOL in October 2025, while some project dependencies (e.g. psycopg-binary 3.1.18) seem incompatible with Python 3.13. 
- We should probably change the Python version in pyproject.toml and django/Dockerfile to Python 3.11.

SCN-MNG:
- Restrict to Python Version 3.9 because of psycopg-binary 3.1.18 being incompatible with Python 3.13.3
- Fix two formatting mistakes with Black.

Resolves #1831

### DIFF
--- a/django/cantusdb_project/main_app/admin/institution.py
+++ b/django/cantusdb_project/main_app/admin/institution.py
@@ -41,7 +41,13 @@ class InstitutionAdmin(BaseModelAdmin):
         "alternate_names",
         "migrated_identifier",
     )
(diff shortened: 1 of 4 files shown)


ahankinson:
Python 3.9 is EOL in October 2025. Better to pin to 3.10, 3.11, or 3.12.

SCN-MNG:
Good point, I'll pin the Python version to 3.11.

Issue #1833, opened 2025-05-14
Default Proofread Fields Status for Chants in Published Sources


lucasmarchd01:
Following the implementation of #1810, a new "other_fields_proofread" checkbox has been added to chants. This field currently defaults to False.

We need to clarify the expected behavior for chants belonging to published sources. If the assumption is that all fields in published sources are inherently proofread, then the "other_fields_proofread" status for these chants should be True (along with the other proofreading fields that have been missed).

A data migration or management command then is required to update the "other_fields_proofread" flag to True for all chants in existing published sources.

@annamorphism, am I correct with this assumption? I feel like I remember this coming up in some of our discussions.

annamorphism:
I think this sounds right.

dchiller:
Yes, I think after #1835 we should do this. I believe in discussions with Debra she was going to provide us with a list of sources for which this was true. I think you're right that this tracked previously published sources (and especially sources that were "published" before the web app days).

annamorphism:
re-opening this issue because I'm confused about it--it looks to me like there are a lot of published sources that are not marked as proofread. did this only happen for very select sources? was the command made but not actually run? am i misunderstanding what the intent was, or did we decide not to do it after all, or...?

annamorphism:
would still like to know if this command was run or not...

liampond:
Looks like it was never run. Would you like me to run it?

annamorphism:
If it could be run on staging and the we could sanity check the edge cases, that would be awesome.

PR #1834, opened 2025-05-15
Implement proofreading tracking and management

### LINKED ISSUE #1492: Proofreaders need some easier way to see their "task list" (chants that need proofreading)

### LINKED ISSUE #1799: Create Proofreading Dashboard for Debras
Add a new dashboard for Debras showing:
- Proofreading thermometer (percent complete)
    - Make Thermometer View Public (Optional)
- Chants and sources needing proofreading
- List of “dead” sources (inactive >3/6/12 months)

### LINKED ISSUE #1800: Create proofreading view for editors
Show editors a list of chants in their assigned sources that still need proofreading.

lucasmarchd01:
This pull request introduces a new proofreading feature to the Cantus Database, allowing for granular tracking of proofreading progress at the source and chant level. It includes a new `ProofreadingStats` model, a management command to calculate and update these stats, a user interface for project managers and editors to oversee the proofreading process, and various enhancements to existing forms and views.

*   **New `ProofreadingStats` Model:**
    *   Defines a new model to track proofreading progress for each source, including metrics for volpiano, manuscript full text, standardized spelling, and other fields.
    *   Includes a custom manager with a method to calculate and update these statistics.
*   **Management Command:**
    *   Adds a management command to trigger the calculation and update of proofreading stats for all sources.
*   **Admin Interface Integration:**
    *   Registers the `ProofreadingStats` model in the Django admin interface for easy management and review.
*   **Proofreading Overview Page:**
    *   Introduces a new page accessible to project managers and editors, providing a filtered list of sources that require proofreading.
    *   Includes filtering and sorting options based on source attributes and proofreading statistics.
*   **Chant Filtering:**
    *   Adds the ability to filter chants in the `SourceBrowseChantsView` based on the `other_fields_proofread` status.
*   **Signal Handling:**
    *   Implements signal handlers to automatically update `ProofreadingStats` whenever a chant is saved or deleted.
*   **Form Improvements:**
    *   Refactors forms to improve Latin text field validation and optimize formset handling.
*   **Testing:**
    *   Adds comprehensive unit tests for the `ProofreadingOverviewView` to ensure functionality and access permissions are working correctly.

**Benefits**

*   Provides a clear overview of proofreading progress across all sources.
*   Enables targeted proofreading efforts by identifying specific areas that need attention.
*   Improves data quality and consistency in the Cantus Database.
*   Enhances the user experience for project managers and editors involved in the proofreading process.

**Related Issues**

#1833, 
Resolves #1799
Resolves #1800
Resolves #1492

**Deployment Notes**

*   Run migrations to create the new `ProofreadingStats` model.
*   Execute the `update_proofreading_stats` management command to initially calculate and populate the statistics for all existing sources.

**Screenshots**

image
Note: We should probably address #1833, and make all published sources "proofread" by changing all the checkboxes to true.


dchiller:
The overview view, form stuff, and filtering all looks good to me.

The new table seems overly complicated to me. I just played around a bit and with the following SQL query was able to get a table of sources and the various chant counts you'd need to calculate the stats you store on the Proofreading Stats model (I just did volpiano here for simplicity, but you could extend in the same way to the other fields):

```sql
SELECT s.id as source_id,
    COUNT(c.id) as num_chants,
    COUNT(c.volpiano_proofread) FILTER( WHERE c.volpiano != '' AND c.volpiano_proofread = True) num_volpiano_proofread,
    COUNT(c.volpiano) FILTER (WHERE c.volpiano != '') num_volpiano
FROM main_app_source s LEFT JOIN main_app_chant c ON s.id = c.source_id 
GROUP BY s.id;
```

Result:

image


So I think you could just have query in the view and calculate these when loading that particular view. Then we're not updating on every chant save (which is something that I've learned in the course of trying to solve timeouts on the bulk edit and create forms is already quite cluttered) and only when someone actually wants that information.

Edit: I obviously didn't do this in the ORM, so it may be that all your Q statements in the custom manager are basically doing this under the hood. My main point is less about the query construction and more about the fact that if I can get this info for every source in a single SQL query, then it seems unnecessary to persist it in a new model that needs to be separately kept up-to-date.

lucasmarchd01:
Great, thanks for the suggestions!
The reason I went with this whole extra model for these calculations is because initially when I did set up these queries in the view, the loading time for this page was very long, or would sometimes time out. I'll go back and see if I missed something and play around with it, because I agree it would be much nicer if we didn't have to do it this way.

dchiller:
Yeah, this query was super quick. I wonder how much it's an ORM thing...obviously it can do a lot of optimizations, but there are probably edge cases (and maybe this is one of them?) where the query constructed by the ORM is just really inefficient when there's a more efficient straight SQL approach.

lucasmarchd01:
See #1835

PR #1847, opened 2025-05-27
Merge `staging` to `production`, 2 June 2025

dchiller:
- [x] Universal viewer implementation (#1848)
    - [x] IIIF viewer behaviour works for sources with associated IIIF manifest links
    - [x] Links to images (external and IIIF) work on source detail pages where appropriate
    - Not a blocker, but noticed #1856 for the future
- [x] Browse Chants page: bulk edit functionality does not time out (#1827)
    - Ah! This doesn't work because we need to update the ansible-managed docker-compose.yml file!  
- [x] Half-bar ("6") in volpiano aligns correctly with chant texts, except when part of a properly encoded missing music section (#1849)
- [x]  Chant search page sorting behaviour
    - [x] both directions of sorting show results with data first (null's are last) for fulltext, melody, and image link columns
    - [x]  feast, service, and genre columns are sorted by name, not ID
    - [x] feast sorting works

Also included in this update is #1850, an update of Django and python dependencies. Testing should ensure that containers build and some general pressure-testing of site features.


dchiller:
image

I notice that the icon for the UV tab says "Cantus Database" but has the UV icon. Is that intentional? I think it might be clearer if it had the Cantus Database logo...

@lucasmarchd01

PR #1848, opened 2025-05-27
Integrate Universal Viewer for IIIF Manifests

### LINKED ISSUE #1790: Integrate Universal Viewer for IIIF manifest viewing
Related to #1789, with one or more IIIF Manifest URLs a new set of links should appear on the Source record that allows users to view the IIIF images.

I have proposed a version of Universal Viewer (https://universalviewer.io) that is hosted on cantusdatabase.org, e.g., "http://cantusdatabase.org/viewer". 

Then when a user clicks one of the links to view the IIIF manifest, it opens a new tab to, e.g., 

https://cantusdatabase.org/viewer?manifest=https://url.to/manifest.json

We do this for RISM Online. See this record:

https://rism.online/sources/400090067

Clicking the "View Images" link on the record opens up this URL in a new tab:

https://rism.online/viewer.html#?manifest=https://iiif.rism.digital/manifest/ch/400090067.json

lucasmarchd01:
This pull request introduces the Universal Viewer (UV) to enable the viewing of IIIF images associated with Source records in the Cantus Database.

**Changes**:

* New `/viewer/` page using `uv.html` (standalone HTML with embedded UV).
* Viewer loads IIIF manifest from a `manifest` query param.
* Source detail page now shows IIIF links (with icon) for `SourceURL` entries of type *IIIF Manifest*.
* Clicking a link opens the Universal Viewer in a new tab.
* Added favicon and static asset path fixes.

**Testing**:

* Add a `SourceURL` with type "IIIF Manifest" and a valid manifest URL.
* Clicking the link should open the full-screen viewer and load the manifest properly.

Resolves #1790

Screenshot 2025-05-27 at 12 36 03 PM
Screenshot 2025-05-27 at 12 36 19 PM


django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -236,6 +237,19 @@ <h4>{{ source.short_heading }}</h4>
             {% if source.image_link %}
                 <a href="{{ source.image_link }}" class="guillemet" target="_blank">View images on external site</a>
             {% endif %}
+            {% for link in source.source_links.all %}
```
dchiller, line 237: Since you've added handling of all types of `SourceURL`s, I would get rid of 237-239 in this PR. Otherwise, if this is merged there will be duplicate links in the sidebar.

django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -236,6 +237,19 @@ <h4>{{ source.short_heading }}</h4>
...
+                {% if link.url_type == 1 %}
+                    <a href="{% url 'universal-viewer' %}?manifest={{ link.url }}"
+                       class="guillemet"
+                       target="_blank">{{ link.get_url_type_display }}
```
dchiller: Don't know if we wanna change the display text here or in the model choices, but better if this doesn't say IIIF Manifest. The url in `SourceURL` is to a manifest, but this link is not...it's just to a viewer with the images. 

"View image on Cantus Database" could work? As an alternative to the existing "View images on external site" (although honestly would change that one now that I think of it...).

django/cantusdb_project/main_app/urls.py
```diff
@@ -378,6 +379,12 @@
         SourceAddImageLinksView.as_view(),
         name="source-add-image-links",
     ),
+    # universal viewer
+    path(
+        "viewer/",
```
dchiller: Service-oriented url here... I understand why...

But maybe `source/abc123/viewer`?

django/cantusdb_project/main_app/templates/uv.html
```diff
@@ -0,0 +1,84 @@
...
+            }
+            gtag('js', new Date());
+
+            gtag('config', 'G-JXEY1VRT2W');
+        </script>
+    {% endif %}
+    {% if not PROJECT_ENVIRONMENT == "PRODUCTION" %}
+        {# to ensure Google/etc. only indexes the main production site, and not staging #}
+        <meta name="robots" content="noindex" />
+    {% endif %}
```
dchiller: We should do this via template inheritance or inclusion...

django/cantusdb_project/main_app/templates/uv.html
```diff
@@ -0,0 +1,84 @@
...
+        var $UV = document.getElementById("uv");
+
+        function resize() {
+        if ($UV) {
+            $UV.style.width = window.innerWidth + "px";
+            $UV.style.height = window.innerHeight + "px";
+        }
+        }
+        window.addEventListener("resize", resize);
+        resize();
```
dchiller: Why the separate script tags?

dchiller: And maybe more consistent with the rest of the site to do this with a js file?

dchiller:
This is nice!

ahankinson:
You don't really need to put your UV HTML file in Django -- you can just include it as a static file. 

Here's the `viewer.html` we use on RISM Online. 

https://github.com/rism-digital/muscatplus_ui/blob/develop/dist/viewer.html

Here's what it looks like: 

https://rism.online/viewer.html#?manifest=https://api.digitale-sammlungen.de/iiif/presentation/v2/bsb00093971/manifest&xywh=-567%2C-179%2C5120%2C3572

I don't think you need the resize handler, do you?

django/cantusdb_project/main_app/urls.py
ahankinson: `viewer` isn't a resource, though... 

I think some service-oriented URLs are fine, as long as you're not publishing them as a resource. 

See my comments below, but I think the viewer URL, including the manifest that it is viewing, should be:

`https://cantusdatabase.org/viewer.html?manifest=https://example.org/some/manifest.json`

With this you don't need to involve Django at all.

django/cantusdb_project/main_app/templates/uv.html
lucasmarchd01: Now that we are including the UV as a static file, do we still need the meta tags included in `base.html`? The reason I avoided inclusion or extension was because we don't want the footer or background images on this page.

Also, now the django template engine isn't involved so template tags won't work.

django/cantusdb_project/static/js/uv-loader.js
```diff
@@ -0,0 +1,12 @@
...
+
+if (manifestFromUrl !== undefined)
+{
+    const data = urlAdaptor.getInitialData({
+        manifest: manifestFromUrl,
+    });
+
+    uv = UV.init("uv", data);
+    urlAdaptor.bindTo(uv);
+}
```
ahankinson: This is so small, do you really need a separate JS file? Can't you include it in the uv.html file itself?

lucasmarchd01: I was following Dylan's suggestion to be more consistent with the rest of the site in keeping js files separate, but I agree maybe in this case it makes more sense to keep it in the html file.

django/cantusdb_project/main_app/templates/uv.html
dchiller: Do we want the google tag on the viewer page? That's for Google Analytics, right? Not sure we need it in that case. 

I agree we can get rid of this template now, and the point it moot.

dchiller (approved):
All good to me if we get rid of the Django template version of the page. I see point about size of separate js file, especially now without the resizer. I'm fine either way.

PR #1876, opened 2025-06-25
Remove segment restrictions for chant browsing and show links conditionally

### LINKED ISSUE #1845: some /chants pages 404?
I'm inspecting https://cantusdatabase.org/source/1000207 --a source I can see and edit just fine. I can tell from teh "view full inventory" that it contains one chant--which I can also view and edit just fine. 
BUT when I try to "View all chants", https://cantusdatabase.org/source/1000207/chants/ gives me a 404.  

Image

This doesn't seem like the right behaviour--it may be that I can't see them because the sources are not published, in which case it should be a permissions error (403?), and also I'd expect to be blocked from the whole source and not just the list of all chants.

lucasmarchd01:
- Remove CANTUS segment check in SourceBrowseChantsView
- Return 404 for sources with no chants in browse/inventory views  
- Only show "View all chants" and "View full inventory" links when source has chants
- Add has_chants context to SourceDetailView

Fixes #1845

### DIFF
--- a/django/cantusdb_project/main_app/views/source.py
+++ b/django/cantusdb_project/main_app/views/source.py
@@ -165,8 +165,7 @@ class SourceBrowseChantsView(CustomAccessMixin, ListView):  # type: ignore[type-
         source: Source = self.source
 
         # Check if source has any chants - if not, return 404
-        chants_in_source = source.chant_set
-        if chants_in_source.count() == 0:
+        if not source.chant_set.exists():
             raise Http404()
 
         context["source"] = source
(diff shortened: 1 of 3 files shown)


django/cantusdb_project/main_app/views/source.py
```diff
@@ -163,10 +163,10 @@ def get_queryset(self) -> QuerySet[Chant]:
...
         context: dict[str, Any] = super().get_context_data(**kwargs)
         source: Source = self.source
-        if not CANTUS_SEGMENT_ID in source.segment_m2m.values_list("id", flat=True):
-            # the chant list ("Browse Chants") page should only be visitable
-            # for sources in the CANTUS Database segment, as sources in the Bower
-            # segment contain no chants
+
+        # Check if source has any chants - if not, return 404
+        chants_in_source = source.chant_set
+        if chants_in_source.count() == 0:
```
ahankinson: You can probably use `exists()` here…

dchiller (approved):

PR #1885, opened 2025-07-25
Fix outstanding permissions errors/issues

dchiller:
This PR addresses some issues discovered during testing for #1881.

- Removes ability of non-superusers to modify assigned editors of sources after source creation
- Changes autocomplete for the `current_editors` field on the SourceCreate page to include all active users
- Add the link to the source edit page from the source detail page for standard users who created the source and are still assigned to the source

### DIFF
--- a/django/cantusdb_project/main_app/permissions.py
+++ b/django/cantusdb_project/main_app/permissions.py
@@ -161,6 +161,11 @@ class CustomAccessMixin(AccessMixin):
             return self.src_perm_cache.get(source.id)
         return self.check_user_assignment(source)
 
+    def user_created_source(self, source: Source) -> bool:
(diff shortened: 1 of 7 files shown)


django/cantusdb_project/main_app/permissions.py
```diff
@@ -161,6 +161,11 @@ def user_assigned_to_source(self, source: Source) -> bool:
             return self.src_perm_cache.get(source.id)
         return self.check_user_assignment(source)
 
+    def user_created_source(self, source: Source) -> bool:
+        if source.created_by == self.user:
```
ahankinson, line 165: `return source.created_by == self.user` does the same thing.

PR #1969, opened 2026-03-12
feat: add IIIF manifest-based image link mapping and fix CSV validation bugs

### LINKED ISSUE #1964: Add automatic folio to URI mapping with spreadsheet validation
From the 4 March 2026 meeting — I would appreciate it if someone could jump in if I got any of the details wrong for this.

Debra needs the ability to link external images to individual chant records on a per-folio basis, not just to the beginning of an image set. For sources with IIIF manifests, the system should automatically parse the manifest JSON to generate a folio-to-image URI mapping.

A spreadsheet should be generated from this mapping so that Debra-level users can review, validate, and make bulk edits before the mapping is applied.

Examples sources
- Ivrea Codex ([source/1000132](https://cantusdatabase.org/source/1000132) — IIIF manifest at `https://iiif.diamm.net/manifests/I-IV-LX/manifest.json`
- Tours 149 ([source/123640](https://cantusdatabase.org/source/123640))

### Requirements:

1. Parse IIIF manifest JSON to extract page/canvas URIs
2. Automatically generate a folio-to-image URI mapping
3. Produce a downloadable/editable spreedsheet of the mapping for Debra to edit/verify
4. Provide a UI button for Debra-level users to trigger the mapping and import the validated spreadsheet
5. Handle cases where a single chant spans multiple pages

### Related:

- #1778
- #940 
- #1728

liampond:
## Summary

Add IIIF manifest integration to the source image links page and fix several bugs in the CSV validation UI.

Closes #1964

### IIIF Manifest Integration
- New `iiif_utils.py` module with functions to fetch IIIF manifests, extract canvases, match them to source folios, and generate a CSV mapping
- New `SourceIIIFMappingView` that generates a downloadable folio-to-image CSV from a source's IIIF manifest
- "Generate IIIF Mapping CSV" button shown on sources that have an associated IIIF manifest
- Folio matching handles zero-padded numbers, various prefixes (`f.`, `fol.`, `Folio`), and recto/verso suffixes
- Regex anchored to prevent false positives (e.g. canvas labels like "Detail 2r" won't match folio `002r`)

### CSV Validation Bug Fixes
- Skip rows with empty folio fields from IIIF CSVs that include unmatched canvases, fixing phantom warnings (e.g. "the following folios do not exist in the source: ,,,,,,")
- Exclude empty strings from image link duplication checks
- Auto-detect semicolon delimiters for locale compatibility (e.g. French)
- Handle missing second column gracefully to prevent JS errors

### UX Improvements
- Add "Validation" header above the check results table
- Remove redundant "No image links duplicated" success text for visual consistency
- Show "Image links saved successfully!" message after saving

## Testing
- 43 unit tests for all IIIF utility functions and the mapping view in `test_iiif_utils.py`
- Tested against all 5 real IIIF manifests currently in CantusDB (all DIAMM, API v2)
- Tested label parsing and image URL extraction against Bodleian and e-codices manifests
- Verified false positive prevention with labels like "Detail 2r", "Image 5v", "Decoration", etc.

### DIFF
--- a/django/cantusdb_project/main_app/iiif_utils.py
+++ b/django/cantusdb_project/main_app/iiif_utils.py
@@ -1,0 +1,316 @@
+"""
+Utilities for parsing IIIF manifests and generating folio-to-image mappings.
+
+Supports both IIIF Presentation API 2.x and 3.0 manifests.
+"""
+
+import csv
+import io
+import re
+from dataclasses import dataclass
+
+import requests
+
+
+@dataclass
+class CanvasInfo:
+    """Represents a single canvas/page extracted from a IIIF manifest."""
+
+    label: str
+    image_url: str
+    canvas_index: int
+
+
+def fetch_manifest(manifest_url: str, timeout: int = 30) -> dict:
+    """
+    Fetch and parse a IIIF manifest JSON from the given URL.
+
+    Args:
+        manifest_url: URL to the IIIF manifest JSON.
+        timeout: Request timeout in seconds.
+
+    Returns:
+        Parsed manifest as a dictionary.
+
+    Raises:
+        requests.RequestException: If the fetch fails.
+        ValueError: If the response is not valid JSON.
+    """
+    response = requests.get(manifest_url, timeout=timeout)
+    response.raise_for_status()
+    return response.json()
+
+
+def _get_label_text(label: object) -> str:
+    """
+    Extract plain text from a IIIF label, handling both API 2.x and 3.0 formats.
+
+    API 2.x: label is a string or a list of strings.
+    API 3.0: label is a dict like {"en": ["Folio 1r"]}.
+    """
+    if isinstance(label, str):
+        return label
+    if isinstance(label, list):
+        # List of strings (API 2.x) or list of dicts
+        for item in label:
+            if isinstance(item, str):
+                return item
+            if isinstance(item, dict):
+                return item.get("@value", str(item))
+        return str(label[0]) if label else ""
+    if isinstance(label, dict):
+        # API 3.0: {"en": ["Folio 1r"], "none": ["f. 1r"]}
+        for values in label.values():
+            if isinstance(values, list) and values:
+                return str(values[0])
+        return ""
+    return str(label)
+
+
+def _extract_image_url_v2(canvas: dict) -> str:
+    """Extract the best image URL from a IIIF 2.x canvas."""
+    images = canvas.get("images", [])
+    if not images:
+        return ""
+    resource = images[0].get("resource", {})
+    # Try to get the IIIF Image API service URL
+    service = resource.get("service", {})
+    if isinstance(service, list):
+        service = service[0] if service else {}
+    service_id = service.get("@id", "")
+    if service_id:
+        # Construct a full image URL from the service
+        return f"{service_id.rstrip('/')}/full/max/0/default.jpg"
+    # Fall back to the resource @id (direct image URL)
+    return resource.get("@id", "")
+
+
+def _extract_image_url_v3(canvas: dict) -> str:
+    """Extract the best image URL from a IIIF 3.0 canvas."""
+    items = canvas.get("items", [])
+    if not items:
+        return ""
+    # Navigate: canvas -> AnnotationPage -> Annotation -> body
+    annotation_page = items[0]
+    annotations = annotation_page.get("items", [])
+    if not annotations:
+        return ""
+    body = annotations[0].get("body", {})
+    # Try service first
+    services = body.get("service", [])
+    if isinstance(services, dict):
+        services = [services]
+    for service in services:
+        service_id = service.get("id", service.get("@id", ""))
+        if service_id:
+            return f"{service_id.rstrip('/')}/full/max/0/default.jpg"
+    # Fall back to body id
+    return body.get("id", "")
+
+
+def extract_canvases(manifest: dict) -> list[CanvasInfo]:
+    """
+    Extract canvas information from a IIIF manifest.
+
+    Supports both Presentation API 2.x and 3.0.
+
+    Args:
+        manifest: Parsed IIIF manifest dictionary.
+
+    Returns:
+        List of CanvasInfo objects, one per canvas/page.
+    """
+    canvases = []
+
+    # Detect API version
+    context = manifest.get("@context", "")
+    is_v3 = "iiif.io/api/presentation/3" in str(context)
+
+    if is_v3:
+        canvas_list = manifest.get("items", [])
+        for i, canvas in enumerate(canvas_list):
+            label = _get_label_text(canvas.get("label", ""))
+            image_url = _extract_image_url_v3(canvas)
+            canvases.append(CanvasInfo(label=label, image_url=image_url, canvas_index=i))
+    else:
+        # API 2.x
+        sequences = manifest.get("sequences", [])
+        if not sequences:
+            return canvases
+        canvas_list = sequences[0].get("canvases", [])
+        for i, canvas in enumerate(canvas_list):
+            label = _get_label_text(canvas.get("label", ""))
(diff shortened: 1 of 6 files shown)


django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,321 @@
...
+        service = service[0] if service else {}
+    service_id = service.get("@id", "")
+    if service_id:
+        # Construct a full image URL from the service
+        return f"{service_id.rstrip('/')}/full/max/0/default.jpg"
+    # Fall back to the resource @id (direct image URL)
+    return resource.get("@id", "")
+
+
+def _extract_image_url_v3(canvas: dict) -> str:
```
ahankinson: Where you use this (and the v2 equivalent) you don't actually check whether the string has a value or not. 

I would change these to return `str | None` and then any place where you're returning an empty string now, return None. You can then catch that and handle it, rather than propagate the error.

django/cantusdb_project/main_app/urls.py
```diff
@@ -379,6 +380,11 @@
         SourceAddImageLinksView.as_view(),
         name="source-add-image-links",
     ),
+    path(
```
ahankinson, line 411: Have you checked this process for speed with large IIIF manifests? If it takes more than $TIMEOUT seconds to generate, then your user will never get to download the CSV.

liampond: Good thinking! I just tested it on Einsiedeln 121 (621 canvases), and the processing was under 5 milliseconds. The bottleneck is fetching, which is already capped at 30 seconds by `fetch_manifest`.

django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,321 @@
...
+        source_folios: List of folio strings from the source's chants.
+
+    Returns:
+        List of dicts with keys: 'folio', 'image_link', 'notes', 'canvas_label'
+    """
+    rows: list[dict[str, str]] = []
+    matched_folios: set[str] = set()
+
+    for canvas in canvases:
+        matched_folio = match_canvas_to_folio(canvas.label, source_folios)
```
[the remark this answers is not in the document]
liampond: Extracted `_build_folio_lookups()` as a separate function and added a `folio_lookups` keyword arg to `match_canvas_to_folio`. `generate_folio_image_mapping` now builds the lookups once and passes them through.

django/cantusdb_project/main_app/views/source.py
```diff
@@ -689,9 +701,96 @@ def get_initial(self) -> dict[str, Any]:
...
+                status=404,
+            )
+
+        # Fetch and parse the manifest
+        try:
+            manifest = fetch_manifest(manifest_link.url)
+        except requests.RequestException as e:
+            logger.exception("Failed to fetch IIIF manifest: %s", manifest_link.url)
+            return JsonResponse(
+                {"error": f"Failed to fetch IIIF manifest: {e}"},
```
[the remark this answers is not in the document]
liampond: Removed `{e}` from the user-facing message. The exception is still logged server-side via `logger.exception()`.

django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,321 @@
...
+    \s*
+    ([rv](?:ecto|erso)?)?        # Optional recto/verso suffix
+    """,
+    re.IGNORECASE | re.VERBOSE,
+)
+
+PAGE_PATTERN = re.compile(
+    r"^(?:p(?:age)?\.?\s*)(\d{1,4})",
+    re.IGNORECASE,
+)
```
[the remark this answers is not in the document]
liampond: Removed

django/cantusdb_project/main_app/views/source.py
```diff
@@ -689,9 +701,96 @@ def get_initial(self) -> dict[str, Any]:
...
+                status=502,
+            )
+
+        # Extract canvases from the manifest
+        canvases = extract_canvases(manifest)
+        if not canvases:
+            return JsonResponse(
+                {"error": "No canvases found in the IIIF manifest."},
+                status=404,
+            )
```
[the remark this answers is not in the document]
liampond: Replaced all `JsonResponse` error returns with `messages.error()` + `HttpResponseRedirect` back to the image links page. Updated tests to expect 302 redirects.

django/cantusdb_project/main_app/iiif_utils.py
liampond: Thanks for the feedback. I changed `_extract_image_url_v2` and `_extract_image_url_v3` to return `str | None`, and updated `CanvasInfo.image_url` to `str | None`. The `None` is converted to `""` at the CSV generation boundary in `generate_folio_image_mapping`.

liampond:
It looks like the test that failed was unrelated to this feature. PR #1971 should fix this.

PouyaMohseni:
@coderabbitai review

django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,364 @@
...
+        mapping: List of dicts from generate_folio_image_mapping.
+
+    Returns:
+        CSV string with headers: folio, image_link, notes, canvas_label
+    """
+    output = io.StringIO()
+    writer = csv.writer(output)
+    writer.writerow(["folio", "image_link", "notes", "canvas_label"])
+    for row in mapping:
+        # Only canvas_label is untrusted; folio/image_link/notes are our own
```
PouyaMohseni: Do you think `image_link should` get the same treatment as `canvas_label`? It's extracted straight from the remote manifest's service/resource/body id fields, so it's just as untrusted as the label; the comment here says it's "our own data," but I don't think that's quite right.

django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,364 @@
...
+        timeout: Request timeout in seconds.
+
+    Returns:
+        Parsed manifest as a dictionary.
+
+    Raises:
+        requests.RequestException: If the fetch fails.
+        ValueError: If the response is not valid JSON.
+    """
+    response = requests.get(manifest_url, timeout=timeout)
```
PouyaMohseni: Since this fetches an admin-stored URL server-side, would it be worth passing `allow_redirects=False` (or otherwise constraining where the request can end up)? Low risk here since only superusers set the manifest URL and only superusers trigger the fetch

PouyaMohseni: Might also be worth a size cap on the response before calling `.json()`

liampond: Going to skip this one: `raise_for_status()` doesn't fire on 3xx, so a manifest stored as `http://` (Bodleian 301s to HTTPS) would fall through to `.json()` and surface as "Failed to fetch IIIF manifest". There's no privilege boundary to protect either — the `is_staff_xnor_is_superuser` constraint means only superusers can set `SourceURL.url`, and this view is superuser-only, so they could just enter an internal address directly. Taking the size cap, though.

django/cantusdb_project/main_app/iiif_utils.py
```diff
@@ -0,0 +1,344 @@
...
+    response.raise_for_status()
+    return response.json()
```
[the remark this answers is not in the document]
liampond: See Pouya's comment below

django/cantusdb_project/main_app/iiif_utils.py
liampond: Yes, you're right. Implemented.

liampond:
@annamorphism would you like to and do you have the time to review this PR if I put it on staging? Or should we wait until Debra is available to review it?

annamorphism:
I think it should be good to put on staging.

PR #1971, opened 2026-03-12
fix(tests): make SourceListViewTest.test_ordering deterministic

### LINKED ISSUE #1970: Flaky test: SourceListViewTest.test_ordering fails intermittently on 'Order by country' subtest
The test `main_app.tests.test_views.test_source.SourceListViewTest.test_ordering` fails intermittently on the **"Order by country"** subtest. I ran the tests 100 times against the `develop` branch, and it failed 4/100 times. I discovered this in PR #1969. 

### Cause

The test creates sources with randomized fake data, including a deliberate duplicate country (lines 1333–1337). It then compares the order produced by Python's `sorted()` against the order returned by Django's `order_by` queryset.

When multiple sources share the same country, the siglum field becomes the tiebreaker. The Python sort key and the Django `order_by` handle siglum differently for private collectors:

- **Django** (`order_by`): always sorts by the raw `holding_institution__siglum` value from the database, unconditionally.
- **Python** (`sorted()` in the test): substitutes `""` for the siglum when `is_private_collector` is `True`.

This means that whenever a private collector happens to share a country with another source (which the randomized test data can produce), the two orderings disagree on where that source should appear, causing the assertion to fail.

liampond:
Fixes #1970

## Summary

- Fix flaky "Order by country" subtest in `SourceListViewTest.test_ordering` that failed ~4% of the time
- The Python sort key used `""` for private collector sigla, but Django's `order_by` uses the raw DB value (`NULL`). When a private collector shared a country with another source, the tiebreaker orderings diverged
- Align the Python sort key with PostgreSQL's `NULLS LAST` behavior and add a `pk` tiebreaker for full determinism
- Ran the test 100 times against the develop branch and it no longer fails

## Root cause

Private collectors have `siglum=NULL` (enforced by a model constraint). The test's Python sort key had an `is_private_collector` guard that substituted `""`, while Django's `order_by("holding_institution__siglum")` sorts by the actual DB value. With the deliberately-created duplicate country, the siglum became the tiebreaker — and the two orderings disagreed.

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views/test_source.py
+++ b/django/cantusdb_project/main_app/tests/test_views/test_source.py
@@ -1374,12 +1374,13 @@ class SourceListViewTest(CustomAccessTestMixin, TestCase):
                         else ""
                     ),
                     (
-                        source.holding_institution.siglum
-                        if source.holding_institution
-                        and not source.holding_institution.is_private_collector
-                        else ""
-                    ),
+                        source.holding_institution.siglum is None,
+                        source.holding_institution.siglum or ""
+                    )
+                    if source.holding_institution
+                    else (True, ""),
                     source.shelfmark,
+                    source.pk,
                 ),
             )
             self.assertEqual(list(expected_source_order), list(response_sources))


liampond:
From lab meeting: we should double- and triple-check that the test is actually failing. Maybe the test is working as intended and this is a rare bug that needs to be fixed.

kyrieb-ekat:
I don't think the test was failing accidentally, I think it was flagging those mismatches and errors correctly! Squeaky wheel, getting grease in this fix, shows the python-postgres logic mismatch.

kyrieb-ekat (approved):
looks good, the `(siglum is None, siglum or "")` tuple neatly mirrors PostgreSQL's `NULLS LAST` behavior, and the `pk` tiebreaker eliminates the remaining non-determinism. 

One small framing note: the test was actually catching a real inconsistency between the Python sort key and DB ordering, so this is a test correction rather than a production bug fix. Either way, the change is good and the test will be reliable going forward.

PR #1975, opened 2026-03-18
Feature/import cantorales

liampond:
## Summary

- Adds a data migration (`0042_import_cantorales`) that calls the `import_cantorales` management command during `manage.py migrate`
- This ensures the Cantorales sources are automatically imported on staging and production without needing to manually SSH in and run the command
- The migration delegates to the existing management command to avoid code duplication (I know you're not typically supposed to do this in Django because the management command might change and mess up the migration, but the migration already relies on the .csv, which itself could change

## Context

The `import_cantorales` management command and CSV were already merged, but the import only runs when the command is manually invoked. Deploying via ansible only runs `migrate` and `collectstatic`, so the new sources never appeared on staging. This migration closes that gap.


liampond:
Nevermind, I was able to modify staging without this workaround. Scrapping this PR.

PR #1991, opened 2026-04-01
feat: Add canadian chant database specific search page

kyrieb-ekat:
Introduces a dedicated chant search page for the Canadian Chant Database
(/CanadianChantDB/chant-search/) that scopes results to segment 4066 by
default while allowing users to expand their search to other Cantus
segments (Cantus DB, Sequence DB, Cantorales) via a checkbox dropdown.

Key additions and changes:

- New CcdbChantSearchView (ccdb.py): extends ChantSearchView with
  segment-scoped queryset logic; handles multi-segment searches by
  calling the parent's get_queryset() once per segment and unioning
  results (required because Django forbids .filter() after .union())

- New ccdb_chant_search.html: full CCDB-styled chant search page with
  tabbed results — 'Chants (N)' and 'Sources (N)' — allowing users to
  switch between result types without leaving the page; source tab runs
  a parallel keyword search across 7 source fields; empty source tab
  includes a tip that sources return from keyword searches

- Databases dropdown filter: Bootstrap checkbox dropdown letting users
  include additional segments in their search; defaults to CCDB only;
  button label reflects the number of selected databases

- Navbar Chant Search link now conditionally points to the CCDB chant
  search page when browsing CCDB pages (via ccdb_site context flag)

- Landing page search bar now submits keywords to ccdb-chant-search
  instead of the source browse page

- Home link with favicon icon added to the segment sidebar nav on all
  CCDB sub-pages

- Photo credit 'Salzinnes Antiphonal' overlaid on the hero banner

- CcdbMixin and CcdbLandingView wired into urls.py so that ccdb_site=True
  context flows to all CCDB pages including the landing page

### DIFF
--- a/django/cantusdb_project/main_app/views/ccdb.py
+++ b/django/cantusdb_project/main_app/views/ccdb.py
@@ -1,0 +1,13 @@
+from django.views.generic import TemplateView
+
+
+class CcdbTeamView(TemplateView):
+    """Project Team page for the Canadian Chant Database."""
+
+    template_name = "ccdb_team.html"
+
+
+class CcdbMapView(TemplateView):
+    """Map of Sources page for the Canadian Chant Database."""
+
+    template_name = "ccdb_map.html"
--- a/django/cantusdb_project/main_app/views/ccdb.py
+++ b/django/cantusdb_project/main_app/views/ccdb.py
@@ -1,13 +1,135 @@
+from typing import Any
+
+from django.db.models import Q, QuerySet
 from django.views.generic import TemplateView
 
+from main_app.views.chant import ChantSearchView
+from main_app.views.source import SourceListView
 
-class CcdbTeamView(TemplateView):
+
+class CcdbMixin:
+    """
+    Adds ccdb_site=True to template context for all CCDB views.
+    This allows the shared navbar to swap the Chant Search link to the
+    CCDB-specific chant search page when browsing the Canadian Chant Database.
+    """
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data(**kwargs)  # type: ignore[misc]
+        context["ccdb_site"] = True
+        return context
+
+
+class CcdbLandingView(CcdbMixin, SourceListView):
+    """Landing page for the Canadian Chant Database."""
+
+    pass
+
+
+class CcdbTeamView(CcdbMixin, TemplateView):
     """Project Team page for the Canadian Chant Database."""
 
     template_name = "ccdb_team.html"
 
 
-class CcdbMapView(TemplateView):
+class CcdbMapView(CcdbMixin, TemplateView):
     """Map of Sources page for the Canadian Chant Database."""
 
     template_name = "ccdb_map.html"
+
+
+AVAILABLE_SEGMENTS = [
+    (4066, "Canadian Chant Database"),
+    (4063, "Cantus Database"),
+    (4064, "Sequence Database"),
+    (4067, "Cantorales in the Americas and Beyond"),
+]
+
+_DEFAULT_SEGMENT_ID = 4066
+
+
+class CcdbChantSearchView(CcdbMixin, ChantSearchView):
+    """
+    Chant search page scoped to the Canadian Chant Database (segment 4066) by default.
+    Users may expand the search to additional database segments via the 'db' GET param.
+    Because ChantSearchView unions Chant and Sequence querysets before returning — and
+    Django forbids filtering a unioned queryset — multi-segment support works by calling
+    the parent's get_queryset() once per selected segment and unioning the results.
+    """
+
+    template_name = "ccdb_chant_search.html"
+
+    def _get_selected_segment_ids(self) -> list[int]:
+        """Return the segment IDs selected via 'db' GET params. Defaults to [4066]."""
+        raw = self.request.GET.getlist("db")
+        ids = [int(v) for v in raw if str(v).isdigit()]
+        return ids if ids else [_DEFAULT_SEGMENT_ID]
+
+    def get_queryset(self) -> QuerySet:
+        # If no search params, return empty queryset immediately (same as parent behaviour)
+        if not self.request.GET:
+            from main_app.models import Chant
+
+            return Chant.objects.none()
+
+        # Compute segment IDs before any GET manipulation so they survive the copy
+        self._segment_ids = self._get_selected_segment_ids()
+
+        # Make GET mutable
+        self.request.GET = self.request.GET.copy()
+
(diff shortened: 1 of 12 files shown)


django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,135 @@
...
+    template_name = "ccdb_team.html"
+
+
+class CcdbMapView(CcdbMixin, TemplateView):
+    """Map of Sources page for the Canadian Chant Database."""
+
+    template_name = "ccdb_map.html"
+
+
+AVAILABLE_SEGMENTS = [
```
ahankinson, line 41: Surely this is defined globally? It doesn't make sense for these to be held only in the ccdb view.

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,135 @@
...
+    (4066, "Canadian Chant Database"),
+    (4063, "Cantus Database"),
+    (4064, "Sequence Database"),
+    (4067, "Cantorales in the Americas and Beyond"),
+]
+
+_DEFAULT_SEGMENT_ID = 4066
```
ahankinson, line 48: This should also be globally accessible (settings?)

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,135 @@
...
+    def _get_selected_segment_ids(self) -> list[int]:
+        """Return the segment IDs selected via 'db' GET params. Defaults to [4066]."""
+        raw = self.request.GET.getlist("db")
+        ids = [int(v) for v in raw if str(v).isdigit()]
+        return ids if ids else [_DEFAULT_SEGMENT_ID]
+
+    def get_queryset(self) -> QuerySet:
+        # If no search params, return empty queryset immediately (same as parent behaviour)
+        if not self.request.GET:
+            from main_app.models import Chant
```
ahankinson, line 71: Why are you importing something inline here? This should go at the top of the file.

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,135 @@
...
+        self.request.GET = base_get
+
+        combined = querysets[0]
+        for qs in querysets[1:]:
+            combined = combined.union(qs, all=True)
+        return combined
+
+    def get_context_data(self, **kwargs: Any) -> dict:
+        context = super().get_context_data(**kwargs)
+        from main_app.models import Source
```
ahankinson, line 107: Imports go at the top of the file.

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,135 @@
...
+        if not self.request.GET:
+            from main_app.models import Chant
+
+            return Chant.objects.none()
+
+        # Compute segment IDs before any GET manipulation so they survive the copy
+        self._segment_ids = self._get_selected_segment_ids()
+
+        # Make GET mutable
+        self.request.GET = self.request.GET.copy()
```
ahankinson, line 79: I have no idea what this is doing, or why. AI code?

kyrieb-ekat:
Thank you @ahankinson , I merged this a hair too early it seems. I'll open a new PR to fix the things you're pointing out.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: Line 76 reads the `segment` parameter(s) out of `request.GET` and stores them before the GET dict is modified. It's done early because the lines that follow would overwrite `request.GET["segment"]` to inject specific segment IDs when delegating to the parent queryset. Stashing the original value first ensures the selected segments aren't lost during that manipulation.

Line 79 (shoutout Claude) was done because Django's `request.GET` is a `QueryDict` that is immutable by default. Trying to set `request.GET["segment"] = ...` directly on line 83 would raise an error. Calling `.copy()` returns a mutable version of the same dict, which can then be freely modified. The assignment replaces the immutable original with the mutable copy.

So, trying to capture user intent before any changes are made, and then the right segment values can be injected with the right segment value when calling the parent `get_queryset()`.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: Maybe something like a constants file, to import segment ID instead of having to hunt them all down in urls.py etc? This would be a larger CDB fix though. I'm still very new to Django, but my understanding is that Django settings are for deployment configuration(s), so because "segment ID 4066" is a fixed database record ID that's the same everywhere, it would/should belong elsewhere. I'll open an issue about this, if we ever change IDs for whatever reason this would make future work a lot easier.

ahankinson: https://github.com/DDMAL/CantusDB/issues/1993

ahankinson: It's already there. 

There is no rule that says Django settings are for deployment configuration. They're there for app config parameters. Yes, in general you should avoid using PKs in this way, but as far as I know these segment IDs are fixed database entries in all environments.

kyrieb-ekat: I understand, thank you!

django/cantusdb_project/main_app/views/ccdb.py
ahankinson: 1. You shouldn't be modifying the GET. That's the incoming parameters from the URL request. Using them as a "shared" storage object is incorrect. There's a reason it's immutable... 
2. I don't see any modification of the GET in any parent queryset? 
3. What problem are you actually trying to solve?

kyrieb-ekat: 1. got it, thank you
2. The parent reads segment directly from self.request.GET.get("segment") in two places (lines 570 and 645). The GET mutation was a workaround to inject a segment value that the parent would then pick up. Sort of treating request.GET as a side channel to pass data into the parent's logic.
3. Allowing the user to search/include segments in their search. I need the parent to filter by a specific segment, but the parent has no clean extension point for that, it only reads segment from request.GET. 

I'll fix this, and try giving the parent an extension point and give the parent an override. Multi segment selections will need some additional thought.

ahankinson: Typically what I do is get the parent queryset *first* and the modify it with the subclass code. Querysets can be chained, so if you need to modify the parent qs, you would do something like: 

```
qs = super().get_queryset()
qs = qs.filter(segment__id=1234)  # or whatever... 
...
return qs
```

PR #1997, opened 2026-04-08
fix: disallow /chants/ and /sources/ in robots.txt

liampond:
Googlebot was hammering `/chants/` and `/sources/` — the two highest-traffic
endpoints — because they were not listed in `robots.txt`. `/chants/` without a
`source` query parameter was also throwing `BadRequest` errors on every hit.

Added `Disallow: /chants/` and `Disallow: /sources/` to `robots.txt`.


liampond:
Wrong branch

Issue #1993, opened 2026-04-02
Use (and maintain) the segment IDs in settings


ahankinson:
Segment IDs are already in the settings.py file:

https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/cantusdb/settings.py#L215-L216

New segment IDs should be added here, and all places in the code should reference these. 

(e.g., these are not correct:

https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/main_app/views/source.py#L52-L53

https://github.com/DDMAL/CantusDB/blob/ed9025d381aecd891b85bc6c5c0f27afd4116262/django/cantusdb_project/main_app/tests/test_views/test_source.py#L290

https://github.com/DDMAL/CantusDB/blob/ed9025d381aecd891b85bc6c5c0f27afd4116262/django/cantusdb_project/main_app/management/commands/update_proofread_status.py#L5

etc.

kyrieb-ekat:
**NOTES**:
Two new segment IDs need to be added to <code>settings.py</code>:</p>
<ul>
<li><code>CCDB_SEGMENT_ID = 4066</code></li>
<li><code>CANTORALES_SEGMENT_ID = 4067</code></li>
</ul>
<p>Once added, the following Python files have hardcoded references that should be updated to use the settings constants:</p>

File | Issue
-- | --
main_app/views/source.py | Duplicate local CANTUS_SEGMENT_ID and BOWER_SEGMENT_ID constants (lines 51–52); hardcoded 4066 and 4067 in get_template_names() (lines 324–326)
main_app/views/chant.py | Hardcoded 4064 (line 884)
main_app/views/api.py | Hardcoded 4064 (line 105)
main_app/views/ccdb.py | All four IDs in AVAILABLE_SEGMENTS, _DEFAULT_SEGMENT_ID, and docstrings
main_app/urls.py | Hardcoded 4066 (lines 319, 325) and 4067 (line 346)
main_app/management/commands/update_proofread_status.py | Duplicate local CANTUS_SEGMENT_ID = 4063 (line 5)


Templates (<code>source_list.html</code>, <code>ccdb_browse.html</code>, <code>navbar_links.html</code>) also contain hardcoded IDs but cannot import Python constants directly, so those can be left as-is.</p>
[...]

PR #2001, opened 2026-04-16
feat: CCDB redesign — new landing page, sub-pages, and chant search

kyrieb-ekat:
## Summary

- Redesigns the Canadian Chant Database landing page (`canadian_chant_db.html`) with updated layout and styling
- Adds CCDB sub-pages: Team, Map, and a tabbed Chant Search with multi-database filter
- Adds `CcdbBrowseView`, `CcdbLandingView`, `CcdbTeamView`, `CcdbMapView`, and `CcdbChantSearchView` backed by a new `ccdb.py` views module
- Registers 4 new URL routes under `/CanadianChantDB/`
- Layout polish: consistent justification, compact footer sponsor logos, font/logo size fixes
- Cosmetic-only reformatting of `import_cantorales.py` and tests (black)

## Test plan

- [x] Visit `/CanadianChantDB/` — landing page renders correctly
- [x] Visit `/CanadianChantDB/browse/` — source browse works
- [x] Visit `/CanadianChantDB/team/` and `/CanadianChantDB/map/` — sub-pages load
- [x] Visit `/CanadianChantDB/chant-search/` — search returns results; multi-db filter works
- [x] Confirm CCDB navbar swaps Chant Search link to CCDB-specific search when `ccdb_site=True`
- [ ] Run existing test suite (`test_import_cantorales.py`)


kyrieb-ekat:
This version calls segments in a non-centralised style that will change once the segment fix goes through to address #1993 in a separate PR in the future.

ahankinson:
These PRs are very hard to review because they mix style changes with functional changes, and it's difficult to tell them apart. 

Can you please remove the style and formatting changes (i.e., no functional changes) from this and submit it as a new PR, and then base the functional changes off that?

ahankinson (changes requested):
Style changes should be separated from functional changes.

kyrieb-ekat:
As requested, the formatting/style changes have been extracted into a separate PR: #2002! This PR now contains only the functional changes (new views, templates, URLs, CSS), and is based on top of #2002. Once #2002 merges into develop, this PR's base will be updated to develop.

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
```diff
@@ -2,7 +2,7 @@
 {% load helper_tags %}
 {% load static %}
 {% block title %}<title>Canadian Chant Database</title>{% endblock %}
-{% block favicon %}<link rel="icon" href="{% static "ccdb-images/CCDB-favicon.png" %}" />{% endblock %}
+{% block favicon %}<link rel="icon" href="{% static 'ccdb-images/CCDB-favicon.png' %}" />{% endblock %}
```
ahankinson: This is a style change only and should be removed (and others like it)

ahankinson (changes requested):
Still contains style-only changes

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
kyrieb-ekat: The style-only change was a quote style switch (" -> ') inside a Django `{% static %}` template tag on the `{% block favicon %}` line of `canadian_chant_db.html`. The line content was otherwise identical. It appeared in two commits that both touched that file, so it slipped through when the branch was originally put together.

I've squashed both instances out so the favicon block line no longer appears in the diff at all. The remaining double -> single quote differences on the logo `` tags are embedded inside sections where the class names, layout, and surrounding HTML structure also change, so they can't be cleanly separated from the functional changes.

Also snagged a multi-line reformatting of the ccdb import in `urls.py`, and a reordering of constants and imports in `ccdb.py`. Both have been folded back into the commit that originally introduced those files, so the branch history is now clean.

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
```diff
@@ -16,314 +16,220 @@
...
+                    </p>
+                        <div class="row align-items-center justify-content-center text-center">
+                        <div class="col-md-auto">
                             <a href="https://dact-chant.ca/" target="_blank">
-                                <img class="img-fluid p-2 ccdb-dact-logo"
-                                    src="{% static "footer/dact-logo.png" %}"
-                                    alt="Digital Analysis of Chant Transmission Logo"
-                                    title="Digital Analysis of Chant Transmission" />
+                                <img class="img-fluid p-2 ccdb-small-logo"
+                                     src="{% static 'footer/dact-logo.png' %}"
```
ahankinson, line 49: More style-only changes.

I'm being picky because templates are the place in Django where changes that are the most likely to have negative effects on database queries (transient model queries). So it's really important to be able to review the changes closely, and not have to wade through double-to-single quote change blocks to see if any actual problems have slipped through.

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,131 @@
+from typing import Any
+
+from django.db.models import Q, QuerySet
+from django.views.generic import TemplateView
+
+from main_app.models import Chant, Source
+from main_app.views.chant import ChantSearchView
+from main_app.views.source import SourceListView
+
+AVAILABLE_SEGMENTS = [
```
ahankinson, line 11: Please set and use the values from the settings.py instead of hardcoding them in every view!!!

https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/cantusdb/settings.py#L215-L216

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,131 @@
...
+    (4066, "Canadian Chant Database"),
+    (4063, "Cantus Database"),
+    (4064, "Sequence Database"),
+    (4067, "Cantorales in the Americas and Beyond"),
+]
+
+_DEFAULT_SEGMENT_ID = 4066
```
ahankinson: !!!!!

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,131 @@
...
+    def get_queryset(self) -> QuerySet:
+        # If no search params, return empty queryset immediately (same as parent behaviour)
+        if not self.request.GET:
+            return Chant.objects.none()
+
+        # Compute segment IDs before any GET manipulation so they survive the copy
+        self._segment_ids = self._get_selected_segment_ids()
+
+        # Make GET mutable
+        self.request.GET = self.request.GET.copy()
```
ahankinson: I feel like we've had this conversation before... you shouldn't be modifying the `GET`!!!

ahankinson (changes requested):
Please take more care with AI. There are quite a few things here that we've already talked about.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: Hi Andrew, yes! I was aiming to address the segment hardcoding at large in the CDB repo in a separate PR to close out #1993 instead of here, which I tried to disclaim at the top of this PR in that first comment.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: In this case 4066 is for the Canadian Chant Database and thus default for the `ccdb.py`; the internal ref I was planning on being adjusted in the subsequent segment fix. *I'll spot add this one now and do a sweep later of segments elsewhere in the repo.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: We did talk about this in a previous (closed/made irrelevant and subsequently deleted) PR!

Mutating `request.GET` is a semantic violation even if `.copy()` sidesteps the immutability error as (if my understanding is right) the object no longer reflects what the user actually requested, which is a side effect that could cause snowballing issues later.

The reason I reached for it is because the parent reads `self.request.GET.get("segment")` directly in two places with no override hook, so I had no otherwise clean way to inject a segment value the parent would pick up.

I did look into the proposed queryset-chaining approach, but it looks like the GET mutation is load-bearing here. The parent applies segment filtering to `chant_set` and `sequence_set` separately before calling `.union()`, in both the `search_bar` and standard search branches:

```python
if segment_id := self.request.GET.get("segment"):
    chant_set = chant_set.filter(source__segment_m2m=segment_id)
    sequence_set = sequence_set.filter(source__segment_m2m=segment_id)
```

Once `.union()` is called Django forbids further filtering, so `super().get_queryset().filter(segment__id__in=...)` won't work, because the segment constraint has to be injected *before* the union happens. I'm still finding my feet in Django-land, so other solutions or workarounds probably exist...

The cleanest solution, I think now, would be adding a `get_segment_id()` hook to `ChantSearchView` that subclasses can override, which I think is best done in a follow-up PR if you'd prefer that over the current approach. Otherwise I think the GET mutation, while semantically inelegant, is the least bad option until a separate, targeted fix can be made upstream.

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
kyrieb-ekat: The quote difference on this `` tag is incidental to the class name change on the same element (`ccdb-dact-logo` to `ccdb-small-logo`), which is a functional layout change. They can't be cleanly separated in the diff.

django/cantusdb_project/main_app/views/ccdb.py
ahankinson: There seem to be a lot of open PRs that do a lot of similar things. I see now that this is a PR to a PR branch. This is very confusing.

django/cantusdb_project/main_app/views/ccdb.py
ahankinson: > I think is best done in a follow-up PR

PRs should be self-contained modifications that are ready to be reviewed and merged. If you have a solution it should be implemented. 

It's not simply "semantically inelegant." You are using the GET as shared state within the view classes, bypassing all the other forms of state sharing and chaining of queries. This can and will become a problem later, particularly if this becomes an established or acceptable pattern.

This is production code, so there is a fairly high quality bar. It's also one where there have been, and will continue to be, a fair number of developers whose time does not overlap on the project, meaning the code should follow best practices as and when it enters the flow towards production, rather than as later patches, since it's never guaranteed that that patch will ever come in, and [technical debt](https://www.ibm.com/think/topics/technical-debt) will accrue.

> Once .union() is called Django forbids further filtering... 

can you defer calling it until after the filtering?

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
ahankinson: The correct method is to ensure your files are formatted to your formatter's standard, and committing those changes, and then modifying the files with the functional changes from a "clean" working tree.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: The request.GET mutation has been replaced in commit `a9d01388`!

The fix adds a `get_segment_id()` hook to `ChantSearchView` (defaulting to `self.request.GET.get("segment")`) and replaces both direct GET reads with calls to it. `CcdbChantSearchView` overrides the hook to return `self._current_segment_id,` an instance variable set immediately before each `super().get_queryset()` call. `request.GET` is never touched. The multi-segment case (calling the parent once per segment, then unioning) works identically, with the only difference is the segment is now injected via the hook rather than by writing into GET.

`CCDB_SEGMENT_ID = 4066` has also been added to settings.py and the hardcoded values in ccdb.py and urls.py replaced accordingly. The remaining 4067 (Cantorales) in AVAILABLE_SEGMENTS will be handled in the planned segment sweep PR.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: The branch-on-branch decision whackiness came from my iterating on the layout approach while the formatting cleanup was still in flight. Layout2 started as a variant of layout1, and then the PR got carved out mid-stream. Layout1 was rebased directly onto develop once it was stable, but layout2 branched before that settled. [#2002](https://github.com/DDMAL/CantusDB/issues/2002) was ok'd to merge, so this can be rebased onto develop and the history will be clean... and I won't do branching like this again, oof.

PR #2008, opened 2026-04-25
fix: remove trailing slashes from exported chant/source URLs

### LINKED ISSUE #1999: Bug: Exported chant and source URLs contain trailing slashes resulting in 404 errors
## Description
Cantus Database (CD) is currently exporting chant and source links with a trailing slash (`/`). However, the CD server does not handle URLs with trailing slashes correctly and returns a 404 Not Found error instead of loading the page or redirecting to the non-slash URL. 

While a temporary workaround was previously implemented on the Cantus Index (CI) side, this needs to be resolved on the CD side as it is now breaking functionality in other tools.

## Examples
**Broken (Currently Exported):**
* `https://cantusdatabase.org/chant/383601/`
* `https://cantusdatabase.org/source/123707/`

**Working (Expected/Correct):**
* `https://cantusdatabase.org/chant/383601`
* `https://cantusdatabase.org/source/123707`

liampond:
fixes #1999 

## Summary

Chant and source URLs written by `update_cached_concordances.py` included a trailing slash (e.g. `/chant/123/`). External tools that consume these exported links — including Cantus Index — were receiving 404s because CantusDB's URL patterns do not include the trailing slash.

- Removes the trailing slash from `chant_uri` and `source_uri` in `update_cached_concordances.py`
- Adds an nginx `rewrite` rule to 301-redirect `/<type>/<id>/` → `/<type>/<id>` for backward compatibility with already-distributed slash URLs
- Updates test assertions in `test_functions.py` to match the corrected URL format

## Notes for reviewers

The nginx config change is in `cantusdb.conf.development`. Developers pulling this branch should re-copy it to `cantusdb.conf` and reload nginx.

## Test plan

- [ ] Run `UpdateCachedConcordancesCommandTest` — assertions now expect no trailing slash
- [ ] Verify that a request to `/chant/<id>/` (with slash) is 301-redirected to `/chant/<id>` in the dev environment
- [ ] Confirm that Cantus Index or another consumer can resolve the exported URLs without error

### DIFF
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -5,10 +5,6 @@
 server {
     listen 80;
 
-    # Strip trailing slashes from chant/source detail URLs:
-    # redirect /chant/<id>/ to /chant/<id> and /source/<id>/ to /source/<id>
-    rewrite ^/((chant|source)/\d+)/$ /$1 permanent;
-
     location = /robots.txt {
         alias /robots.txt;
     }
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -5,6 +5,9 @@
 server {
     listen 80;
 
+    # Strip trailing slashes from detail URLs: redirect /chant/123/ to /chant/123
+    rewrite ^(.*/\d+)/$ $1 permanent;
+
     location = /robots.txt {
         alias /robots.txt;
     }
--- a/config/nginx/conf.d/cantusdb.conf.development
+++ b/config/nginx/conf.d/cantusdb.conf.development
@@ -5,8 +5,9 @@
 server {
     listen 80;
 
-    # Strip trailing slashes from detail URLs: redirect /chant/123/ to /chant/123
-    rewrite ^(.*/\d+)/$ $1 permanent;
+    # Strip trailing slashes from chant/source detail URLs:
+    # redirect /chant/<id>/ to /chant/<id> and /source/<id>/ to /source/<id>
+    rewrite ^/((chant|source)/\d+)/$ /$1 permanent;
 
     location = /robots.txt {
         alias /robots.txt;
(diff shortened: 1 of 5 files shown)


ahankinson:
Did you try the Django APPEND_SLASHES setting I linked to in the issue? That takes care of the 301.

liampond:
> Did you try the Django APPEND_SLASHES setting I linked to in the issue? That takes care of the 301.

Unless I'm missing something, I believe `APPEND_SLASHES` adds a slash if the URL does _not_ contain a trailing slash. The issue here goes the opposite direction — we have a trailing slash that we would like to get rid of.

ahankinson:
`APPEND_SLASHES` tells Django to treat `foo/id` and `foo/id/` as the same. 

`curl -v "https://www.diamm.ac.uk/sources/117" > /dev/null` -> 301 to `/sources/117/`

liampond:
[Django's documentation](https://docs.djangoproject.com/en/6.0/ref/settings/#append-slash) says "if the request URL does not match any of the patterns in the URLconf and **it doesn’t end in a slash**, an HTTP redirect is issued to the same URL with a slash appended." I don't see anything that says what happens if it **does** end in a slash. 

What we want is `curl -v "https://www.diamm.ac.uk/sources/117/" > /dev/null` -> 301 to `/sources/117`.

ahankinson:
> However, the CD server does not handle URLs with trailing slashes correctly and returns a 404 Not Found error instead of loading the page or redirecting to the non-slash URL.

Your site should not care if the client appends, or doesn’t append, the slash. You can try to control the URLs you publish to one API or another, but this won’t fix the 404 issue if some other client appends it.

The fix is not to strip all slashes, the fix is to treat slash and non-slash requests as supported.

liampond:
Alright, I'll do that. Thanks for the feedback.

PouyaMohseni (approved):
Looks good. Could you squash the commits before merging to keep the commit history clean?

ahankinson:
I don't actually see the changes added to the settings.py? 

https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/cantusdb/settings.py

liampond:
I believe `APPEND_SLASH` is set to `True` by default. Is there another change for `settings.py` that you had in mind?

ahankinson:
Yeah, you’re right. So why were you getting a 404 with only a slash as a difference?

liampond:
Because the URL **with** a slash was undefined, giving a 404. `APPEND_SLASH` only appends, it does not try removing slashes.

PR #2017, opened 2026-04-25
fix: re-enable volpiano display for MS 73 (source 680970) in templates

kyrieb-ekat:
## Summary

- Re-enables volpiano melody display for source 680970 (MS 73) across all affected templates
- Display was temporarily disabled in #1635; this resolves #1893
- Each changed location retains a comment referencing both issues for traceability

## Files changed

- `browse_chants.html` — volpiano notation in chant rows
- `chant_detail.html` — volpiano notation field and "Melody with text" section (2 spots)
- `chant_search.html` — row-suppression wrapper removed; volpiano indicator restored
- `source_lists/source_list.html`, `cantorales.html`, `canadian_chant_db.html` — melody count display

## Test plan

- [ ] Browse chants for MS 73 (source 680970) while logged out — confirm volpiano notation renders
- [ ] Open a chant detail page for a chant in MS 73 with volpiano — confirm notation and "Melody with text" both appear
- [ ] Source list page — confirm melody count shows for MS 73
- [ ] Run `python manage.py test main_app`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1635, opened 2024-09-20
Temporary fix: don't display volpiano for MS 73


dchiller:
Since there is currently no mechanism for displaying parts of chant information depending on proofreading status, we'll temporarily turn off volpiano display for MS-73 while it is being entered/proofread.

PR #2018, opened 2026-04-25
fix: re-enable MS 73 (source 680970) volpiano display in templates and melody search

### LINKED ISSUE #1893: turn melodies for MS 73 back on
Now that all melodies in this source have been proofread, we should revert  #1635 so the melodies are public and viewable!

kyrieb-ekat:
Closes #1893

Re-enables source 680970 (MS 73) which was temporarily disabled in #1635.

This PR combines the fixes from both #2017 and #2018 (per reviewer request):

**Backend (views):**
- Removed `.exclude(source__id=680970)` from the `ajax_melody_search` queryset in `api.py`
- Removed the early-return block in `ChantSearchMSView.get_queryset` that was skipping melody search for source 680970
- Removed the `context["melodies"]` assignment in `ChantSearchView.get_context_data` that was added only to support a template guard

**Frontend (templates):**
- Removed all `chant.source.id != 680970` conditional checks from template `{% if %}` statements in `browse_chants.html`, `chant_detail.html`, and `chant_search.html`
- Removed the row-suppression wrapper in `chant_search.html`
- Restored melody count display in `source_list.html`, `cantorales.html`, and `canadian_chant_db.html`

### DIFF
--- a/django/cantusdb_project/main_app/templates/browse_chants.html
+++ b/django/cantusdb_project/main_app/templates/browse_chants.html
@@ -154,8 +154,8 @@
                                 <p>
                                     {{ chant.manuscript_full_text_std_spelling|default:"" }}
                                     <br />
-                                    <!-- See #1635. Temporarily disable volpiano display for this source. -->
-                                    {% if chant.volpiano and chant.source.id != 680970 %}
+                                    <!-- Volpiano display for source 680970 (MS 73) was temporarily disabled in #1635; re-enabled in #1893. -->
+                                    {% if chant.volpiano %}
                                         <span style="font-family: volpiano; font-size:25px">{{ chant.volpiano|default:"" }}</span>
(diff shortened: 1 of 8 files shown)


PouyaMohseni (changes requested):
~~This looks good but in `search_chant.html` you should also remove the hardcoded blocks from the original [#1635](https://github.com/DDMAL/CantusDB/issues/1635) workaround.~~

I see you have a separate PR for that (#2017). Since both PRs are resolving the same issue, I would suggest closing #2017 and bringing those changes into this PR, so the fix is complete in one place and nothing is left behind.

PouyaMohseni (approved):
Looks good!

kyrieb-ekat:
Merged to Develop; on deck to staging then prod (11 may)

Issue #2012, opened 2026-04-25
From the source page, we should be able to find sources that have no indexed chants


kyrieb-ekat:
i.e., column header sorting for the `Chants/Melodies` column head. 

(?) if no chants indexed, do they jump to the top as null?

annamorphism:
wait, IS this a duplicate of #1761? or were we talking there about cases where the inventory is missing an MS full text?

PR #2028, opened 2026-05-02
`Image Link` and `Chants/Melodies` column sorting to source list page

### LINKED ISSUE #2011: Allow table sorting on source page by images
Akin to how sorting by melody and all items with melodies jump to top; do same with images. 

Column sorting _only_.

PouyaMohseni:
fixes #2011

### DIFF
--- a/django/cantusdb_project/main_app/views/source.py
+++ b/django/cantusdb_project/main_app/views/source.py
@@ -522,6 +522,9 @@ class SourceListView(CustomAccessMixin, ListView):  # type: ignore[type-arg]
                 f"{sort_prefix}holding_institution__siglum",
                 f"{sort_prefix}shelfmark",
             ]
+        elif order_param == "num_chants":
+            # NULL number_of_chants sorts last when ascending
+            order_by_args = [f"{sort_prefix}number_of_chants"]
         else:
             order_by_args = [
                 f"{sort_prefix}holding_institution__siglum",
--- a/django/cantusdb_project/main_app/templates/source_lists/source_list.html
+++ b/django/cantusdb_project/main_app/templates/source_lists/source_list.html
@@ -132,11 +132,7 @@
                     <th scope="col" class="text-wrap" style="text-align:center">Summary</th>
                     <th scope="col" class="text-wrap" style="text-align:center">Date/Origin</th>
                     <th scope="col" class="text-wrap" style="text-align:center">Image Link</th>
-                    <th scope="col" class="text-wrap" style="text-align:center">
-                        Chants
-                        <br />
-                        /&nbsp;Melodies
-                    </th>
+                    {% sortable_header request "num_chants" "Chants<br />/&nbsp;Melodies" %}
                 </tr>
             </thead>
             <tbody>


liampond (changes requested):
When I click to sort by 'Chants / Melodies' on the right, sort by descending doesn't work (the first result has 0 chants).

Image

liampond:
Sorting by Image Link appears to sort the results in alphabetical order of the image URL. Why don't we make the sorting the same as what exists on the site (within the groups with and without images)?

debralacoste:
I'm on staging right now and I don't see the red headers on either of these columns. I see one above these comments within this thread, though. What am I missing?

annamorphism:
> I'm on staging right now and I don't see the red headers on either of these columns. I see one above this comments within this thread, though. What am I missing?

well, I don't see them either, if it's any consolation...
image

PouyaMohseni:
Sorry, this was initially in a wrong column; it is still under review.

PouyaMohseni:
@coderabbitai review

liampond (changes requested):
Please fix the issues from Black and rebase, then I'll approve.

PouyaMohseni:
It is now rebased and you can review.

django/cantusdb_project/main_app/tests/test_views/test_source.py
```diff
@@ -1929,6 +1929,82 @@ def test_ordering(self) -> None:
             self.assertEqual(
                 list(reversed(expected_source_order)), list(response_sources_reverse)
             )
+        with self.subTest("Order by has_image"):
+            src_with_img = make_fake_source(image_link="https://example.com/img.jpg")
+            src_no_img = make_fake_source(image_link=None)
+            src_empty_img = make_fake_source(image_link="")
```
liampond: These two subtests create six new sources partway through `test_ordering`, after the earlier subtests have already compared the response against the locally built `sources` list — so they only pass because they happen to run last, and appending another subtest below would break the ones above. Could you move the fixtures to the top of the method, or split these two out into their own test?

django/cantusdb_project/main_app/tests/test_views/test_source.py
```diff
@@ -1929,6 +1929,82 @@ def test_ordering(self) -> None:
...
+                self.assertIn(src_with_img.id, result_ids)
+                self.assertIn(src_no_img.id, result_ids)
+                self.assertIn(src_empty_img.id, result_ids)
+
+                if expect_images_first:
+                    # Once we see a source without an image every subsequent
+                    # source must also lack one.
+                    seen_no_image = False
+                    for s in result:
+                        if not s.image_link:
```
liampond: These checks assert on `s.image_link`, the same field the sort key reads, so they'd still pass even if every `SourceURL`-backed gallery sorted into the wrong group. A source built with only an `EXTERNAL_IMAGES` `SourceURL`, asserted on `external_images_url`, would catch it.

django/cantusdb_project/main_app/views/source.py
```diff
@@ -678,7 +689,31 @@ def get_queryset(self) -> QuerySet[Source]:
...
+                _has_image_sort=Case(
+                    When(
+                        image_link__isnull=False,
+                        image_link__gt="",
+                        then=Value(True),
+                    ),
+                    default=Value(False),
+                    output_field=BooleanField(),
+                )
+            )
```
liampond: The cell in this column renders `source.external_images_url`, which prefers an `EXTERNAL_IMAGES` `SourceURL` and only falls back to `image_link` — but this `Case` reads `image_link` alone, so on my dev database 20 of the 21 sources that display an "Images" link sort into the no-image group (just 1 source still has the legacy field set). Worth keeping the `Case`/`When` wrapper when you widen the condition: a bare `Q(image_link__gt="") | Exists(...)` comes back NULL rather than False for those rows, and since the first click maps to `.desc()` Postgres puts NULLs first, which inverts the sort.

django/cantusdb_project/main_app/views/source.py
```diff
@@ -700,6 +735,11 @@ def get_queryset(self) -> QuerySet[Source]:
                 f"{sort_prefix}shelfmark",
                 f"{sort_prefix}id",
             ]
+        elif order_param == "num_chants":
+            if sort_desc:
+                order_by_args = [F("number_of_chants").desc(nulls_last=True), "id"]
+            else:
+                order_by_args = [F("number_of_chants").asc(nulls_last=True), "id"]
```
liampond: `number_of_chants` is only written by `update_source_chant_count` on chant save or delete, so a source that has never had a chant stays NULL rather than 0 — on my dev database that's 111 of 126 published sources, and none hold a literal 0. With `nulls_last=True` on both directions those 111 can't be brought to the top either way, which is what #2012 asks for, and the column prints them as "0" regardless via `default_if_none`. Could you also use the same siglum/shelfmark/id tiebreakers as the `has_image` branch here — chant counts tie constantly, so `id` alone shuffles the rest of the page into database order.

liampond (changes requested):
This should also fix #2012, maybe link it?

liampond (approved):

PR #2031, opened 2026-05-03
docs: add CLAUDE.md and ignore personal Claude Code config

### LINKED ISSUE #2029: Add a CLAUDE.md file to the repo
Since we are using Claude Code, the [best practices document](https://code.claude.com/docs/en/best-practices) suggests that we should have a CLAUDE.md file to tell Claude about the repo each time a new session is made. Each developer should make a separate gitignored version with their own personal preferences.

liampond:
I would appreciate everyone's input on this file. I've tried to follow the [Claude Code best practices guide](https://code.claude.com/docs/en/best-practices) as much as possible, but please let me know if there's inaccurate, missing, or unnecessary information, or if certain points are better suited for the personal `CLAUDE.local.md`. 

### Summary

- Adds `CLAUDE.md` at the repo root with team-shared context for Claude Code
- Adds `CLAUDE.local.md` and `.claude*` to `.gitignore` so each developer can keep personal Claude Code preferences
- Content is scoped to things Claude can't reliably infer by reading the code itself (e.g., commands, gotchas, conventions, etc.)

Closes #2029.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,86 @@
+# Cantus Database
+
+Django app for the Cantus Database — a digital catalogue of medieval Latin chant manuscripts. Built and maintained by DDMAL at McGill.
+
+## Stack
+
+Django + Postgres + Redis/Celery, managed with Poetry. `django-reversion` tracks model history; `volpiano-display-utilities` is a custom DDMAL dependency installed from git. See `pyproject.toml` for versions.
+
+## Local development
+
+Docker Compose is the only supported local setup — bare `manage.py runserver` is not used.
+
+```bash
+docker compose up
+```
+
+## Tests
+
+Tests run inside the dev `django` container. Make sure the dev stack is up (`docker compose up -d`), then:
+
+```bash
+docker compose exec -T django python manage.py test main_app.tests
+```
+
+Run a single module or class to keep the loop tight:
+
+```bash
+docker compose exec -T django python manage.py test main_app.tests.test_functions
+```
+
+Uses Django's built-in test runner, **not pytest**. Exec'ing into the existing dev container is faster than spinning up a clean-room run.
+
+## Verifying changes
+
+Beyond running tests:
+
+- `docker compose exec -T django python manage.py check` — fast sanity check after settings, model, or URL changes.
+- `docker compose exec -T django python manage.py makemigrations --dry-run` — confirms whether a migration is needed after model changes.
+- UI/template changes can't be verified from the terminal — say so explicitly rather than claiming success.
+
+## Lint, format, types
+
+**No pre-commit hook is configured** — run `black`, `mypy`, `pylint`, and `djlint` yourself before pushing.
+
+## Branches and PRs
+
+Three long-lived branches: `production`, `staging`, and `develop`. PR against `develop`.
+
+Branch names: `feat/<topic>` or `fix/<topic>`.
+
+Commit messages: Conventional Commits (`feat:`, `fix:`, `refactor:`, `docs:`, `chore:`).
+
+## Architecture
+
+- `users/` defines a custom `Group` and `GroupMembership` — **distinct from `django.contrib.auth.Group`, which is unregistered.** Don't confuse the two.
+- `RevisionMiddleware` from `django-reversion` is active, so most model writes are tracked automatically.
+
+## User permissions
+
+Five access tiers, enforced via `main_app/permissions.py` (`CustomAccessMixin`, `get_sources_visible_to_user`):
+
+- **Anonymous** — sees only `published=True` sources.
+- **Logged-in user (no group)** — published sources plus any sources they're explicitly assigned to via `Source.current_editors`. Can edit those assigned sources.
+- **`editor` group** — broad edit privileges across the site.
+- **`global viewer` group** — can see *all* sources, published or not.
+- **Superuser** — bypasses every permission check.
+
+Group memberships have **expiration dates** (`GroupMembership.expiration`); `user_group_valid()` treats an expired membership as no membership.
+
+Old groups `contributor` and `project manager` are deprecated. Only `editor` and `global viewer` are active. Some documentation refers to a "Debra" — this means superuser, after Debra Lacoste, the project manager. There is no `Debra` role in code.
+
+## Migrations
+
+- Migrations auto-run on container build.
+- Data migrations using `RunPython` should define `reverse_code` where applicable.
+
+## Project quirks
+
+- **Single `published` field controls visibility.** OldCantus had separate `published` and `visible`; that was intentionally collapsed. `published=True` → everyone sees it; `published=False` → only logged-in users. Do not reintroduce a `visible` field.
+- **CSV/JSON exports diverge from OldCantus.** Some endpoints (`/json-node`, `/json-activity`) behave differently or aren't implemented. Verify the OldCantus shape before claiming parity — README has the diff.
+
+## More context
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,86 @@
...
+docker compose up
+```
+
+## Tests
+
+Tests run inside the dev `django` container. Make sure the dev stack is up (`docker compose up -d`), then:
+
+```bash
+docker compose exec -T django python manage.py test main_app.tests
+```
```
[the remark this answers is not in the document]
liampond: I'm not sure first-time setup instructions belong in `CLAUDE.md`, although this is a fair point.

CLAUDE.md
```diff
@@ -0,0 +1,86 @@
...
+- **`global viewer` group** — can see *all* sources, published or not.
+- **Superuser** — bypasses every permission check.
+
+Group memberships have **expiration dates** (`GroupMembership.expiration`); `user_group_valid()` treats an expired membership as no membership.
+
+Old groups `contributor` and `project manager` are deprecated. Only `editor` and `global viewer` are active. Some documentation refers to a "Debra" — this means superuser, after Debra Lacoste, the project manager. There is no `Debra` role in code.
+
+## Migrations
+
+- Migrations auto-run on container build.
```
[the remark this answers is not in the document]
liampond: Corrected

CLAUDE.md
```diff
@@ -0,0 +1,86 @@
...
+- Data migrations using `RunPython` should define `reverse_code` where applicable.
+
+## Project quirks
+
+- **Single `published` field controls visibility.** OldCantus had separate `published` and `visible`; that was intentionally collapsed. `published=True` → everyone sees it; `published=False` → only logged-in users. Do not reintroduce a `visible` field.
```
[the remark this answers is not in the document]
liampond: Corrected

Zganz404 (commented):
Good improvement overall. One edge case: empty string vs null aren't treated the same downstream — might need a normalization step.

CLAUDE.md
```diff
@@ -0,0 +1,74 @@
...
+
+UI/template changes can't be verified from the terminal — say so explicitly rather than claiming success.
+
+**No pre-commit hook is configured** — run `black`, `mypy`, `pylint`, and `djlint` yourself before pushing.
+
+## Branches and PRs
+
+Three long-lived branches: `production`, `staging`, and `develop`. PR against `develop`.
+
+Branch names: `feat/<topic>` or `fix/<topic>`.
```
kyrieb-ekat, line 35: I might suggest also adding a line specifying work on segments begin with a segment prefix; for CCDB items, to distinguish them from general CDB fixes I've been opening them as CCDB/feat: blah.

CLAUDE.md
```diff
@@ -0,0 +1,74 @@
...
+# Confirm a migration is needed after model changes
+docker compose exec -T django python manage.py makemigrations --dry-run
+```
+
+**Migrations do not auto-run.** Apply them manually: `docker compose exec -T django python manage.py migrate`. Data migrations using `RunPython` should define `reverse_code` where applicable.
+
+UI/template changes can't be verified from the terminal — say so explicitly rather than claiming success.
+
+**No pre-commit hook is configured** — run `black`, `mypy`, `pylint`, and `djlint` yourself before pushing.
+
```
kyrieb-ekat, line 30: A helpful item to mention in this block would be that the mac-side foilder visible in VS Code may be empty stubs—the git repo lives inside a Docker volume, accessible at `code/cantusdb_project` inside of the container. Additionally, all git operations, file edits, and test runs happen _inside_ that container and not on the host. I've encountered some initial issues if this isn't explicitly stated in the opening of a chat with Claude, since when Claude code runs from the host working directory it can't see the source files directly, so an explicit mention might help spare some time. An instruction to provide to a new dev for opening or finding the container in VS Code for the first time would be to use Cmd-Shift-P "Dev Containers: Reopen in Dev Container".

CLAUDE.md
```diff
@@ -0,0 +1,74 @@
...
+Group memberships have **expiration dates** (`GroupMembership.expiration`); `user_group_valid()` treats an expired membership as no membership.
+
+Old groups `contributor` and `project manager` are deprecated. Only `editor` and `global viewer` are active.
+
+## Gotchas
+
+- **Single `published` field controls visibility.** OldCantus had separate `published` and `visible`; that was intentionally collapsed. Do not reintroduce a `visible` field. See *Architecture and permissions* for what `published` actually gates — it's not a simple "anonymous vs. logged-in" split.
+- **CSV/JSON exports diverge from OldCantus.** Some endpoints (`/json-node`, `/json-activity`) behave differently or aren't implemented. Verify the OldCantus shape before claiming parity — README has the diff.
+
```
kyrieb-ekat, line 67: Perhaps here or in an additional section on troubleshooting might be something on an error I've encountered before, which is "Bus error on git checkout" (macOS). If git crashes with `Bus error` inside the container, first check disk space with `df -h .` — the container mounts the Mac host filesystem and inherits its space constraints. Workaround: `git config core.packedGitWindowSize 1 && git config core.packedGitLimit 1 && git config core.preloadindex false`, then retry. As a last resort for branch switches, manually set `echo "ref: refs/heads/develop" > .git/HEAD && git reset --hard HEAD`. Also: the bus error leaves a `.git/index.lock` file — always `rm -f .git/index.lock` before retrying.

CLAUDE.md
kyrieb-ekat: oh, also, if the devcontainer fails to start with a healthcheck error for `cantusdb-postgres-1`, the container is stale (created before a healthcheck was added). Fix: `docker compose rm -f postgres` on the host, then reopen.

CLAUDE.md
```diff
@@ -0,0 +1,74 @@
...
+Branch names: `feat/<topic>` or `fix/<topic>`.
+
+Commit messages: Conventional Commits (`feat:`, `fix:`, `refactor:`, `docs:`, `chore:`).
+
+**Keep PRs small and focused.** Sprawling diffs are hard to review and slow the project down.
+
+- Watch for scope creep. If you find yourself refactoring surrounding code, touching many files, or pulling in unplanned schema/migration work, pause and warn the user before continuing.
+- New dependencies need explicit sign-off — don't add them silently.
+
+## Architecture and permissions
```
kyrieb-ekat, line 44: Would a section describing the segments, where they are and their hierarchy as related to CDB be helpful? It might prevent some future confusion. For CCDG this might be helpful, as it has its own teampltes and CSS. The shared stylesheets like `sticky_header.css` affect both CCDB and the main app, so both should be tested with a modification, potentially?

CLAUDE.md
liampond: Done

CLAUDE.md
liampond: I'm not using Dev Containers so I think this would be more appropriate for your CLAUDE.local.md

CLAUDE.md
liampond: Also possibly best for your CLAUDE.local.md -- I haven't run into this yet

CLAUDE.md
liampond: Hmm, I'm kind of torn. How much development on segments do you think we'll do in the future? Because this might not need to be loaded into the context for every session. In general, Anthropic recommends avoiding explaining the structure of the codebase, since the agent can easily find that information themselves when necessary.

PR #2032, opened 2026-05-05
feat: extend source general search to cover all metadata fields

### LINKED ISSUE #1695: Make sure previous identifiers and source identifiers are searchable
As we move towards widespread use of our source identifier model, we should make sure these are searchable.

### LINKED ISSUE #1979: Search needs to return results from "DACT ID" and "Fragmentarium ID" fields
On this page: https://cantusdatabase.org/chant-search/ I need to be able to find records with their DACT ID if that ID is not in the shelfmark (for example, this one: https://cantusdatabase.org/source/1000102).

I found this one because I knew the shelfmark, but I couldn't find it via D:06e4d.

### LINKED ISSUE #1988: make "olim" IDs searchable in public interface
Currently we store some alternate identifiers in the back end, with assorted relationships like "former shelfmark." This seems to be searchable in the back end, but not visible or searchable on the public site. Ideally we'd like people who have come across "Benevento V.19" in a citation to find "Benevento 19" through these olim identifiers.

### LINKED ISSUE #2014: General search should search all possible/available metadata

PouyaMohseni:
- Add `fragmentarium_id` and `dact_id` fields to general search Q objects
- Add `identifiers__identifier` lookup to cover all `SourceIdentifier` types (olim/former shelfmarks, alternative names, etc.)
- Add `provenance__name` to general search, which was omitted in the old Cantus
- Collapse internal whitespace (as mentioned by Ich in the April 25th meeting)
- Include colon in unquoted term Regex so IDs are kept as a single token instead of being splitted by `:`

resolves #2014, resolves #1979, resolves #1988, resolves #1695

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views/test_source.py
+++ b/django/cantusdb_project/main_app/tests/test_views/test_source.py
@@ -10,7 +10,7 @@ from django.urls import reverse
 from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.auth import get_user_model
 
-from main_app.models import Source, Chant, Differentia
+from main_app.models import Source, Chant, Differentia, SourceIdentifier
 from main_app.tests.make_fakes import (
     make_fake_source,
     make_fake_segment,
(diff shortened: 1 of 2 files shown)


liampond (changes requested):
Some thoughts:
- Add at least one positive test for each of the four new Q Objects (maybe one of them can have a `D:06e4d`-style token so we test the new regex).
- Update the comment at [source.py#L405-L409](https://github.com/DDMAL/CantusDB/blob/develop/django/cantusdb_project/main_app/views/source.py#L405-L409)
- Are you sure we want to remove the `\b` at the end of the regex filter?

liampond (approved):

PR #2033, opened 2026-05-05
fix: normalize US country name variants

### LINKED ISSUE #1956: united states is divided
In Browse Sources, we have both "USA" and "United States". One gets results, the other does not:

Image

PouyaMohseni:
- Add data migration 0042 to map US variants to `United States of America`

resolves: #1956

### DIFF
--- a/django/cantusdb_project/main_app/management/commands/normalize_us_country_name.py
+++ b/django/cantusdb_project/main_app/management/commands/normalize_us_country_name.py
@@ -1,0 +1,20 @@
+from django.core.management.base import BaseCommand
+
+from main_app.models import Institution
+
+US_NAME_VARIANTS = ["USA", "United States"]
+CANONICAL_NAME = "United States of America"
+
+
+class Command(BaseCommand):
+    help = "Normalize US country name variants to 'United States of America'"
+
+    def handle(self, *args: str, **options: str) -> None:
+        updated = Institution.objects.filter(country__in=US_NAME_VARIANTS).update(
+            country=CANONICAL_NAME
+        )
+        self.stdout.write(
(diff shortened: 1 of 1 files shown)


ahankinson:
Does this need to be a migration? Why not just write a script and do it? 

Edit: To clarify, if this is a migration it will be carried along forever, and will be applied to every database change. But this seems like a "one-and-done" where the existing data simply needs to be cleaned up. For that, a simple one-off script can do it. 

If you want to ensure that it's *always* done, this seems like a good use for a custom data validation. 

```
def validate_country_name(value):
    if value in ("USA", "United States"):
        raise ValidationError(
            _("This field should use the full name, 'United States of America'; currently it is: %(value)s "),
            params={"value": value},
        )

... then in the Model:

country = model.CharField(..., validators=[validate_country_name])
```

This can also be expanded to other accepted country names, if needed.

See: https://docs.djangoproject.com/en/6.0/ref/validators/

PouyaMohseni:
> Does this need to be a migration? Why not just write a script and do it?

I originally thought the migration would be cleaner, but looking at other scripts in the management commands, it sure belongs there.

> If you want to ensure that it's always done, this seems like a good use for a custom data validation.

I'm not sure about this, as the free-text field might have other inputs like "US" or "U.S.A.". On the other hand, the only cleanup has been regarding these two versions, so it makes sense from that viewpoint.

ahankinson:
> as the free-text field might have other inputs like "US" or "U.S.A."

Those can be added to the list of things to be validated.

ahankinson (approved):

ahankinson:
I've approved it, but you might want to open a new issue about country name validation.

PR #2034, opened 2026-05-05
Disable the `notation` links in source and chant view

### LINKED ISSUE #1995: 500 when you click a Bower script name
When viewing a source from the Bower database, we can now see a box that includes a notation identification, which is hyperlinked. When the hyperlink is clicked, you get a 500. 

Image
Image

PouyaMohseni:
resolves #1995

### DIFF
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -494,9 +494,7 @@
                             <br />
                         {% endif %}
                         {% if source.notation.exists %}
-                            <b>
-                                <a href="{% url 'notation-detail' source.notation.all.first.id %}">{{ source.notation.all.first.name }}</a>
-                            </b>
+                            Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source.notation.all.first.name }}</b>
                             <br />
                         {% endif %}
                         {% if source.inventoried_by.exists %}
--- a/django/cantusdb_project/main_app/templates/chant_detail.html
+++ b/django/cantusdb_project/main_app/templates/chant_detail.html
@@ -493,8 +493,8 @@
                             Cursus: <b>{{ source.cursus|default_if_none:"" }}</b>
                             <br />
                         {% endif %}
-                        {% if source.notation.exists %}
-                            Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source.notation.all.first.name }}</b>
+                        {% if source_notation %}
+                            Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source_notation.name }}</b>
                             <br />
                         {% endif %}
                         {% if source.inventoried_by.exists %}
--- a/django/cantusdb_project/main_app/templates/source_detail.html
+++ b/django/cantusdb_project/main_app/templates/source_detail.html
@@ -235,8 +235,8 @@
                 Cursus: <b>{{ source.cursus|default_if_none:"" }}</b>
                 <br />
             {% endif %}
-            {% if source.notation.exists %}
-                Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source.notation.all.first.name }}</b>
+            {% if source_notation %}
+                Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source_notation.name }}</b>
                 <br />
             {% endif %}
         {% if source.inventoried_by.exists %}
--- a/django/cantusdb_project/main_app/templates/source_detail.html
+++ b/django/cantusdb_project/main_app/templates/source_detail.html
@@ -236,10 +236,9 @@
                 <br />
             {% endif %}
             {% if source.notation.exists %}
-                Notation: <b><a href="{% url 'notation-detail' source.notation.all.first.id %}">
-                {{ source.notation.all.first.name }}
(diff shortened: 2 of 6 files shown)


django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -494,9 +494,7 @@ <h4>Melodies</h4>
                             <br />
                         {% endif %}
                         {% if source.notation.exists %}
-                            <b>
-                                <a href="{% url 'notation-detail' source.notation.all.first.id %}">{{ source.notation.all.first.name }}</a>
-                            </b>
+                            Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source.notation.all.first.name }}</b>
```
ahankinson: Unless you are really careful, traversing query sets like this in templates is a really easy way to multiply sql queries. You should always do queries in the view, and use the template layer just for rendering.

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -47,6 +47,8 @@
 from main_app.mixins import JSONResponseMixin
 from users.models import User
 
+BOWER_SEGMENT_ID = 4064
```
ahankinson: Use the IDs set in settings.py!!!!

django/cantusdb_project/main_app/views/chant.py
```diff
@@ -326,6 +328,10 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         source = chant.source
 
         context["user_can_edit_chant"] = self.user_assigned_to_source(source)
+        context["bower_segment"] = (
+            source is not None
+            and BOWER_SEGMENT_ID in source.segment_m2m.values_list("id", flat=True)
```
ahankinson: Why are you fetching a whole list and doing the “in” in Python when you can do it as a DB query?

django/cantusdb_project/main_app/views/chant.py
PouyaMohseni: I stuck with the practices in the `source` view; I’ll refactor both views.

django/cantusdb_project/main_app/templates/chant_detail.html
PouyaMohseni: I will move the query to the view as part of the refactor, since this was the original approach before the commit.

django/cantusdb_project/main_app/views/chant.py
ahankinson: Yeah, those should also use the constants in settings.py

django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -235,11 +235,10 @@ <h4 class="font-serif">{{ source.short_heading }}</h4>
                 Cursus: <b>{{ source.cursus|default_if_none:"" }}</b>
                 <br />
             {% endif %}
-            {% if source.notation.exists %}
-                Notation: <b><a href="{% url 'notation-detail' source.notation.all.first.id %}">
-                {{ source.notation.all.first.name }}
-            </a></b>
-        {% endif %}
+            {% if source_notation %}
+                Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source_notation.name }}</b>
```
ahankinson, line 239: If a source has multiple notations, why not show all of them?

PouyaMohseni: Thank you for your comment. I kept the previous logic since this only concerns disabling links. Although the field is modeled as many-to-many, I don't see any source with multiple notations. I could open a question on the issue page asking whether this is the type of modeling we need.

django/cantusdb_project/main_app/templates/chant_detail.html
```diff
@@ -493,10 +493,8 @@ <h4>Melodies</h4>
...
                             <br />
                         {% endif %}
-                        {% if source.notation.exists %}
-                            <b>
-                                <a href="{% url 'notation-detail' source.notation.all.first.id %}">{{ source.notation.all.first.name }}</a>
-                            </b>
+                        {% if source_notation %}
+                            Notation{% if bower_segment %} (Bower){% endif %}: <b>{{ source_notation.name }}</b>
                             <br />
                         {% endif %}
```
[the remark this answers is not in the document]
PouyaMohseni: It is a small temporary change

django/cantusdb_project/main_app/templates/source_detail.html
```diff
@@ -235,11 +235,10 @@ <h4 class="font-serif">{{ source.short_heading }}</h4>
...
+                <br />
+            {% endif %}
```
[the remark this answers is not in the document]
PouyaMohseni: It is a small temporary change

liampond:
I just added some tests. Otherwise, everything looks good to me. @PouyaMohseni, let me know if the tests look good. If they do, I'll approve the PR.

PouyaMohseni:
The tests look good and comprehensive: nice. Yet, I do not know whether adding tests for a temporary change is relevant.

liampond:
That's true. Who knows when we'll get around to fixing this, though — the data cleanup could take a while. Maybe it was unnecessary.

liampond (approved):

PR #2044, opened 2026-05-08
feat: add CCDB sub-pages, chant search, and multi-database filter

kyrieb-ekat:


django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,122 @@
...
+
+from main_app.models import Chant, Source
+from main_app.views.chant import ChantSearchView
+from main_app.views.source import SourceListView
+
+AVAILABLE_SEGMENTS = [
+    (settings.CCDB_SEGMENT_ID, "Canadian Chant Database"),
+    (settings.CANTUS_SEGMENT_ID, "Cantus Database"),
+    (settings.BOWER_SEGMENT_ID, "Sequence Database"),
+    (4067, "Cantorales in the Americas and Beyond"),
```
ahankinson: Why isn’t this in settings too?

kyrieb-ekat: Oops, oversight, fixed now!

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,122 @@
...
+        self._segment_ids = self._get_selected_segment_ids()
+
+        if len(self._segment_ids) == 1:
+            self._current_segment_id = self._segment_ids[0]
+            return super().get_queryset()
+
+        # Multi-segment: call the parent once per segment, then union the results.
+        # We cannot filter after .union(), so segment injection happens via get_segment_id().
+        querysets = []
+        for seg_id in self._segment_ids:
```
ahankinson, line 88: Where is `self._current_segment_id` initialised?

kyrieb-ekat: @ahankinson It's set inside `get_queryset()`. line 78 for single-segment, line 85 in the loop. The `getattr` default on line 68 handles calls before that.

ahankinson: That's where it's set, but where is it initialised?

kyrieb-ekat: d'oh; okay the coffee has officially not kicked in. I just added __init__ before `_get_selected_segment_ids` and simplified line 68 to swap `getattr` for direct access.

ahankinson: You probably also want to do the same to `self._segment_ids`?

kyrieb-ekat: Done. Thank you.

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -58,15 +58,18 @@ class CcdbChantSearchView(CcdbMixin, ChantSearchView):
...
     def _get_selected_segment_ids(self) -> list[int]:
         """Return the segment IDs selected via 'db' GET params. Defaults to CCDB segment."""
         raw = self.request.GET.getlist("db")
         ids = [int(v) for v in raw if str(v).isdigit()]
         return ids if ids else [settings.CCDB_SEGMENT_ID]
 
     def get_segment_id(self) -> Optional[str]:
-        seg = getattr(self, "_current_segment_id", None)
-        return str(seg) if seg is not None else None
+        return str(self._current_segment_id) if self._current_segment_id is not None else None
```
ahankinson: It's coming in as a str, and being used as a str, so why do you convert it to an `int`? If it's just about comparing it to the value in the settings, better to cast that to a str() once, than doing the conversion all the time.

kyrieb-ekat: got it; updated so it only calls the once and remains [int] for the ORM filter

django/cantusdb_project/main_app/views/ccdb.py
```diff
@@ -0,0 +1,126 @@
...
+    Django forbids filtering a unioned queryset — multi-segment support works by calling
+    the parent's get_queryset() once per selected segment and unioning the results.
+    """
+
+    template_name = "ccdb_chant_search.html"
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self._current_segment_id: Optional[str] = None
+        self._segment_ids: list[int] = []
```
ahankinson, line 64: Just a note on style; not a blocker.... 

As of Python 3.10, `Optional[foo]` can also be written as `foo | None`: 

```py
self._current_segment_id: str | None = None
```

The typing module also defines `List` but it is being deprecated in favour of the built-in `list`.

All that is to say that importing the `typing` module is no longer required for most type annotations. You should probably gradually migrate the code to using the builtins, which means not mixing the `typing` module-style annotations with the built-in style annotations.

ahankinson (approved):
I'm going to approve it so that I'm not a blocker, but in general the chant querying logic here needs to be simplified (not just for ccdb, but all querying logic in `chant.py`.

django/cantusdb_project/main_app/views/ccdb.py
kyrieb-ekat: This is very helpful, thank you! I'll add this as a comment to the issue you made about formatting, so we can start tracking general formatting sweeps we need to do for CDB+.

PR #2051, opened 2026-05-11
Replace century dropdown with year-range slider

### LINKED ISSUE #991: Source List page: Filtering by Century could be improved
I just got an email from Debra:

A suggestion has come in that I think will improve the use of the pull-down list for century for CantusDB sources.

> I'm on this page:
> https://cantusdatabase.org/sources/?segment=4063
> 
> Would it be possible to have "Century" just as: "9th century", "10th century", etc. up to "17th century" in the pull-down list, and then have 2 additional fields with this for one: "1st half", "2nd half", and then this for the next: "1st quarter", "2nd quarter​‌", "3rd quarter," "4th quarter."
> (no year dates are needed, then)
> 
> The pull-down lists will be shorter and easier to manage (see screenshot).
> 
> I know this changes the search programming, which you currently have matching all of the same century (I think) no matter which part of the century is selected. (So, this is already great, that it pulls up all sources of the same century even if some have a more detailed specification.)
> 
> Does this make it more streamlined or more difficult on the back end? For me, I can see it being more streamlined for the user, because I can specify the century if I know it, and then if I know which half I can choose that, and if I know even more I can continue to the quarter and specify that (or I can leave it blank at whatever point I am unsure).
> But, if it splits up the data too much and you can't attach "1st half" because it's not 'century-specific' then I understand the difficulty. In my description, above, "1st half" would be in a separate pull-down and would be applied whenever the user wanted that, based on their selection of the "century" field - but there wouldn't be a different "1st half" option for each century --- instead, just one that would rely on the previous field for context.  If you know what I mean.
> 
> Also, there is a 9th and a 09th in the current list that should be cleaned up, regardless of what happens to this Century field(s).

This would be a nice thing to implement, and I can see how it would improve user experience.

This entails two main tasks:
- [x] Merging all "9th Century" sources into the "09th Century" (update: the "9th Century" was actually "9th century (875-900)" - I renamed it to "09th century (875-900)", so we haven't lost any information on which quarter-century this Century object represented)
- [ ] Adapting the Century model to allow filtering by half- and quarter-century, and update Source List page to implement this filtering.

We have options for the second of these points. I'll create a comment below this one and start recording some of the thoughts I have.

PouyaMohseni:
- Replaces the `Century` dropdown on the source list, Canadian Chant DB, and Cantorales filter forms with a dual-handle year-range slider that submits `dateStart` / `dateEnd` and matches sources whose centuries overlap the selected range (inclusive). 
- Adds `min_date` / `max_date` fields to the `Century` model (migration 0042), populated for all existing rows by a data migration (0043). New centuries  created via admin auto-fill these fields on save when the name matches a known pattern.
- Slider bounds are rounded out to the nearest decade so `step="10"` reaches both endpoints, and the max is clipped to the current decade so future century stubs (e.g. an empty "21st century") don't stretch the slider.
- The regex parser used by the migration, the model save override, and  the test factory is in the shared module (: `main_app/century_dates.py`).

resolves #991

Admin auto-fill for new centuries:
autofill

Year-range filter: 
filter

### DIFF
--- a/django/cantusdb_project/main_app/migrations/0044_populate_century_dates.py
+++ b/django/cantusdb_project/main_app/migrations/0044_populate_century_dates.py
@@ -52,7 +52,7 @@ def reverse_populate_century_dates(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("main_app", "0042_century_date_fields"),
+        ("main_app", "0043_century_date_fields"),
     ]
 
     operations = [
--- a/django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
+++ b/django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
@@ -8,6 +8,7 @@
     <link href="{% static 'css/ccdb_style.css' %}"
           rel="stylesheet"
           media="screen" />
+    <link href="{% static 'css/source_list.css' %}" rel="stylesheet" />
 {% endblock %}
 {% block header %}
     {% include "header/sticky_header.html" %}
@@ -191,15 +192,37 @@
                                         </select>
                                     </div>
                                     <div class="col-lg-6">
-                                        <label for="centuryFilter" class="small form-label">Century</label>
-                                        <select id="centuryFilter"
-                                                name="century"
-                                                class="form-control form-select form-select-sm">
-                                            <option value="">- Any -</option>
-                                            {% for century in centuries %}
-                                                <option value="{{ century.id }}">{{ century.name }}</option>
-                                            {% endfor %}
-                                        </select>
+                                        <div class="d-flex justify-content-between align-items-center">
+                                            <label class="small form-label fw-bold mb-0">Year range (overlapping)</label>
+                                            <button type="button"
+                                                    id="rangeResetBtn"
+                                                    class="btn btn-link btn-sm p-0 small text-danger">Clear</button>
+                                        </div>
+                                        <div class="range-slider">
+                                            <div class="range-slider-track"></div>
+                                            <div class="range-slider-fill" id="rangeSliderFill"></div>
+                                            <input type="range"
+                                                   id="dateStartFilter"
+                                                   name="dateStart"
+                                                   class="range-slider-input"
+                                                   min="{{ date_range_min }}"
+                                                   max="{{ date_range_max }}"
+                                                   step="10"
+                                                   value="{{ request.GET.dateStart|default:date_range_min }}" />
+                                            <input type="range"
+                                                   id="dateEndFilter"
+                                                   name="dateEnd"
+                                                   class="range-slider-input"
+                                                   min="{{ date_range_min }}"
+                                                   max="{{ date_range_max }}"
+                                                   step="10"
+                                                   value="{{ request.GET.dateEnd|default:date_range_max }}" />
+                                        </div>
+                                        <div class="d-flex justify-content-between small mt-1">
+                                            <span class="fw-bold">{{ date_range_min }}</span>
+                                            <span class="fw-bold text-primary" id="rangeSelectedLabel"></span>
+                                            <span class="fw-bold">{{ date_range_max }}</span>
+                                        </div>
                                     </div>
                                     <div class="col-lg-6">
                                         <label for="prodMethodFilter" class="small form-label">Manuscript/Print</label>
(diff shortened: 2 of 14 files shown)


django/cantusdb_project/main_app/tests/test_views/test_source.py
```diff
@@ -1006,49 +1008,42 @@ def test_filter_by_provenance(self):
...
-        multiple_century_source = make_fake_source(
-            published=True,
-            shelfmark="source",
-        )
-        multiple_century_source.century.set([ninth_century, tenth_century])
+        fifteenth_century_source = make_fake_source(published=True, shelfmark="15th")
+        fifteenth_century_source.century.set([fifteenth_century])
 
-        # display sources in ninth century
+        # Range 800-999 overlaps 9th and 10th centuries; not 15th
```
kyrieb-ekat, line 1025: so not to be overly technical but the final year of a century is the "rollover" of the next year. So, 1500 is both the last year of the preceding century and the first year of the ensuing. People will use those numbers such, and potentially be confused there.

django/cantusdb_project/main_app/admin/source.py
ahankinson, line 74: Why is this here?

django/cantusdb_project/main_app/tests/test_views/test_source.py
annamorphism: funny, I was about to suggest the opposite, which is that we should impose some consistency on years like 1500 so they don't belong to two groups (otherwise a search for "1500-1600" might get you a whole bunch of stuff labelled "1400-1500" that you really don't want.) But maybe that's a data cleaning question and not relevant to this test...

django/cantusdb_project/main_app/admin/source.py
PouyaMohseni: Sorry, I was rebasing to undraft the PR.

annamorphism:
> Year-range filter: filter

is this an actual screenshot of this search? bc the first result is not from 1500-1770 (but maybe it was cached from something you had done earlier...)

django/cantusdb_project/main_app/tests/test_views/test_source.py
ahankinson: Yeah, if you overlap end and start you risk incorporating two century ranges instead of one, since you will get two matches for what the “14th century” expands to, meaning you will get 1400-1600, since 1500 belongs both to the 14th century (1400-1500) and to the 15th century (1500-1600)

In most of the systems I’ve built, it’s e.g., 1400-1499.

I have also seen interpretations put the centuries at 1401-1500.

liampond: From monthly meeting: Debra and Jennifer would like 1400-1499 for 15th century

liampond:
From monthly Cantus meeting: make post-Vatican II 1965

PouyaMohseni:
Reverted to the previous century-to-year mapping function as was requested.

django/cantusdb_project/main_app/migrations/0044_populate_century_dates.py
```diff
@@ -0,0 +1,60 @@
+# Data migration to populate min_date and max_date fields for Century model
+# Based on century data patterns in the database
+
+from django.db import migrations
+
+from main_app.century_dates import century_name_to_dates
+
+
```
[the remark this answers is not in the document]
PouyaMohseni: I think this is acceptable here because century_dates.py is a pure utility with no model dependencies.

django/cantusdb_project/main_app/templates/source_lists/canadian_chant_db.html
```diff
@@ -191,15 +192,37 @@ <h5 class="mb-2">Search Chants</h5>
...
-                                                name="century"
-                                                class="form-control form-select form-select-sm">
-                                            <option value="">- Any -</option>
-                                            {% for century in centuries %}
-                                                <option value="{{ century.id }}">{{ century.name }}</option>
-                                            {% endfor %}
-                                        </select>
+                                        <div class="d-flex justify-content-between align-items-center">
+                                            <label class="small form-label fw-bold mb-0">Year range (overlapping)</label>
+                                            <button type="button"
```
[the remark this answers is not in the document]
PouyaMohseni: This is generally a valid suggestions but out of scope for this PR as there is no aria-label for other labels.

liampond (approved):
I would recommend copying `century_name_to_dates` into the migration instead of importing it. If `century_dates.py` ever gets refactored, moved, or deleted, the migration could fail. It's up to you. Otherwise, looks good.

PouyaMohseni:
The only use of `century_dates.py` is mapping century to dates, whether in the migration or data entry. So, if we refactor it for one, we need the refactored one for the other.

Issue #732, opened 2023-06-12
We need to write tests for our Password Reset views


jacobdgm:
Writing tests for the new password reset views has proven to be a tough nut to crack. Opening this issue so we can quickly get the password reset feature up on staging and then to production.

Here is the not-quite-working test I currently have:

```python
class PasswordResetViewsTest(TestCase):
    def setUp(self):
        # for testing, set our email backend to send to local memory
        # (e.g. so we can access mail.outbox)
        settings.EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"

        self.user = get_user_model().objects.create(
            email="presbyter_iohannes@exemplum.com"
        )
        self.user.set_password("pass")
        self.user.save()

    def test_password_reset_views(self):
        response_1 = self.client.get(reverse("reset_password"))
        self.assertEqual(response_1.status_code, 200)
        self.assertEqual(response_1.template_name[0], "registration/reset_password.html")

        response_2 = self.client.post(
            reverse("reset_password"),
            {"email": "presbyter_iohannes@exemplum.com"}, # this has to be an email
            # address existing in the database, or django won't send the email
        )
        self.assertEqual(response_2.status_code, 302)
        self.assertEqual(response_2.url, reverse("reset_password_done"))
        self.assertEqual(len(mail.outbox), 1)

        token = response_2.context[0]["token"]
        uid = response_2.context[0]["uid"]

        print(uid)

        response_2_5 = self.client.get(
            reverse(
                "reset_password_confirm",
                kwargs={"token": token, "uidb64": uid},
            )
        )
        self.assertEqual(response_2_5.status_code, 200) # <- the test fails here, as the status code is 302
            # it tries to redirect us to /reset/MQ/set-password
        response_3 = self.client.post(
            reverse(
                "reset_password_confirm",
                kwargs={"token": token, "uidb64": uid},
            ),
            {"new_password1": "egoverushomo", "new_password2": "egoverushomo"},
        )
        self.assertEqual(response_3.status_code, 302)
        self.assertEqual(response_3.url, reverse("reset_password_complete"))
```

The test above was adapted from https://stackoverflow.com/questions/18264724/testing-django-1-5-reset-password-form-how-to-generate-the-token-for-the-test#answer-19738048. In researching this, it looks like discussion related to https://stackoverflow.com/questions/50415700/unable-to-create-an-integration-test-for-djangos-reset-password-flow could also be useful.

kyrieb-ekat:
[...]

PR #2054, opened 2026-05-12
Add tests for password reset and pagination

### LINKED ISSUE #321: Pagination Tests
There are currently no tests of pagination in the test suites.

It seems there was an effort at one point to write an abstract test class that could be plugged into all `ListViewTest`s to test pagination, but it was never hooked up to anything (see it near the top of the file at https://github.com/DDMAL/CantusDB/pull/318/commits/e7f30349cce8cc1b0ae9e8c257936ad155347065). Someone who is more motivated and/or more clever than me might be able to use this existing code to test pagination in all list views.

PouyaMohseni:
- Added password reset tests for display, email sending, complete reset flow, and invalid token handling
- Added pagination tests for 5 view classes: testing full pages, overflow, last page, and error cases

resolves #321 and #732

### DIFF
--- a/django/cantusdb_project/main_app/tests/test_views/test_auth.py
+++ b/django/cantusdb_project/main_app/tests/test_views/test_auth.py
@@ -2,8 +2,12 @@
 Tests for views in views/auth.py
 """
 
-from django.test import TestCase
+from django.contrib.auth.tokens import default_token_generator
+from django.core import mail
+from django.test import TestCase, override_settings
 from django.urls import reverse
+from django.utils.encoding import force_bytes
+from django.utils.http import urlsafe_base64_encode
 from django.contrib.auth import get_user_model
 
 
@@ -40,3 +44,60 @@ class ChangePasswordViewTest(TestCase):
         self.assertEqual(
             response_2.status_code, 200
         )  # if login failed, status code will be 302
+
+
+@override_settings(EMAIL_BACKEND="django.core.mail.backends.locmem.EmailBackend")
+class PasswordResetViewsTest(TestCase):
+    def setUp(self):
+        self.user = get_user_model().objects.create(email="reset@test.com")
+        self.user.set_password("oldpass")
+        self.user.save()
+
+    def test_reset_password_form_renders(self):
+        response = self.client.get(reverse("reset_password"))
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "registration/reset_password.html")
+
+    def test_reset_password_sends_email(self):
+        response = self.client.post(
+            reverse("reset_password"), {"email": "reset@test.com"}
+        )
+        self.assertRedirects(response, "/reset-password-sent/")
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn("reset", mail.outbox[0].subject.lower())
+        self.assertIn("reset@test.com", mail.outbox[0].to)
+
+    def test_reset_password_confirm_completes(self):
+        uid = urlsafe_base64_encode(force_bytes(self.user.pk))
+        token = default_token_generator.make_token(self.user)
+
+        # Django redirects to …/set-password/ on valid token; follow to land there
+        confirm_response = self.client.get(
+            reverse("reset_password_confirm", kwargs={"uidb64": uid, "token": token}),
+            follow=True,
+        )
+        self.assertEqual(confirm_response.status_code, 200)
+
+        set_password_url = confirm_response.wsgi_request.path
+        post_response = self.client.post(
+            set_password_url,
+            {"new_password1": "newpass123!", "new_password2": "newpass123!"},
+        )
+        self.assertRedirects(post_response, "/reset-password-complete/")
+
+        self.assertTrue(
+            self.client.login(email="reset@test.com", password="newpass123!")
+        )
+        self.assertFalse(self.client.login(email="reset@test.com", password="oldpass"))
+
+    def test_reset_password_invalid_token(self):
+        uid = urlsafe_base64_encode(force_bytes(self.user.pk))
+        response = self.client.get(
+            reverse(
+                "reset_password_confirm",
+                kwargs={"uidb64": uid, "token": "invalid-token"},
+            ),
+            follow=True,
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertFalse(response.context.get("validlink", False))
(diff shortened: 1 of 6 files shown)


django/cantusdb_project/main_app/tests/test_views/test_user.py
```diff
@@ -66,3 +69,63 @@ def test_shown_users(self) -> None:
         response = self.client.get(reverse("indexer-list"))
         self.assertEqual(response.status_code, 200)
         self.assertCountEqual([self.user_with_source], response.context["indexers"])
+
+    def test_pagination(self) -> None:
+        paginate_by = IndexerListView.paginate_by
+        full_pages = 2
+        # setUpTestData already creates 1 visible indexer (user_with_source),
+        # so create paginate_by * full_pages - 1 sources to reach exactly 2 full pages.
```
kyrieb-ekat, line 77: okay I see what this is doing, but potentially this could cause silent/unknown failure in the event we do ever test with a second user + a source; the page count would be wrong and `len(indexers) == paginate_by` would fail, I think, without notice. This can be isolated, or have the count queried before new records are created.

django/cantusdb_project/main_app/tests/test_views/test_user.py
```diff
@@ -66,3 +69,63 @@ def test_shown_users(self) -> None:
...
+        for _ in range(paginate_by * full_pages - 1):
+            make_fake_source(published=True)
+
+        for page_num in range(1, full_pages + 1):
+            response = self.client.get(reverse("indexer-list"), {"page": page_num})
+            self.assertEqual(response.status_code, 200)
+            self.assertTrue(response.context["is_paginated"])
+            self.assertEqual(len(response.context["indexers"]), paginate_by)
+
+        overflow = random.randint(1, paginate_by - 1)
```
kyrieb-ekat, line 88: random can make reproducing a test hard. so, if we end up with a random overflow = 1 and our view drops a record, the test would fail/not flag successfully. Possibly seed the random, or add a fixed value where overflow explicitly equals paginate+by // 2. It looks like random is used in this way in all of the tests in this PR

django/cantusdb_project/main_app/tests/test_views/test_user.py
```diff
@@ -66,3 +69,63 @@ def test_shown_users(self) -> None:
...
+        for invalid_page in [-1, 0, "lst", full_pages + 2]:
+            response = self.client.get(reverse("indexer-list"), {"page": invalid_page})
+            self.assertEqual(response.status_code, 404)
+
+
+class UserSourceListViewTest(TestCase):
+    def test_pagination(self) -> None:
+        paginate_by = UserSourceListView.paginate_by
+        full_pages = 2
+        user = make_fake_user()
```
kyrieb-ekat, line 114: in the other tests you use `setUpTestData`, is this different for a particular reason? if so, leave a comment explaining, or maybe keep it consistent in all tests.

django/cantusdb_project/main_app/tests/test_views/test_auth.py
```diff
@@ -40,3 +44,60 @@ def test_change_password(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, "registration/reset_password.html")
+
+    def test_reset_password_sends_email(self):
+        response = self.client.post(
+            reverse("reset_password"), {"email": "reset@test.com"}
+        )
+        self.assertRedirects(response, "/reset-password-sent/")
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn("reset", mail.outbox[0].subject.lower())
```
kyrieb-ekat, line 67: okay this is a small one but I've run into it before where any email has "reset" in the subject this would pass. we could set a specific expected subject string; maybe consider explicitly asserting `len(mail.outbox[0].to) == 1` ?

really just a hardening note

django/cantusdb_project/main_app/tests/test_views/test_chant.py
```diff
@@ -2061,6 +2061,34 @@ def test_image_link_column(self):
...
+        overflow = random.randint(1, paginate_by - 1)
+        for _ in range(overflow):
+            make_fake_chant(source=source)
+
+        response = self.client.get(reverse("chant-search"), {"page": full_pages + 1})
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.context["chants"]), overflow)
+
+        response = self.client.get(reverse("chant-search"), {"page": "last"})
+        self.assertEqual(response.status_code, 200)
```
kyrieb-ekat, line 2088: nitpick in the name of testing; this only returns/asserts a 200, not that it's the "last page". potentially asserting that the page object's `is_last_page` or that the item count equals `overflow` will help in fully verifying behavior. What you have is fine, this is just a thought.

django/cantusdb_project/main_app/tests/test_views/test_user.py
```diff
@@ -66,3 +69,63 @@ def test_shown_users(self) -> None:
...
+        self.client.force_login(user)
```
kyrieb-ekat, line 115: what will happen if a user without access tries to access something/how do we handle unauthenticated access

kyrieb-ekat (changes requested):
Looks very good. I made a couple comments, some of which are nitpicks. The big things I'd like addressed are the random/random seed issue for reproducibility and the `IndexerListview` coupling to existing test data. Once those are tweaked/looked at, this can be merged.

django/cantusdb_project/main_app/tests/test_views/test_user.py
PouyaMohseni: Thank you for your review: a good point to consider, specifically as this issue is present in the previous tests. I will refactor that and seed random in all tests.

django/cantusdb_project/main_app/tests/test_views/test_chant.py
PouyaMohseni: Again, this has been a pre-existing pattern that this PR followed for consistency. They are, yet, worth fixing: I've addressed both across all tests in the codebase

django/cantusdb_project/main_app/tests/test_views/test_auth.py
PouyaMohseni: The subject string comes from Django's built-in password reset template, so asserting an exact string would couple the test to template wording that's outside our control and could shift with an upgrade. I think this is is enough to confirm it's the reset email and not something else.

django/cantusdb_project/main_app/tests/test_views/test_user.py
PouyaMohseni: The assumption is intentional and documented in the comment above the loop. It would only break if someone modifies setUpTestData in this same class, which they'd notice immediately since the test would fail (with a notice).

kyrieb-ekat (approved):
LGTM, nicely done!

PouyaMohseni:
As discussed, we will not merge unnecessary approved PRs to develop until we finish the migration.

PR #2058, opened 2026-05-13
cherry-pick: add source_data_contributed_by field onto staging

PouyaMohseni:
This is already in develop (#1986), but Debra also wants it on staging.


liampond (approved):
Looks good!

PR #2066, opened 2026-05-19
fix: encode spaces in image_link URLs before Django validation

### LINKED ISSUE #1868: Unable to save chants in /source/699262
It's because of the image links.

I ran into this issue while I was creating a management command to update all chants in published sources.

PouyaMohseni:
- Replace spaces in the link before the validator runs
- Add unit tests for image link functionality

--------
Update: The previous form-level NormalizedURLField only normalized values that passed through the public edit forms. Django admin, management commands, and any direct ORM write still hit URLValidator with raw spaces and
failed.

Introduce NormalizedURLField (models.URLField subclass) used by Source.image_link and BaseChant.image_link. It normalizes in both to_python (full_clean) and get_prep_value (save), so every write path benefits. Its formfield() returns NormalizedURLFormField so ModelForms and admin auto-generated forms normalize user input too.

The field deconstructs back to models.URLField, so no migration is needed and the DB column is unchanged.

--------

resolves #1868

### DIFF
--- a/django/cantusdb_project/main_app/forms.py
+++ b/django/cantusdb_project/main_app/forms.py
@@ -26,6 +26,7 @@ from .models import (
     Century,
     Sequence,
 )
+from .models.url_field import NormalizedURLFormField
 from .widgets import (
     TextInputWidget,
     VolpianoInputWidget,
@@ -1023,7 +1024,7 @@ class ImageLinkForm(forms.Form):
         initial = kwargs.get("initial")
         if initial:
             for folio in initial:
-                self.fields[folio] = forms.CharField(
+                self.fields[folio] = NormalizedURLFormField(
                     widget=HiddenInput(attrs={"class": "img-link-input"}),
                     required=False,
                 )
--- a/django/cantusdb_project/main_app/models/url_field.py
+++ b/django/cantusdb_project/main_app/models/url_field.py
@@ -1,0 +1,41 @@
+from typing import Any, Optional
+
+from django import forms
+from django.db import models
+
+
+def _normalize(value: Optional[str]) -> Optional[str]:
+    if isinstance(value, str):
+        return value.strip().replace(" ", "%20")
+    return value
+
+
+class NormalizedURLFormField(forms.URLField):
+    """Form URLField that encodes spaces before Django's URLValidator fires."""
+
+    def to_python(self, value: Any) -> str:
+        return super().to_python(_normalize(value))
+
+
+class NormalizedURLField(models.URLField):
+    """
+    Model URLField that strips whitespace and percent-encodes spaces.
+
+    Normalizes on validation (full_clean) and on save (get_prep_value), so
+    forms, admin, management commands, and direct ORM writes all benefit.
+    """
+
+    def to_python(self, value: Any) -> Optional[str]:
+        return super().to_python(_normalize(value))
+
+    def get_prep_value(self, value: Any) -> Optional[str]:
+        return super().get_prep_value(_normalize(value))
+
+    def formfield(self, **kwargs: Any) -> Optional[forms.Field]:
+        kwargs.setdefault("form_class", NormalizedURLFormField)
+        return super().formfield(**kwargs)
+
+    def deconstruct(self) -> tuple:
+        # Report as plain URLField so makemigrations doesn't see a field change.
+        name, _, args, kwargs = super().deconstruct()
+        return name, "django.db.models.URLField", args, kwargs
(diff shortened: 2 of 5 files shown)


django/cantusdb_project/main_app/forms.py
```diff
@@ -56,6 +56,15 @@
 ]
 
 
+class NormalizedURLField(forms.URLField):
+    """URLField that encodes spaces before Django's URLValidator fires."""
+
+    def to_python(self, value: Any) -> str:
+        if value:
+            value = value.strip().replace(" ", "%20")
```
ahankinson: Is there a reason why you are not using a URL encoding library?

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote

PouyaMohseni: Because it would double encode a valid link unless I had an extensive safe-character management

django/cantusdb_project/main_app/models/url_field.py
```diff
@@ -0,0 +1,41 @@
+from typing import Any, Optional
+
+from django import forms
+from django.db import models
+
+
+def _normalize(value: Any) -> Any:
```
ahankinson: Could you explain what this is doing? It’s returning None if the value is a string, but then using string methods on the value and returning that. Maybe instead of using Any you can narrow the types down on this?

django/cantusdb_project/main_app/forms.py
```diff
@@ -1023,7 +1024,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         initial = kwargs.get("initial")
         if initial:
             for folio in initial:
-                self.fields[folio] = forms.CharField(
+                self.fields[folio] = NormalizedURLFormField(
```
ahankinson, line 1027: Do you still need the form field if it’s normalizing at the model level?

PouyaMohseni: I believe yes, `ImageLinkForm.save()` calls `.update()`, so the model field covers the DB write. But Django runs URLValidator at form-validation time, before `.save()` is ever called. With a plain URLField here, a URL with spaces fails validation and the form returns a 200 instead of saving

django/cantusdb_project/main_app/models/url_field.py
PouyaMohseni: it returns the encoded string when the value is a str, and passes none otherwise.

I narrowed the type down.

django/cantusdb_project/main_app/tests/test_functions.py
```diff
@@ -490,3 +484,64 @@ def test_get_ci_text_search(self) -> None:
...
+            "https://example.com/Folio%2092r.jpg",
+        )
+
+    def test_deconstruct_reports_as_plain_urlfield(self):
+        # Keeps makemigrations from generating a no-op migration.
+        _, path, _, _ = self.field.deconstruct()
+        self.assertEqual(path, "django.db.models.URLField")
+
+    def test_formfield_returns_normalizing_form_field(self):
+        self.assertIsInstance(self.field.formfield(), NormalizedURLFormField)
```
liampond: Claude has suggested that we add another test — why not? The current tests check the field, but none of them try to save a model.

```
class ImageLinkSpaceEncodingIntegrationTest(TestCase):
    """End-to-end coverage for issue #1868: saving a model whose image_link
    contains spaces must succeed and persist the URL with spaces encoded."""

    def test_chant_with_spaced_image_link_saves(self) -> None:
        # Mirrors the management-command path from #1868: a direct .save(),
        # which BaseModel.save() routes through full_clean()/URLValidator.
        chant = make_fake_chant()
        chant.image_link = "https://example.com/Folio 92r.jpg"
        chant.save()
        chant.refresh_from_db()
        self.assertEqual(chant.image_link, "https://example.com/Folio%2092r.jpg")

    def test_source_with_spaced_image_link_saves(self) -> None:
        source = make_fake_source()
        source.image_link = "https://example.com/image gallery.jpg"
        source.save()
        source.refresh_from_db()
        self.assertEqual(source.image_link, "https://example.com/image%20gallery.jpg")

    def test_save_does_not_raise_on_spaces(self) -> None:
        # Without the fix this raises ValidationError ("Enter a valid URL"),
        # which is exactly the failure reported in #1868.
        chant = make_fake_chant()
        chant.image_link = "https://example.com/a b c.jpg"
        try:
            chant.save()
        except ValidationError:
            self.fail("Saving a chant with spaces in image_link should not raise.")
```

liampond (approved):

PR #2070, opened 2026-05-27
Data Cleaning and Validation Tasks

PouyaMohseni:


django/cantusdb_project/main_app/models/data_check_config.py
```diff
@@ -0,0 +1,49 @@
+from django.db import models
+
+
+class DataCheckConfig(models.Model):
+    FREQUENCY_CHOICES = [
```
ahankinson: Consider using enumeration types:

https://docs.djangoproject.com/en/6.0/ref/models/fields/#enumeration-types

```py
class FrequencyChoices(models.TextChoices):
    DAILY = "daily", "Daily"
    WEEKLY = "weekly", "Weekly"
    MONTHLY = "monthly", "Monthly"
```

This way you can do `FrequencyChoices.WEEKLY` in your model default instead of hardcoding the value. 

If it's an internal field that is unlikely to need a human-readable value, you might also consider using IntFields for your choices. They're slightly more efficient, you don't have to set a max value (which may bite you later if you need a longer value) and since the Choice field gives you the human-readable value as well, it will still appear as "Daily" or "Weekly" to the users.

django/cantusdb_project/main_app/models/data_check_config.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    last_run = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text=(
+            "Set automatically after each completed run. "
+            "Used to determine the editing window when scope is 'Only edited'."
+        ),
+    )
+    recipients = models.TextField(
```
ahankinson: Don't users have e-mail addresses defined? Why not wire this up to the user's table and choose the set of users that get the e-mail?

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,345 @@
-from typing import List, Dict, Any
+import json
```
ahankinson: why not use `ujson`? it's installed in the environment.

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,345 @@
...
+    """Fetch the genre for a single cantus_id from CI. Returns (cantus_id, genre_or_None)."""
+    try:
+        response = requests.get(f"{CI_DOMAIN}/json-cid/{cantus_id}", timeout=10)
+        if response.status_code == 200:
+            text = response.text.encode().decode("utf-8-sig")
+            data = json.loads(text)
+            if isinstance(data, dict) and data.get("info"):
+                return cantus_id, data["info"].get("field_genre")
+    except requests.exceptions.RequestException:
+        pass
```
ahankinson: Shouldn't you log an error or something here?

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,345 @@
...
+        "blank_cantus_id": check_blank_cantus_id(chant_ids),
+        "blank_mode": check_blank_mode(chant_ids),
+        "blank_invitatory_differentia": check_blank_invitatory_differentia(chant_ids),
+    }
+
+    # Update last_run before sending so a mail failure doesn't trigger a re-run
+    DataCheckConfig.objects.filter(pk=config.pk).update(last_run=now)
+
+    recipients = [r.strip() for r in config.recipients.split(",") if r.strip()]
+    if not recipients:
```
ahankinson: If the job runs but there's nobody to send it to, it just exits silently? There should be at least some log message that this happened.

liampond:
Claude is telling me that this won't run without a celery beat scheduler in this repo and in the ansible one.

PouyaMohseni:
@liampond, There is no need to run this locally. The celery beat is already added to the ansible.

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,352 @@
...
+        .values("source_id", "source__published", "folio", "s_sequence")
+        .annotate(count=Count("id"))
+        .filter(count__gt=1)
+        .order_by("source_id", "folio", "s_sequence")
+    )
+
+    published = [g for g in chant_dups if g["source__published"]] + \
+                [g for g in seq_dups if g["source__published"]]
+    unpublished = [g for g in chant_dups if not g["source__published"]] + \
+                  [g for g in seq_dups if not g["source__published"]]
```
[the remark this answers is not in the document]
PouyaMohseni: I think as this runs at most weekly/monthly the duplicate queries are negligible in practice and not worth the extra complexity.

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,363 @@
...
+        .exclude(genre__isnull=True)
+        .filter(cantus_id__in=ci_ids)
+        .values_list("cantus_id", "genre__name")
+        .distinct()
+    )
+    distinct_cids = {cid for cid, _ in pairs}
+
+    # Fetch CI genres in parallel
+    ci_genres: dict[str, Optional[str]] = {}
+    with ThreadPoolExecutor(max_workers=CI_WORKERS) as executor:
```
liampond, line 191: Should add a rate limit so we don't crash their site

liampond: Possibly a time limit as well?

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,363 @@
...
+                        f"genre={getattr(item.genre, 'name', '')}  "
+                        f"mode={item.mode}"
+                    )
+        lines.append("")
+
+    return "\n".join(lines)
+
+
+@shared_task(name="cantusdb.run_data_checks")
+def run_data_checks() -> None:
```
liampond, line 382: Maybe add some tests for this?

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,363 @@
...
+
+    for section in ("published", "unpublished"):
+        items = result.get(section) or []
+        lines.append(f"--- {section.capitalize()} ({len(items)}) ---")
+        if not items:
+            lines.append("No issues found.")
+        else:
+            for item in items:
+                if isinstance(item, dict):
+                    # duplicate folio/sequence results are dicts
```
liampond, line 356: I think it would be nice to include `chant_id` here, perhaps as links so it's easy for reviewers to go to the chants? If you add this, remember to change `chant_dups` and `seq_dups`.

django/cantusdb_project/main_app/admin/data_check_config.py
```diff
@@ -0,0 +1,11 @@
...
+
+from main_app.models.data_check_config import DataCheckConfig
+
+
+@admin.register(DataCheckConfig)
+class DataCheckConfigAdmin(admin.ModelAdmin):
+    list_display = ("frequency", "scope", "last_run")
+    list_filter = ("frequency", "scope")
+    readonly_fields = ("last_run",)
+    filter_horizontal = ("recipients",)
```
liampond, line 11: This is minor, but I think we should disable the ability to add new config rows. Users should edit the existing row instead.

```suggestion
    filter_horizontal = ("recipients",)

    def has_add_permission(self, request) -> bool:
        # Singleton config: allow creating it once, then only editing that row.
        return not DataCheckConfig.objects.exists()
```

liampond:
There is already a Cantus Index module in `cantusindex.py` that handles the domain, request, and BOM/JSON decoding. Can you route the CI calls through the existing helper instead?

liampond:
The genre check fetches from Cantus Index with a 300s timeout, and at ~10 workers with a per-request delay, it can only cover on the order of ~10k IDs per run. If it times out, it looks like there's no warning in the logs.

Do you know if 300s is enough? And regardless, I think we should flag/log if the genre check times out.

liampond:
How about a test for a cantus_id that has more than one local genre, one matching CI and one not, asserting that the mismatching one is still flagged? Claude is telling me there are about 789 cantus_ids with multiple genres, so it thinks this is worth it. Up to you.

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,454 @@
...
+                    )
+                else:
+                    # Chant model instances
+                    lines.append(
+                        f"  source_id={item.source_id}  "
+                        f"chant_id={item.id}  "
+                        f"source={getattr(item.source, 'siglum', item.source_id)}  "
+                        f"folio={item.folio}  "
+                        f"cantus_id={item.cantus_id}  "
+                        f"genre={getattr(item.genre, 'name', '')}  "
```
liampond, line 373: This accesses `item.genre`, but `check_cantus_ids_not_in_ci`, `check_blank_cantus_id`, and `check_blank_mode` only `select_related("source")` — so this is one extra query per row. Add `"genre"` to their `select_related`.

django/cantusdb_project/main_app/tasks.py
```diff
@@ -1,10 +1,454 @@
...
+    """
+    qs = (
+        Chant.objects.filter(id__in=chant_ids)
+        if chant_ids is not None
+        else Chant.objects.all()
+    )
+
+    invalid = qs.filter(
+        Q(position="M") & ~Q(service__name__in=["V", "V2"])
+        | Q(position="B") & ~Q(service__name="L")
```
liampond: `~Q(service__name__in=["V","V2"])` won't match rows where `service` is NULL (SQL `NOT (NULL IN ...)` is NULL), so a position-M/B chant with no service assigned slips through. If you want those flagged, add `| Q(service__isnull=True)` to each branch.

PouyaMohseni: Great catch on this. Thank you

PR #2084, opened 2026-06-10
feat: add static server-maintenance notice banner

liampond:
## What

Adds a site-wide notification banner announcing the upcoming server
maintenance window:

> **SERVICE NOTICE:** Server maintenance is happening on Friday, June 12.
> We apologize for any disruption.

It's a yellow Bootstrap `alert-warning` rendered full-width at the top of
every page (just below `<body>`, above the header).

## Why static HTML instead of the SiteBanner model (#2083 / #2080)

The full admin-managed `SiteBanner` feature can't land cleanly on
`production` right now — `production` is behind `staging` and lacks
migrations 0041/0042, so cherry-picking the banner's `0043_sitebanner`
migration runs into migration-history conflicts.

Since this is a one-off, time-boxed notice, this PR skips all of that:
no model, no migration, no context processor, no admin. Just three lines
of static markup in `base.html`. Nothing to migrate, nothing that can
fail at deploy time, and trivial to remove.

## Scope / notes

- Shows on **every page in every environment** this branch deploys to
  (no `PROJECT_ENVIRONMENT` gate) — appropriate for a server-wide notice.
- No admin toggle and no dismiss button by design.
- **Remove after the maintenance window:** revert this commit or delete
  the branch from `production`.
- This is deliberately not a replacement for the proper `SiteBanner`
  feature (#2083), which should still land via the normal
  `develop → staging → production` path once migration history is sorted.

## Changes

- `django/cantusdb_project/templates/base.html` (+3)

## Testing

Ran the dev stack locally on this branch — homepage and `/login/` return
200 and the banner renders as expected (verified in the page HTML and a
browser screenshot).


liampond:
Wrong branch

PR #2093, opened 2026-06-18
Fix `import_cantorales` overwrites + make migration 0042 schema-only

### LINKED ISSUE #2059: BUG: on staging it appears that the cantorales import overwrote existing manuscripts
I saw that the staging depiction of one of my manuscripts (HRC 145) was portrayed showing an additional editor I wasn't aware of, and a change to some of the metadata in the source description. My production description and metadata of the source are different. I was preserved as the editor and contributor for the item on staging, but it looks like the information added was a direct pull from the cantorales spreadsheet, overwriting my existing line in the description field. 

We will need to test this, and the extent, as well as identify what manuscripts exist inside Cantorales and CDB to figure out the extent of this.

liampond:
Resolves #2059. Builds on #2049.

## 1. Stop the import from overwriting existing sources (#2059)

`import_cantorales` used `update_or_create` keyed on `(holding_institution,
shelfmark)`, so a CSV row matching a source already in CDB overwrote its
metadata and added unexpected editors — this corrupted HRC 145 and others on
staging.

The command now **skips any existing source entirely** (logging it for review)
and only creates new ones. New sources also get their `Provenance` FK set, not
just `provenance_notes`, so provenance shows on the browse-sources page.

Tests: a regression test that an existing source is left fully untouched
(fields, editors, segment, no duplicate) plus two provenance tests.

## 2. Migration 0042 is now schema-only

`move_cantorales_contributors` `RunPython` step is dropped from 0042 (it now
only adds the field). That data step — moving Cantorales contributors from
`inventoried_by` to `source_data_contributed_by` — moves to a dated one-off
script (`main_app/management/one_off/`, run via `manage.py shell <`, reusing
`settings.CANTORALES_SEGMENT_ID`, idempotent).

This is safe where 0042 is already applied (e.g. staging): Django tracks
migrations by name, so the edited file does not re-run there — only fresh
databases get the schema-only version. Migration history is allowed to differ
across environments.


django/cantusdb_project/main_app/management/commands/import_cantorales.py
```diff
@@ -261,14 +263,45 @@ def handle(self, *args, **options):
...
 
+                # --- Skip existing sources entirely (issue #2059) ---
+                # An earlier version of this import used update_or_create, which
+                # overwrote curated metadata and added unexpected editors on
+                # sources that already existed in CDB. We now refuse to modify
+                # any existing source: if one already matches
+                # (holding_institution, shelfmark) we skip the whole row
+                existing = Source.objects.filter(
+                    holding_institution=institution, shelfmark=shelfmark
+                ).first()
```
ahankinson: You can use `.exists()` instead of retrieving the record, if all you want to do is find out if you have a match. 

https://docs.djangoproject.com/en/6.0/ref/models/querysets/#exists

liampond: Thanks!

PouyaMohseni:
@coderabbitai review

django/cantusdb_project/main_app/management/commands/import_cantorales.py
```diff
@@ -260,13 +270,32 @@ def handle(self, *args, **options):
                             f"  Created institution: {institution} ({rism})"
                         )
```
PouyaMohseni, line 264: The new `.exists()` skip check runs after the institution lookup/creation block. If a row's RISM institution doesn't yet exist in the DB, a new Institution gets created even when the source itself is about to be skipped as a pre-existing duplicate.

liampond: Implemented

django/cantusdb_project/main_app/management/commands/import_cantorales.py
```diff
@@ -287,70 +332,91 @@ def handle(self, *args, **options):
...
+                    contributor_raw = row[COL_CONTRIBUTOR].strip()
+                    if contributor_raw:
+                        for name in re.split(r",\s*", contributor_raw):
+                            name = name.strip()
+                            if name in contributor_users:
+                                source.source_data_contributed_by.add(
+                                    contributor_users[name]
+                                )
+
+                    reversion.set_comment("import_cantorales: issue #2059")
```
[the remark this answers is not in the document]
liampond: Already covered — segment_m2m, century, and source_data_contributed_by are in Source's
serialized fields, and I confirmed a revert restores all three without `follow`. Adding them
would also revert the related Segment/Century/User rows on any source revert, which we don't
want; `follow` is set to ('identifiers', 'source_links') because those are reverse relations
that genuinely need it.

kyrieb-ekat (approved):
Good fix. confirmed the create-only + skip-existing logic actually prevents the HRC 145-style overwrite, and traced the image_link→SourceURL claim against source_detail.html's rendering. Two non-blocking nitpicks: no `reversion.set_user()` alongside `set_comment()`, and contributor matching is by full_name only (fine for this CSV, just noting it). LGTM.

